PR #5, opened 2022-01-18
build: fix ci build caching

kormide:
@thesayyn 

Github actions caches don't update already existing caches, so the cache that gets loaded could be old. This sets a new cache using a cache key derived from a hash of files. The use of restore-keys: bazel-cache- will find the most recent cache stored scoped to this branch or a parent branch.


thesayyn:
Sweet. thanks Derek!

PR #6, opened 2022-01-19
fix: move cache flags out of ci bazelrc

kormide:
@thesayyn Some more fixes to ci build cacheing. See the rules-template changes here: https://github.com/bazel-contrib/rules-template/pull/4

Apparently bazelrc doesn't recognize environment variables.


thesayyn:
Thanks!

PR #7, opened 2022-01-20
build: move ci cache flags to bazelrc

kormide:
Sorry to keep changing this, but we've settled on putting the cache flags in bazelrc in rules-template since we realized that it supports `~` paths. Feel free to keep or close as this doesn't change anything functionally.

### DIFF
--- a/.github/workflows/ci.bazelrc
+++ b/.github/workflows/ci.bazelrc
@@ -3,6 +3,9 @@
 
 # Debug where options came from
 build --announce_rc
+# This directory is configured in GitHub actions to be persisted between runs.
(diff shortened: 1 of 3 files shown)


thesayyn:
This one feels right.

kormide:
It looks like there was an http 404 in the ci build, but I don't think it was anything caused by my changes. Are you able to re-trigger the ci for me?

thesayyn:
> It looks like there was an http 404 in the ci build, but I don't think it was anything caused by my changes. Are you able to re-trigger the ci for me?

sure thing.

PR #27, opened 2022-08-16
Fix paths to caches for Bazel in CI

tetsuok:
This fixes paths to caches for Bazel in CI, in the same way aspect-build/rules_js#366 fixed.


alexeagle (approved):
thanks!

PR #31, opened 2022-08-23
feat: implement structure_test

thesayyn:


example/js/BUILD.bazel
```diff
@@ -40,3 +30,12 @@ oci_tarball(
     image = ":image",
     repotags = ["example/js:latest"],
 )
+
+structure_test(
+    name = "test",
+    config = ["test.yaml"],
+    image = ":image",
+    target_compatible_with = [
+        "@platforms//os:linux",
```
f0rmiga, line 39: Why only Linux?

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
+"contains container helper functions for js_binary"
```
f0rmiga, line 1:
```suggestion
"This module contains container helper functions for js_binary."
```

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@aspect_bazel_lib//lib:paths.bzl", "to_manifest_path")
+load("@rules_python//python:defs.bzl", "py_binary")
+
+# BAZEL_BINDIR has to be set to '.' so that js_binary preserves the PWD when running inside container.
+# See https://github.com/aspect-build/rules_js/tree/dbb5af0d2a9a2bb50e4cf4a96dbc582b27567155#running-nodejs-programs
+# for why this is needed.
+_LAUNCHER_TMPL = """
+export BAZEL_BINDIR=.
+source {executable_path}
```
f0rmiga, line 14: Why does it use `source` instead of `exec`?

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+        content = _LAUNCHER_TMPL.format(executable_path = executable_path),
+        is_executable = True,
+    )
+
+    return launcher
+
+def _runfile_path(ctx, file, runfiles_dir):
+    return "/".join([runfiles_dir, to_manifest_path(ctx, file)])
+
+def _should_include(destination, include, exclude):
```
f0rmiga, line 32: This function deserves a comment.

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+    included = include in destination or include == ""
+    excluded = exclude in destination and exclude != ""
+    return included and not excluded
+
+def _runfiles_impl(ctx):
+    default = ctx.attr.binary[DefaultInfo]
```
f0rmiga, line 38: The label `default` is meaningless unless the person reading the code comes to this line.

```suggestion
    default_info = ctx.attr.binary[DefaultInfo]
```

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+    file_map = {}
+
+    if _should_include(original_executable_path, ctx.attr.include, ctx.attr.exclude):
+        file_map[original_executable_path] = executable
+
+    if _should_include(executable_path, ctx.attr.include, ctx.attr.exclude):
+        file_map[executable_path] = launcher
+
+    manifest = default.files_to_run.runfiles_manifest
+    runfiles_dir = "/".join([ctx.attr.root, manifest.short_path.replace(manifest.basename, "")[:-1]])
```
f0rmiga, line 54: I see you are doing lots of path manipulation by hand like this. Why not use skylib instead? https://github.com/bazelbuild/bazel-skylib/blob/main/docs/paths_doc.md

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+
+    files = depset(transitive = [default.files, default.default_runfiles.files])
+
+    for file in files.to_list():
+        destination = _runfile_path(ctx, file, runfiles_dir)
+        if _should_include(destination, ctx.attr.include, ctx.attr.exclude):
+            file_map[destination] = file
+
+    # executable and launcher should not go into runfiles directory so we add it to files here
+    files = depset([executable, launcher], transitive = [files])
```
f0rmiga, line 64: This part is confusing as `files` gets reassigned. This can easily introduce code smell as the code gets maintained over time.

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+    for symlink in default.data_runfiles.symlinks.to_list():
+        destination = "/".join([runfiles_dir, ctx.workspace_name, symlink.path])
+        if not _should_include(destination, ctx.attr.include, ctx.attr.exclude):
+            continue
+        if hasattr(file_map, destination):
+            file_map.pop(destination)
+        info = PackageSymlinkInfo(
+            target = "/%s" % _runfile_path(ctx, symlink.target_file, runfiles_dir),
+            destination = destination,
+            attributes = {"mode": "0777"},
```
f0rmiga, line 79: Why not `0700`?

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+# ```
+# to determine a file is file or directory pkg_tar checks `is_directory` if it is false then it makes the assumption
+# that the file is in fact is a file which not true for `symlink_file` above. this is where pkg_tar fails to build
+# a correct tar file.
+# In order to fix this, manifest written pkg_tar should be fixed by looking for files that are symlink instead of file.
+
+# TODO(thesayyn): remove this once pkg_tar is fixed.
+#
+# See: https://github.com/bazelbuild/rules_pkg/issues/115#issuecomment-1190494335
+BUILD_TAR = """
```
f0rmiga, line 135: I foresee that this will live here for a while still. Since this is not a trivial code, please move it to a separate file and consume it as input to the rule. Ideally, you would also set up a linter for Python. Can be using pre-commit.

example/js/js_image_layer.bzl
```diff
@@ -0,0 +1,260 @@
...
+
+    runfiles_kwargs = dict(
+        common_kwargs,
+        binary = binary,
+        root = root,
+    )
+
+    pkg_tar_kwargs = dict(
+        kwargs,
+        # Be careful with this option. Leave it as is if you don't know what you are doing
```
f0rmiga, line 223: I know what you mean, but this comment is irrelevant. Instead of phrasing it this way, point people to the issue that discusses this.

oci/BUILD.bazel
```diff
@@ -16,6 +16,12 @@ toolchain_type(
     visibility = ["//visibility:public"],
 )
 
+toolchain_type(
+    name = "st_toolchain_type",
+    visibility = ["//visibility:public"],
+)
+
+
```
f0rmiga, line 24:
```suggestion
```

oci/private/structure_test.bzl
```diff
@@ -0,0 +1,62 @@
+"""Loading oci_tarball into the daemon and executes structure_test executable
```
f0rmiga, line 1: This comment is not helpful to anyone unfamiliar with this file.

oci/private/structure_test.bzl
```diff
@@ -0,0 +1,62 @@
...
+}
+
+CMD = """\
+#!/usr/bin/env bash
+set -o pipefail -o errexit -o nounset
+
+readonly ST_PATH="{st_path}"
+readonly FIXED_ARGS="{fixed_args}"
+
+"$ST_PATH" test $FIXED_ARGS $@
```
f0rmiga, line 18:
```suggestion
#!/usr/bin/env bash

exec "{st_path}" test {fixed_args} "$@"
```

oci/private/structure_test.bzl
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+def _structure_test_impl(ctx):
+    st = ctx.toolchains["@contrib_rules_oci//oci:st_toolchain_type"].st_info
```
f0rmiga, line 22: Same as `default` above. At least use `st_info`. Even though I could argue that `st_info` itself is a bad naming for the provider.

oci/private/structure_test.bzl
```diff
@@ -0,0 +1,62 @@
...
+            fixed_args = " ".join(fixed_args),
+        ),
+        is_executable = True,
+    )
+
+    runfiles = ctx.runfiles(files = ctx.files.image + ctx.files.config + st.structure_test_files)
+
+    return DefaultInfo(runfiles = runfiles, executable = launcher)
+
+# TODO: https://github.com/bazel-contrib/rules_oci/issues/40
```
f0rmiga, line 54: Since you already have the issue, remove the TODO from code.

oci/private/structure_test.bzl
```diff
@@ -0,0 +1,62 @@
+"""Loading oci_tarball into the daemon and executes structure_test executable
+
+For some reason if DOCKER_HOST needs to be specified, use --test_env=DOCKER_HOST='unix://<path_to_sock>'
+Beware that this is going to lead to cache misses.
+"""
+attrs = {
```
f0rmiga, line 6:
```suggestion
_attrs = {
```

oci/repositories.bzl
```diff
@@ -110,6 +110,43 @@ zot_repositories = repository_rule(
     },
 )
 
+STRUCTURE_TEST_BUILD_TMPL = """\
+# Generated by container/repositories.bzl
+load("@contrib_rules_oci//oci:toolchain.bzl", "structure_test_toolchain")
+structure_test_toolchain(
+    name = "st_toolchain",
```
f0rmiga, line 117: Don't be afraid of using long labels. `structure_test_toolchain` is much better than `st_toolchain`.

oci/toolchain.bzl
```diff
@@ -153,6 +153,76 @@ registry_toolchain = rule(
     },
     doc = """Defines a registry toolchain.
 
+For usage see https://docs.bazel.build/versions/main/toolchains.html#defining-toolchains.
+""",
+)
+
+
```
f0rmiga, line 160:
```suggestion
```

f0rmiga (changes requested):

example/js/js_image_layer.bzl
thesayyn: 1- avoid $0 type of bugs. RUNFILES helper relies on that heavily. 
2- any future implicit reliance of executable_path being the entrypoint.

thesayyn: 1- avoid $0 type of bugs. RUNFILES helper relies on that heavily. 
2- any future implicit reliance of executable_path being the entrypoint.

thesayyn: 1- avoid $0 type of bugs. RUNFILES helper relies on that heavily. 
2- any future implicit reliance of executable_path being the entrypoint.

example/js/BUILD.bazel
thesayyn: container rule aka oci_image is not ready to handle transitions. will handle them #33

example/js/js_image_layer.bzl
thesayyn: 1- avoid $0 type of bugs. RUNFILES helper relies on that heavily. 
2- any future implicit reliance of executable_path being the entrypoint.

example/js/js_image_layer.bzl
thesayyn: I don't know. this was the default in rules_pkg.

example/js/js_image_layer.bzl
thesayyn: to avoid taking it as a dependency. Since windows style paths are irrelevant in this context, I don't see the value.

oci/private/structure_test.bzl
thesayyn: I am not sure if we are consistent about invoking the tools. this differs from what we've recently merged. https://github.com/bazel-contrib/rules_oci/blob/5b41558ede238413d9ae6025f7021a106da95241/oci/private/container.sh.tpl#L39-L42

example/js/js_image_layer.bzl
f0rmiga: I can think of at least a major value of it using skylib's `paths`: readability and maintainability. Point me to a project that doesn't have skylib as a dependency (demo repos don't count).

example/js/js_image_layer.bzl
f0rmiga: It doesn't need to be visible to users. Again, this is not trivial code; please, move it to a separate file.

example/js/js_image_layer.bzl
f0rmiga: The way it's phrased is not helpful. Please, remove the `Leave it as is ...` sentence to one that points people to the documentation and explains why.

oci/private/structure_test.bzl
f0rmiga: You are comparing apples and oranges. One is a full script with different code paths, the other is a wrapper around a single tool and some args. https://tldp.org/LDP/abs/html/x17974.html

example/js/js_image_layer.bzl
f0rmiga: We need to understand the code that we copy-paste.

example/js/js_image_layer.bzl
thesayyn: still, it's easier to tell users to copy just one file than two and change their build files. left a comment here to remind us to move it to a separate file when we promote it to the public API. https://github.com/aspect-build/rules_js/issues/304#issuecomment-1325935905

example/js/js_image_layer.bzl
thesayyn: Sure.

example/js/js_image_layer.bzl
thesayyn: Fair point. I'll replace them.

example/js/js_image_layer.bzl
thesayyn: tried with exec. it doesn't work with it.

f0rmiga (approved):

alexeagle:
@thesayyn WDYT about naming it `oci_structure_test`? I think all rules ought to have a common prefix to prevent collisions and make it clear what ruleset it comes from.

PR #34, opened 2022-09-25
feat: implement oci_image_index

### LINKED ISSUE #17: FR: implement oci_index to support multi arch images
An example build file looks like below 

```py
# this image will have 4 layers in total
oci_image(
    name = "debian10",
    layers = [
        ":debian10_fs",
        ":debian10_passwd",
        ":libs"
    ]
)

platforms = {
    # 1:2+ transitions
    "@platforms//linux": [
        "@platforms//cpu:x86_64",    
        "@platforms//cpu:arm64",
        "@platforms//cpu:armel",
        "@platforms//cpu:ppc",
    ]
}

# will yield multi-platform image
oci_index(
    name = "debian_multi",
    image = ":debian10",
    platforms = platforms
)
```

thesayyn:
closes #17


oci/private/index.bzl
```diff
@@ -0,0 +1,77 @@
+"Implementation details for oci_index rule"
```
gregmagolan, line 1: is the naming `index` a common name for a multi-arch OCI compatible image?

oci/private/index.sh.tpl
```diff
@@ -0,0 +1,61 @@
...
+    case "$ARG" in
+        (--output=*) OUTPUT="${ARG#--output=}"; create_oci_layout "$OUTPUT" ;;
+        (--image=*) CURRENT_IMAGE="${ARG#--image=}"; add_image "$CURRENT_IMAGE" "$OUTPUT" ;;
+        (--blob=*) copy_blob "${CURRENT_IMAGE}" "$OUTPUT" "${ARG#--blob=}" ;;
+        (*) echo "Unknown argument ${ARG}"; exit 1;;
+    esac
+done
+
+
+export checksum=$(shasum -a 256 "${OUTPUT}/manifest_list.json" | cut -f 1 -d " ")
```
gregmagolan, line 56: you could get the sha256 has of the manifest_list.json file with an action and make it more portable since some systems may not have `shasum` installed?

https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/hash/hash.bzl

gregmagolan: check `cli/brew/brew_artifacts.bzl` in silo for an example of how to use that bit of starlark that ships with Bazel in a rule

oci/private/index.sh.tpl
```diff
@@ -0,0 +1,61 @@
...
+export size=$(wc -c < "${OUTPUT}/manifest_list.json")
```
gregmagolan, line 57: Is `wc` available on all systems?

example/multi_arch/test.yaml
```diff
@@ -0,0 +1,7 @@
+schemaVersion: "2.0.0"
+
+commandTests:
+  - name: "echo hello"
+    command: "bash"
+    args: ["test.bash"]
+    expectedOutput: ["hello world!"]
```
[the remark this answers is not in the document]
thesayyn: yes. but structure_test does not work with multi arch images yet. this file is no-op. i should remove it.

oci/private/index.bzl
thesayyn: index is a technical name for multi-arch image as it's made of list of manifests. maybe `oci_image_index` is a better name? 
https://github.com/opencontainers/image-spec/blob/main/image-index.md

oci/private/index.sh.tpl
thesayyn: looks like it depends on python rules. I don't feel good about using that. I'll look for alternatives.

thesayyn: especially with python3 python2 mess bazel has. maybe https://github.com/jonelo/jacksum/releases is a more sensible choice here.

oci/private/index.sh.tpl
gregmagolan: I think wc is fine. `shasum` is going to be problematic tho on some linux distros.

I had a `shasum` call landed in Silo and Thulio replaced it there since it caused issues.

oci/private/index.sh.tpl
gregmagolan: sounds good. carrying a python dep is bad.

tho I'm also fine with landing this as is since most linux distros will have shasum and we can fix it as a follow-up.

example/multi_arch/test.yaml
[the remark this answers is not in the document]
gregmagolan: 👍

oci/private/index.sh.tpl
thesayyn: I'll do some looking to replace it with a hermetic tool.  https://github.com/jonelo/jacksum/releases looks good. though it requires a java_binary target. How do you feel about having dep on built-in java rules?

oci/private/index.bzl
gregmagolan: yup. lets switch it to `oci_image_index`. that looks like name in the spec as well https://specs.opencontainers.org/image-spec/image-index/?v=v1.0.1

thesayyn: agreed.

oci/private/index.sh.tpl
thesayyn: filed https://github.com/bazel-contrib/rules_oci/issues/41

oci/private/index.sh.tpl
thesayyn: filed https://github.com/bazel-contrib/rules_oci/issues/41 for future.

gregmagolan (approved):

gregmagolan (approved):

Issue #19, opened 2022-08-05
FR: implement oci_push


thesayyn:
Features

- Can be run `bazel run //image:push -- --registry index.docker.io --tag latest --tag 1.x.x`.  
- Supports push of the image with multiple tags

PR #36, opened 2022-09-28
feat: implement oci_cosign

thesayyn:
#19


cosign/private/sign.bzl
```diff
@@ -0,0 +1,77 @@
+"Implementation details for sign rule"
+
+_DOC = """Sign an oci_image using cosign binary at a remote registry.
+
+It sings the image by its digest determined beforehand.
```
gregmagolan, line 5: "signs the image"

would be great if it also sang but probably asking too much of Bazel there

cosign/private/sign.bzl
```diff
@@ -0,0 +1,77 @@
...
+)
+
+cosign_sign(
+    name = "sign",
+    image = ":image",
+    repository = "index.docker.io/org/image"
+)
+```
+
+`repository` attribute can be overridden using the `--repository` option.
```
gregmagolan, line 19: nit: `--repository` flag

cosign/private/sign.bzl
```diff
@@ -0,0 +1,77 @@
...
+}
+
+def _cosign_sign_impl(ctx):
+    cosign = ctx.toolchains["@contrib_rules_oci//cosign:toolchain_type"]
+    yq = ctx.toolchains["@aspect_bazel_lib//lib:yq_toolchain_type"]
+
+    if ctx.attr.repository.find(":") != -1 or ctx.attr.repository.find("@") != -1:
+        fail("repository attribute should not contain digest or tag.")
+
+    executable = ctx.actions.declare_file("cosign_%s" % ctx.label.name)
```
gregmagolan, line 49: "cosign_%s.sh" % ctx.label.name

?

cosign/toolchain.bzl
```diff
@@ -0,0 +1,78 @@
+"""This module implements the cosign-specific toolchain rule."""
+
+def _to_manifest_path(ctx, file):
+    if file.short_path.startswith("../"):
+        return "external/" + file.short_path[3:]
```
gregmagolan, line 5: ohhh. this is going to depend on legacy_external_runfiles and will break in repositories that have `--nolegacy_external_runfiles`.

the correct approach is what we did with yq and just pass the binary itself to the toolchain info:
https://github.com/aspect-build/bazel-lib/blob/2689a48940007588e87f63acda704d72f2e2b6c2/lib/private/yq_toolchain.bzl#L111
and not do any conversion to manifest path

cosign/private/sign.bzl
```diff
@@ -0,0 +1,77 @@
...
+    ctx.actions.expand_template(
+        template = ctx.file._sign_sh_tpl,
+        output = executable,
+        is_executable = True,
+        substitutions = {
+            "{{cosign_path}}": cosign.cosign_info.cosign_path,
+            "{{yq_path}}": yq.yqinfo.bin.path,
```
gregmagolan, line 56: related to other comment, if you pass the binary File to the toolchain info then you choose between path and short_path here...

here this should be `yq.yqinfo.bin.short_path` so you're not depending on legacy external runfiles

cosign/toolchain.bzl
```diff
@@ -0,0 +1,78 @@
...
+    template_variables = platform_common.TemplateVariableInfo({
+        "COSIGN_BIN": cosign_path,
+    })
+    default = DefaultInfo(
+        files = depset(cosign_files),
+        runfiles = ctx.runfiles(files = cosign_files),
+    )
+    cosign_info = CosignInfo(
+        cosign_path = cosign_path,
+        cosign_files = cosign_files,
```
gregmagolan, line 43: you're already passing cosign_files via the `default` attribute of `platform_common.ToolchainInfo` below so you don't need the `cosign_files`

should probably follow the same pattern as we did for YQ:

```
    # Make the $(YQ_BIN) variable available in places like genrules.
    # See https://docs.bazel.build/versions/main/be/make-variables.html#custom_variables
    template_variables = platform_common.TemplateVariableInfo({
        "YQ_BIN": binary.path,
    })
    default_info = DefaultInfo(
        files = depset([binary]),
        runfiles = ctx.runfiles(files = [binary]),
    )
    yq_info = YqInfo(
        bin = binary,
    )

    # Export all the providers inside our ToolchainInfo
    # so the resolved_toolchain rule can grab and re-export them.
    toolchain_info = platform_common.ToolchainInfo(
        yqinfo = yq_info,
        template_variables = template_variables,
        default = default_info,
    )
```

https://github.com/aspect-build/bazel-lib/blob/2689a48940007588e87f63acda704d72f2e2b6c2/lib/private/yq_toolchain.bzl#L111

gregmagolan: I didn't review the other toolchains in this rule set but perhaps we need to make similar changes in those as well

gregmagolan: ohhh.. if we're supporting a fixed `cosign_path` attribute in the toolchain however then we have no File object to pass 😞 

we don't have a yq_path option in the yq toolchain

gregmagolan: yq: https://github.com/aspect-build/bazel-lib/blob/2689a48940007588e87f63acda704d72f2e2b6c2/lib/private/yq_toolchain.bzl#L127

gregmagolan: How about passing the File object _or_ the path as a string either in the same field or in different fields. If its the same field the consumer will have to check if the value is a "string" and if it is treat it as an absolute path on disk.

Your `readonly COSIGN="$RUNFILES_ROOT/{{cosign_path}}"` will need adjusting as well if `{{cosign_path}}` is an absolute path.

Or for the mvp here, just do what yq does and keep it simple and don't allow passing an absolute path to the toolchain

cosign/private/sign.bzl
thesayyn: oh god. good catch.

cosign/toolchain.bzl
thesayyn: followed the approach of yq for the MVP. removed `chdir` behavior for now to keep it simple.

cosign/private/sign.bzl
```diff
@@ -0,0 +1,77 @@
...
+    if ctx.attr.repository.find(":") != -1 or ctx.attr.repository.find("@") != -1:
+        fail("repository attribute should not contain digest or tag.")
+
+    executable = ctx.actions.declare_file("cosign_sign_{}.sh".format(ctx.label.name))
+    ctx.actions.expand_template(
+        template = ctx.file._sign_sh_tpl,
+        output = executable,
+        is_executable = True,
+        substitutions = {
+            "{{cosign_path}}": cosign.cosign_info.binary.short_path,
```
gregmagolan, line 55: so nice

gregmagolan (approved):

PR #38, opened 2022-10-12
feat: implement oci_push

### LINKED ISSUE #19: FR: implement oci_push

thesayyn:
closes #19

### DIFF
--- a/oci/private/push.bzl
+++ b/oci/private/push.bzl
@@ -2,16 +2,17 @@
 
 _DOC = """Push an oci_image or oci_image_index to a remote registry.
 
-Pushing and tagging are performed sequentially which MAY lead to non-atomic pushes if one the following events occur 
+Pushing and tagging are performed sequentially which MAY lead to non-atomic pushes if one the following events occur;
 
 - Remote registry rejects a tag due to various reasons. eg: forbidden characters, existing tags 
 - Remote registry closes the connection during the tagging
 - Local network outages
 
 In order to avoid incomplete pushes oci_push will push the image by its digest and then apply the `default_tags` sequentially at
-remote registry. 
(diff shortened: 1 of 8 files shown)


oci/private/push.sh.tpl
```diff
@@ -0,0 +1,47 @@
...
+ALL_ARGS=(${FIXED_ARGS[@]} $@)
+set -- ${ALL_ARGS[@]}
+
+REPOSITORY="{{repository}}"
+TAGS=()
+ARGS=()
+
+while (( $# > 0 )); do
+  case $1 in
+    (-t|--tag)
```
gregmagolan, line 19: nit: ordering of the flag handling is reversed between `tag` and `repository` with `-t|--tag` and then `--tags=*` for tag and `--repository=*` and `-r|--repository` for repository

oci/private/push.bzl
```diff
@@ -0,0 +1,106 @@
...
+        ":app_windows_amd64",
+    ]
+)
+
+oci_push(
+    image = ":app_image",
+    repository = "ghcr.io/<OWNER>/image",
+    default_tags = ["6.4.1-rc1"]
+)
+```
```
gregmagolan, line 51: Should mention that you need to `bazel run` to push and that the tool accepts `-r|--repository` and `-t|--tag` flags so you can set the repository & tag at runtime. For tag this is particularly useful since when pushing semver versions users probably want the version to come from the latest tag on the repository and not be hard-coded in the BUILD file.

gregmagolan (approved):

Issue #40, opened 2022-11-17
Move structure_test to distroless-tools


thesayyn:

alexeagle:
We'll just own starlark code for the rule, but we are not responsible for the Go program provided by https://github.com/GoogleContainerTools/container-structure-test

PR #42, opened 2022-12-22
refactor: rename structure_test to oci_structure_test

alexeagle:
I think it's better to have a consistent prefix and avoid any possible naming conflicts.


gregmagolan (approved):
I agree. The `js_*` and `npm_*` common prefixes worked out really nicely in rules_js. Its a good practice to follow for a rule set.

thesayyn:
the upstream container-structure-test repository barely gets any bug fixes and has been dead for a good two years.  Also, it doesn't keep up with the container ecosystem as well.

Ideally, we shouldn't be on the hook for bugs that might arise from having it hosted here. and that's the reason why it's a hidden part of the public API and has no documentation https://github.com/bazel-contrib/rules_oci/tree/main/docs.

I had filed https://github.com/bazel-contrib/rules_oci/issues/40 to remind us to move it out of rules_oci.

alexeagle:
Oh hm interesting, OTOH since we use it in all the examples here it does seem like every user of rules_oci will need some structure test right? should everyone depend on distroless-tools, even if they don't have any usage of distroless?

thesayyn:
> OTOH since we use it in all the examples here it does seem like every user of rules_oci will need some structure test right

that's right. though I am not sure how popular `container_test` rule from rules_docker was.

> should everyone depend on distroless-tools, even if they don't have any usage of distroless?

if they want the structure_test capability they will need to bring `distroless-tools` ruleset as well. not necessarily has to use `distroless` as a base image.

PR #51, opened 2023-01-13
Cache results per OS

tetsuok:
Currently, GitHub Actions caches on macOS are not uploaded because the key is shared between Ubuntu and macOS and, because CI in Ubuntu normally finishes much earlier (See screenshot of macOS build below). This prevents Bazel on macOS from fully taking advantage of GitHub Actions caches. (macOS runner always downloads caches built by the Ubuntu runner, but those caches are not always usable for macOS such as Go object files)
This PR allows to use cache keys per OS so that caches can be used on macOS.

Screenshot 2023-01-13 at 10 19 56


thesayyn (approved):
Great catch. Thanks!

PR #57, opened 2023-01-16
refactor: cleanup toolchains

### LINKED ISSUE #41: remove undeclared dependency on shasum
Currently, `oci_image_index`  depends on `wc` and `shasum` installed on host machine. These dependencies are undeclared and may lead `oci_image_index` to fail if any of these tools are absent on the host. 

Alternatives considered; 

https://github.com/jonelo/jacksum - requires dependency on `java_binary`. 
https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/hash/hash.bzl - requires dependency on `py_binary`

thesayyn:
closes #41 and disabled legacy external runfiles

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,6 +3,7 @@
 # Settings that apply only to CI are in .github/workflows/ci.bazelrc
 
 build --incompatible_strict_action_env
+common --nolegacy_external_runfiles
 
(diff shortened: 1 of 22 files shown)


gregmagolan (approved):
🦖

PR #59, opened 2023-01-27
chore: add example for pulling

thesayyn:


example/pull/BUILD.bazel
```diff
@@ -0,0 +1,28 @@
...
+
+pkg_tar(
+    name = "app",
+    srcs = ["app.bash"],
+)
+
+oci_image(
+    name = "image",
+    architecture = select({
+        "@platforms//cpu:arm64": "arm64",
```
alexeagle, line 12: Why have select when the values are the same?

example/pull/pull.bzl
```diff
@@ -0,0 +1,31 @@
+"a repository rule fetching container images using crane"
+_PULL_TMPL = """\
+# based on https://github.com/bazel-contrib/rules_oci/blob/b900a6322ae4e68b431f7d65ebfef604657bd030/example/BUILD.bazel#L19-L43
+
+genrule(
+    name = "image",
+    outs = ["layout"],
+    cmd = "$(CRANE_BIN) pull {image} $@ --format=oci --platform=linux/{architecture}",
```
alexeagle, line 8: This doesn't seem desirable to have the fetch occur in a build action. Can we call this from the repository rule instead?

example/pull/BUILD.bazel
thesayyn: they are different, `amd64` and `arm64`.

example/pull/pull.bzl
thesayyn: I preferred this because makes it a lazy fetch rather than a eager repository rule.  since `local = True` is set to true it's almost identical to running the tool via `rctx.execute`, no-remote, no-sandboxing etc. except maybe for subtle differences in `$PATH`?

I'd have to write more for the repository_rule equivalent of this. I'm fine making it a two-phase repository rule with HTTP cache.

alexeagle:
I think this is replaced with #67

PR #70, opened 2023-02-14
stamping for tags

### LINKED ISSUE #69: Support stamping
I should be able to push a specific tag with something like `bazel run --stamp --embed_label=1.2.3 //path/to:my_oci_push`

Will use https://github.com/aspect-build/bazel-lib/blob/main/docs/stamping.md

alexeagle:
Fixes #69

### DIFF
--- a/examples/push/BUILD.bazel
+++ b/examples/push/BUILD.bazel
@@ -32,7 +32,7 @@ stamped_tags(
     image_tags = [
         # With --stamp, use the --embed_label value, otherwise use 0.0.0
         """($stamp.BUILD_EMBED_LABEL // "0.0.0")""",
-        "latest",
+        "nightly",
     ],
 )
 
--- a/examples/push/BUILD.bazel
+++ b/examples/push/BUILD.bazel
@@ -1,4 +1,4 @@
-load("//oci:defs.bzl", "oci_image", "oci_image_index", "oci_push")
+load("//oci:defs.bzl", "oci_image", "oci_image_index", "oci_push", "stamped_tags")
 
 oci_image(
     name = "image",
@@ -9,8 +9,8 @@ oci_image(
 
 oci_push(
     name = "push_image",
-    default_tags = ["latest"],
     image = ":image",
+    image_tags = ["latest"],
     repository = "index.docker.io/<ORG>/image",
 )
 
@@ -27,10 +27,19 @@ oci_image_index(
     ],
 )
 
+stamped_tags(
+    name = "stamped",
+    image_tags = [
+        # With --stamp, use the --embed_label value, otherwise use 0.0.0
+        """($stamp.BUILD_EMBED_LABEL // "0.0.0")""",
+        "latest",
+    ],
+)
+
 oci_push(
     name = "push_image_index",
-    default_tags = ["nightly"],
     image = ":image_index",
+    image_tags = ":stamped",
     repository = "index.docker.io/<ORG>/image",
 )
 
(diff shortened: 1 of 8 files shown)


alexeagle:
fyi @psalaberria002 in case you'd like to comment

psalaberria002:
> fyi @psalaberria002 in case you'd like to comment

This solves stamping for tags, but not oci labels. Any thoughts on how to stamp the labels? It should be done when building the image ideally, not in oci_push.

alexeagle:
Agreed, @thesayyn and I talked about that, anything that belongs in the image metadata will be a separate mechanism (maybe also using `jq` to produce stamped json, then merge that file in with other metadata we already control)

oci/defs.bzl
```diff
@@ -1,13 +1,77 @@
...
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
 
 oci_tarball = _oci_tarball
 oci_image = _oci_image
 oci_image_index = _oci_image_index
-oci_push = _oci_push
+oci_push_rule = _oci_push
 structure_test = _structure_test
+
+def stamped_tags(name, image_tags, **kwargs):
```
thesayyn, line 18: is this going to be in our public API contract once we hit 1.0? I'd like to keep our public API lean. 
I don't have any suggestions about how to keep this userland though.

examples/push/BUILD.bazel
```diff
@@ -27,10 +27,19 @@ oci_image_index(
     ],
 )
 
+stamped_tags(
```
thesayyn, line 30: naming suggestion: `stamp_tags`  usually our rules use verb names instead of a version specific to a tense.

examples/push/BUILD.bazel
```diff
@@ -9,8 +9,8 @@ oci_image(
 
 oci_push(
     name = "push_image",
-    default_tags = ["latest"],
     image = ":image",
+    image_tags = ["latest"],
```
thesayyn, line 13: hmm, `image_tags` sounds wrong to me since it's the manifest that gets tagged within that repository. how about `repotags` so we have a unified naming with `oci_tarball`? 

https://github.com/bazel-contrib/rules_oci/blob/main/docs/tarball.md#oci_tarball

oci/private/push.sh.tpl
```diff
@@ -44,4 +45,8 @@ REFS=$(mktemp)
 for tag in "${TAGS[@]+"${TAGS[@]}"}"
 do
   "${CRANE}" tag $(cat "${REFS}") "${tag}"
-done
\ No newline at end of file
+done
+
+if [[ -e "${TAGS_FILE:-}" ]]; then
+  cat "${TAGS_FILE}" | xargs -n1 "${CRANE}" tag $(cat "${REFS}")
```
thesayyn, line 51: oh, we should probably use coreutils for these. could be done as a follow-up for all rules though.

oci/defs.bzl
alexeagle: It could go to aspect_bazel_lib perhaps. Or it can go in userland to start. I could imagine a tool chain for a expand_template binary

thesayyn: maybe userland is a better start.  I have too many ideas for useful utilities like this but I don't do any of them because they don't belong to the public API.

alexeagle: moved to the example

examples/push/BUILD.bazel
alexeagle: okay sure, matters less now that it's in the example

examples/push/BUILD.bazel
alexeagle: okay, renamed to repotags

oci/private/push.sh.tpl
alexeagle: Yeah or we could read the `{{tags}}` file if it exists and just populate the `TAGS` array, then they'll be picked up in the loop above. there's not enough bats testing for this file so I'm scared to make that change right now

thesayyn: I agree. we could do that as a follow-up.

thesayyn (approved):
🤌🏻

PR #77, opened 2023-02-24
feat: demonstrate a replacement for go_image

alexeagle:


thesayyn (approved):
awesome

examples/go/README.md
```diff
@@ -0,0 +1,92 @@
...
+Now that we know it's `gcr.io/distroless/base` we can pull the same base image by adding to WORKSPACE:
+
+```
+load("@contrib_rules_oci//oci:pull.bzl", "oci_pull")
+
+oci_pull(
+    name = "distroless_base",
+    digest = "sha256:ccaef5ee2f1850270d453fdf700a5392534f8d1a8ca2acda391fbb6a06b81c86",
+    image = "gcr.io/distroless/base",
+    platforms = ["linux/amd64","linux/arm64"],
```
seh, line 49: When I run the following command, I see different digests for the AMD64 and ARM64 images:
```shell
docker manifest inspect gcr.io/distroless/base \
  | jq --raw-output \
       '.manifests[] | select(.platform.os == "linux") | {arch: .platform.architecture, digest}'
```
What allows you to specify one digest here and assume that the resulting image will work for both of these machine architectures?

thesayyn: crane has a digest command that lets you query the digest by running `crane digest <IMAGE>`.  alternatively, you could specify `tag` and `reproducible = False` attributes and replace the attributes according to the warning message oci_pull will print.

seh: This all makes more sense to me today. With Using a combination of _crane digest_, _docker manifest inspect_, and GCP's Web console to inspect the GCR entries, I now see how we're able to point at an image that contains a _manifest list_, with several entries, each for a different combination of operating system and machine architecture.

With the _docker manifest inspect_ command, it's not possible—or at least I couldn't figure out a way—to see the image digest for the image nominated by the tagged name. It dumps the manifest list, but doesn't tell you the digest of the image containing that manifest list. Using the _crane digest_ command closes that gap.

Thank you for the suggestion, @thesayyn.

PR #79, opened 2023-02-24
fix: use digest when loading image unto daemon

thesayyn:


oci/private/structure_test.bzl
```diff
@@ -24,9 +24,12 @@ exec "{st_path}" test {fixed_args} "$@"
 def _structure_test_impl(ctx):
     st_info = ctx.toolchains["@contrib_rules_oci//oci:st_toolchain_type"].st_info
 
-    default_image_tag = "{workspace}.local/{package}/{name}:latest".format(
-        workspace = ctx.workspace_name,
-        package = ctx.label.package.replace("/", "_"),
+    workspace = "default_workspace" if ctx.workspace_name == "__main__" else ctx.workspace_name
```
alexeagle, line 27: Instead of that special case how about just escaping whatever name? __foo__ is a legal workspace name

thesayyn: right... that'd be better. I'll fix it.

thesayyn: done

oci/private/structure_test.bzl
```diff
@@ -21,12 +21,20 @@ CMD = """\
 exec "{st_path}" test {fixed_args} "$@"
 """
 
+def _strip_illegal_chars(str):
```
alexeagle, line 24: "Strip" suggests throw away, how about escape?

Also I would just replace with one character

oci/private/structure_test.bzl
```diff
@@ -21,12 +21,20 @@ CMD = """\
...
+    return str.replace("_", "underscore").replace("/", "slash")
```
alexeagle, line 25: Are those the only illegal chars?

alexeagle (commented):
actually I think this is a regression.

previously you could write some user instructions like
`bazel build :some-image; podman run bazel/some-image:latest`
but now you can't predict it anymore

thesayyn:
we have `oci_tarball` for that and people should use that since images guaranteed to have tags you specify via `repotags` attribute.

alexeagle (approved):

PR #80, opened 2023-02-26
chore: point to js example in rules_js

alexeagle:
The one here is bound to go out-of-date.


thesayyn (approved):

alexeagle:
@thesayyn I think removing that CI step is fine since we have the same coverage in `structure_test`s - does that sound right to you?

thesayyn:
that'd probably be fine though we lost coverage for `oci_tarball`.  It could be better to make it a sh_test but if you don't have the time, I'd be fine with addressing it later after filing an issue.

PR #84, opened 2023-02-28
fix: include attachment in the runfiles

thesayyn:


cosign/private/attach.bzl
```diff
@@ -54,11 +54,11 @@ def _cosign_attach_impl(ctx):
     fixed_args = ["--repository", ctx.attr.repository]
 
     if ctx.attr.type == "sbom":
-        fixed_args.extend(["--sbom", ctx.file.attachment.path])
+        fixed_args.extend(["--sbom", ctx.file.attachment.short_path])
```
alexeagle, line 57: Why this change?

examples/attach/test.bash
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -o pipefail -o errexit -o nounset
 
+export HOME="$TEST_TMPDIR"
```
alexeagle, line 4: Should there be an added assertion?

cosign/private/attach.bzl
thesayyn: these go into runfiles where it needs to refer to the file by its short_path because the path contains output tree segments.  this never worked for generated files but worked for source files since those have identical short_path and path.

thesayyn: https://github.com/bazel-contrib/rules_oci/blob/982972a830655537dd1ff8f85f6815c52410e50b/cosign/private/sign.bzl#L60 uses short_path as both should

examples/attach/test.bash
thesayyn: I don't think we need one for this. it is there because cosign wants to read credentials which unrestricted at `bazel run`

cosign/private/attach.bzl
alexeagle: if it's for a runfiles lookup, then I think you need our `to_manifest_path` helper, short_path isn't always correct. e.g. https://github.com/aspect-build/rules_esbuild/pull/108 or https://github.com/aspect-build/rules_js/blob/0b4e900ea299ea6e77d45f1f2b5f03167bb7f3c0/js/private/js_image_layer.bzl#L134

thesayyn: I am not sure if we need this. I have added a test case where we sign an oci_image target from an external workspace.

examples/attach/BUILD.bazel
```diff
@@ -10,7 +17,7 @@ oci_image(
 
 cosign_attach(
     name = "attach",
-    attachment = "sbom.spdx",
+    attachment = ":sbom_generated.spdx",
     image = ":image",
     repository = "test",
     type = "sbom",
```
alexeagle, line 23: don't you need to use the external repo in one of these attributes to see a potential failure?

thesayyn: I added the test for signing. I'll add a test for this one as well.

thesayyn: done

alexeagle (approved):

PR #90, opened 2023-03-08
chore: remove contrib_ prefix from ruleset name

alexeagle:
We did this originally to conform with https://bazel.build/rules/deploying#hosting_and_naming_rules which says that shortnames like this only belong in the bazelbuild org. But on the Bazel Central Registry, this hasn't been followed, there are rules_cue and rules_kustomize for example. We may as well have the shorter name too.


kormide (approved):

alexeagle:
I'll do the publish to BCR now

PR #92, opened 2023-03-11
fix: rust does not have a 2022 edition

bcmyers:
Small fix. 

Rust does not have a 2022 edition. It's current editions are 2015, 2018 and 2021.

See: https://doc.rust-lang.org/edition-guide/


thesayyn (approved):

thesayyn:
Thanks!

PR #94, opened 2023-03-12
feat: support the driver argument of container-test

alexeagle:
This allows many tests to run on a platform with no container runtime, like the BCR presubmit

### DIFF
--- a/oci/private/tarball.bzl
+++ b/oci/private/tarball.bzl
@@ -87,7 +87,9 @@ def _tarball_impl(ctx):
 
     tarball = _tar_tarball(ctx, blobs, manifest)
 
-    return DefaultInfo(files = depset([tarball]))
+    return [
+        DefaultInfo(files = depset([tarball])),
+    ]
 
 oci_tarball = rule(
     implementation = _tarball_impl,
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 docs/cosign*.md
 docs/image*.md
+docs/structure*.md
 docs/pull.md
 docs/push.md
 docs/tarball.md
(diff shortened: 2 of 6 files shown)


thesayyn:
image attribute already accepts a single file. maybe instead of introducing a new attribute we just make the image attribute accept both oci_tarball and oci_image ?

alexeagle:
Hm yeah lemme figure out the providers so I can still properly constrain legal values and distinguish which it is

alexeagle:
okay @thesayyn PTAL

oci/private/image.bzl
```diff
@@ -175,4 +178,5 @@ oci_image = rule(
         "@rules_oci//oci:registry_toolchain_type",
         "@aspect_bazel_lib//lib:jq_toolchain_type",
     ],
+    provides = [OciImageInfo],
```
alexeagle, line 181: I wonder if `oci_image_index` should also provide this?

thesayyn: I feel like we can do this without providers. simply looking into the extension?

thesayyn: if the file ends with `.tar` then we assume it's an oci_tarball?

alexeagle: okay

oci/private/tarball.bzl
```diff
@@ -87,7 +87,9 @@ def _tarball_impl(ctx):
 
     tarball = _tar_tarball(ctx, blobs, manifest)
 
-    return DefaultInfo(files = depset([tarball]))
```
thesayyn, line 90: nit: revert

thesayyn (approved):

oci/private/tarball.bzl
alexeagle: Oops, I turned on "Require conversation resolution before merging"  so that doesn't happen again.

PR #101, opened 2023-03-15
chore(bzlmod): disable windows for now

alexeagle:
See #97

PR #109, opened 2023-03-17
chore: add e2e test for crane as registry

thesayyn:
closes #101

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,5 +1,5 @@
 bcr_test_module:
-  module_path: "e2e/smoke"
+  module_path: "e2e/crane_as_registry"
   matrix:
     platform: ["debian10", "macos", "ubuntu2004", "windows"]
   tasks:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,13 +24,12 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         folder:
-          [
-            ".",
-            "e2e/custom_registry",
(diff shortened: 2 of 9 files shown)


.bcr/presubmit.yml
```diff
@@ -1,5 +1,5 @@
 bcr_test_module:
-  module_path: "e2e/smoke"
+  module_path: "e2e/crane_as_registry"
```
alexeagle, line 2: I don't think we should change this, we use "smoke" in all our rulesets as the e2e that BCR presubmit runs. Can we just green up that one on windows?

thesayyn: I could switch places. make smoke test use crane and the new test cover zot

thesayyn: on second thought, that would make us recommend crane as the default registry, not sure if that's what we want. i don't want users memory to be filled with big blobs

.github/workflows/ci.yaml
```diff
@@ -23,7 +23,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        folder: [".", "e2e/custom_registry", "e2e/wasm", "e2e/smoke"]
+        folder:
+          [
```
alexeagle, line 27: nit: I think a yaml multi-line list should just use the hyphen syntax like
```
folder:
  - .
  - e2e/...
  - ...
```

alexeagle (approved):

.bcr/presubmit.yml
thesayyn: done

PR #114, opened 2023-03-23
refactor: rename x86_64 to amd64

thesayyn:
Allows us to use `repo_utils.platform` and matches our other repositories.

PR #115, opened 2023-03-23
feat: implement auth for oci_pull

thesayyn:
Needs #114 to go green

### DIFF
--- a/e2e/custom_registry/WORKSPACE
+++ b/e2e/custom_registry/WORKSPACE
@@ -26,13 +26,15 @@ oci_register_toolchains(
 load("@rules_oci//oci:pull.bzl", "oci_pull")
 
 oci_pull(
-    name = "distroless_static",
-    digest = "sha256:c3c3d0230d487c0ad3a0d87ad03ee02ea2ff0b3dcce91ca06a1019e07de05f12",
-    image = "gcr.io/distroless/static",
+    name = "debian",
+    image = "index.docker.io/library/debian",
     platforms = [
-        "linux/amd64",
         "linux/arm64",
+        "linux/amd64",
     ],
+    # Don't make a debian_unpinned repo and print a warning about the tag
+    reproducible = False,
+    tag = "latest",
 )
 
 ############################################
--- a/oci/pull.bzl
+++ b/oci/pull.bzl
@@ -48,6 +48,8 @@ def _file_exists(rctx, path):
     result = rctx.execute(["stat", path])
     return result.return_code == 0
 
+# Path of the auth file is determined by the order described here;
+# https://github.com/google/go-containerregistry/tree/main/pkg/authn#tldr-for-consumers-of-this-package
 def _get_auth_file_path(rctx):
     # this is the standard path where registry credentials are stored
     config_path = "{}/.docker/config.json".format(rctx.os.environ["HOME"])
@@ -65,6 +67,11 @@ def _get_auth_file_path(rctx):
         XDG_RUNTIME_DIR = rctx.os.environ["XDG_RUNTIME_DIR"]
 
     config_path = "{}/containers/auth.json".format(XDG_RUNTIME_DIR)
+
+    # podman support overriding the standard path for the auth file via this special environment variable.
+    # https://docs.podman.io/en/latest/markdown/podman-login.1.html#authfile-path
+    if "REGISTRY_AUTH_FILE" in rctx.os.environ:
+        config_path = rctx.os.environ["REGISTRY_AUTH_FILE"]
 
     if _file_exists(rctx, config_path):
         return config_path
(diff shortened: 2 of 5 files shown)


oci/pull.bzl
```diff
@@ -228,7 +386,12 @@ oci_pull_rule = repository_rule(
         "image": attr.string(doc = "The name of the image we are fetching, e.g. gcr.io/distroless/static", mandatory = True),
         "identifier": attr.string(doc = "The digest or tag of the manifest file", mandatory = True),
         "platform": attr.string(doc = "platform in `os/arch` format, for multi-arch images"),
+        "toolchain_name": attr.string(default = "oci", doc = "Value of name attribute to the oci_register_toolchains call in the workspace."),
     },
+    environ = [
```
alexeagle, line 391: comment about why these are needed?

oci/pull.bzl
```diff
@@ -37,24 +37,164 @@ oci_image(
...
+        auth = json.decode(auth_raw)
+        pattern = {
+            "type": "pattern",
+            "pattern": "Bearer <password>",
+            "password": auth["token"],
+        }
+
+    return pattern
+
+def _auth_basic(rctx, registry, repository, identifier):
```
alexeagle, line 103: @gregmagolan do you think it's worth sharing code from our npm repository rule via something in bazel_lib?

oci/pull.bzl
```diff
@@ -37,24 +37,164 @@ oci_image(
...
+        XDG_RUNTIME_DIR = rctx.os.environ["XDG_RUNTIME_DIR"]
+
+    config_path = "{}/containers/auth.json".format(XDG_RUNTIME_DIR)
+
+    if _file_exists(rctx, config_path):
+        return config_path
+
+    return None
+
+def _auth_anonymous(rctx, registry, repository, identifier):
```
alexeagle, line 74: this is specific to dockerhub right? I know the function is basically no-op when the registry isn't `index.docker.io` but to me that suggests that the caller should be the one checking the registry argument and not call this function.

(unless you think we'll add detection for other registries in here)

oci/pull.bzl
```diff
@@ -37,24 +37,164 @@ oci_image(
...
 load("@aspect_bazel_lib//lib:paths.bzl", "BASH_RLOCATION_FUNCTION")
+load("@aspect_bazel_lib//lib:base64.bzl", "base64")
+load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
+
+def _strip_host(url):
+    # TODO: a principled way of doing this
+    return url.replace("http://", "").replace("https://", "").replace("/v1/", "")
+
+def _file_exists(rctx, path):
+    result = rctx.execute(["stat", path])
```
alexeagle, line 48: feels like a portability concern, isn't there a way to detect this in the starlark interpreter without any subprocess?

alexeagle (commented):
thanks for doing this one separate from credential helper auth

oci/pull.bzl
thesayyn: Yes. we'll probably add other registries here as people come up with new ones. Unfortunately, we can't generalize this because we can't read the WWW-Authenticate header.

oci/pull.bzl
```diff
@@ -73,13 +216,35 @@ WARNING: fetching from %s without an integrity hash. The result will not be cach
...
-        return None
+def _crane_label(rctx):
+    return Label("@{}_crane_{}//:crane".format(rctx.attr.toolchain_name, repo_utils.platform(rctx)))
+
+def _download_manifest(rctx, identifier, output):
+    _download(rctx, identifier, output, "manifests")
+    bytes = rctx.read(output)
+    manifest = json.decode(bytes)
+
+    if manifest["schemaVersion"] == 1:
```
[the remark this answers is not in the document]
thesayyn: that'd be when the image indeed has schema version 1 which neither we nor any recent tooling support. it's been deprecated for years and nobody uses it. 

about it is being unset, it is not likely to be seen as https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions and https://docs.docker.com/registry/spec/manifest-v2-2/ mandates it. 

Thinking more about this; we should add a friendly message when we encounter a schemaVersion being unset or set to 1 even after the fallback. However, we should do that in a follow-up as this is out of scope.

oci/pull.bzl
thesayyn: nothing I could find. afaik nothing in the rctx can tell whether a file exists.

thesayyn: maybe @gregmagolan knows as he has done a lot of repository rule stuff

gregmagolan: Calling `rctx.path()` on a non-existent file will just fail. Although you also have to be careful with using `exists` since calling `rctx.path` on a dynamic label (a label that was _not_ passed in via an attribute but rather constructed dynamically) will cause the repository rule to restart so its best to do all of these dynamical label lookups at the beginning of the repository rule before you do any work that takes time or else you'll end up doing that work multiple times to fetch the repository

oci/pull.bzl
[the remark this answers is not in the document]
alexeagle: I think my request for now is just that the error message should start with the actual condition we observed, and then give our interpretation of it. That way, if we interpret incorrectly, the user has a chance of figuring that out.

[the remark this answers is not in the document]
thesayyn: That sounds good. I'll do that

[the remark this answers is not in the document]
thesayyn: done

oci/pull.bzl
thesayyn: yeah. It would be great to put that helper somewhere in bazel-lib.

e2e/custom_registry/WORKSPACE
```diff
@@ -26,13 +26,15 @@ oci_register_toolchains(
 load("@rules_oci//oci:pull.bzl", "oci_pull")
 
 oci_pull(
-    name = "distroless_static",
-    digest = "sha256:c3c3d0230d487c0ad3a0d87ad03ee02ea2ff0b3dcce91ca06a1019e07de05f12",
-    image = "gcr.io/distroless/static",
+    name = "debian",
+    image = "index.docker.io/library/debian",
```
alexeagle, line 30: I think the `docker` tool at least allows you to omit the `index.` part of the image name, right? Could be a separate PR.

PR #118, opened 2023-03-24
chore(bzlmod): add trivial pass-through module extension for oci_pull

alexeagle:
However it trips over the labels in our generated build files making assumptions about repository names, needs some fixing

### DIFF
--- a/oci/extensions.bzl
+++ b/oci/extensions.bzl
@@ -23,7 +23,7 @@ Base name for generated repositories, allowing more than one set of toolchains t
 Overriding the default is only permitted in the root module.
 """, default = "oci"),
     "crane_version": attr.string(doc = "Explicit version of crane.", mandatory = True),
-    "zot_version": attr.string(doc = "Explicit version of zot.", mandatory = True),
+    "zot_version": attr.string(doc = "Explicit version of zot. If not supplied, then only crane is used."),
 })
 
 def _oci_extension(module_ctx):
--- a/oci/extensions.bzl
+++ b/oci/extensions.bzl
@@ -1,6 +1,21 @@
 "extensions for bzlmod"
 
 load(":repositories.bzl", "oci_register_toolchains")
+load(":pull.bzl", "oci_pull")
+
+# TODO: it sucks that the API of the oci_pull macro has to be repeated here.
+pull = tag_class(attrs = {
+    "name": attr.string(doc = "Name of the generated repository"),
+    "image": attr.string(doc = """the remote image without a tag, such as gcr.io/bazel-public/bazel"""),
+    "platforms": attr.string_list(doc = """for multi-architecture images, a dictionary of the platforms it supports
+            This creates a separate external repository for each platform, avoiding fetching layers."""),
+    "digest": attr.string(doc = """the digest string, starting with "sha256:", "sha512:", etc.
+            If omitted, instructions for pinning are provided."""),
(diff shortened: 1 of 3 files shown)


oci/extensions.bzl
```diff
@@ -1,6 +1,21 @@
...
+    "image": attr.string(doc = """the remote image without a tag, such as gcr.io/bazel-public/bazel"""),
+    "platforms": attr.string_list(doc = """for multi-architecture images, a dictionary of the platforms it supports
+            This creates a separate external repository for each platform, avoiding fetching layers."""),
+    "digest": attr.string(doc = """the digest string, starting with "sha256:", "sha512:", etc.
+            If omitted, instructions for pinning are provided."""),
+    "tag": attr.string(doc = """a tag to choose an image from the registry.
+            Exactly one of `tag` and `digest` must be set.
+            Since tags are mutable, this is not reproducible, so a warning is printed."""),
+    "reproducible": attr.bool(doc = """Set to False to silence the warning about reproducibility when using `tag`.""", default = True),
+})
```
kormide, line 18: Can you just share the `attrs` object with the rule? Then add any extra attributes for the macro (if there is one).

kormide: https://github.com/aspect-build/rules_js/blob/main/npm/extensions.bzl#L157

alexeagle: I tried that at first, however this calls a wrapper macro where the argument are nearly non-overlapping. For instance the underlying rule has a `platform` attribute, but the macro has a loop so it accepts `platforms` instead.

alexeagle:
I made an example to test this, and there's a bug. Problem here is

```
ERROR: /home/alexeagle/Projects/bazel-examples/oci_go_image/BUILD.bazel:34:10: no such target '@rules_oci~override~oci~distroless_base//:image': target 'image' not declared in package '' defined by /shared/cache/bazel/user_base/359b45a24cf55cc22380e941fe3bcf54/external/rules_oci~override~oci~distroless_base/BUILD.bazel
```

which is caused by https://github.com/bazel-contrib/rules_oci/blob/main/oci/pull.bzl#L246

I wanted to have nice syntax sugar for users to choose a base image like just `base = "@distroless_static"` but one option here is to give that up and require a label like `@distroless_static//:image` - then we don't need to know the repo name here to create a target with a matching name.

kormide:
> I made an example to test this, and there's a bug. Problem here is
> 
> ```
> ERROR: /home/alexeagle/Projects/bazel-examples/oci_go_image/BUILD.bazel:34:10:
> no such target '@rules_oci~override~oci~distroless_base//:distroless_base': target 'distroless_base' not declared in package '' 
> defined by /shared/cache/bazel/user_base/359b45a24cf55cc22380e941fe3bcf54/external/rules_oci~override~oci~distroless_base/BUILD.bazel
> ```
> 
> which is caused by https://github.com/bazel-contrib/rules_oci/blob/main/oci/pull.bzl#L246
> 
> I wanted to have nice syntax sugar for users to choose a base image like just `base = "@distroless_static"` but one option here is to give that up and require a label like `@distroless_static//:image` - then we don't need to know the repo name here to create a target with a matching name.

I'm missing some pieces here. So you want to be able to pass in `base = "@distroless_static"` into `oci_image`? As in you want to pass a string rather than a label and have it do the equivalent of appending the `//:image` target?

alexeagle:
okay the downstream example is green now!

kormide (approved):

PR #123, opened 2023-03-27
refactor: make pinning instructions a fail message

thesayyn:
<s>BEHAVIOR CHANGE: `reproducible = True` is used to stop the execution of the repository rule. it's not the case anymore

the reason for this is that to fetch the manifest, we have to be inside a repository_rule and not a macro, if we stop the execution early in the macro then users wouldn't get the warning message. we could use `fail` instead of `print` to stop the execution. However, I think this matches the behavior of other rules like `http_archive`</s>

the reason why delta is big 
- <s>replacing usage of `sha256sum` with `coreutils`</s>
- removes `pin_tag` rule with `oci_alias` which is responsible for printing the warning.
- replaces bazel runnable target that prints the instructions with buildozer command

### DIFF
--- a/oci/pull.bzl
+++ b/oci/pull.bzl
@@ -412,16 +412,6 @@ alias(
 )
 """
 
-def _coreutils_label(rctx):
-    return Label("@coreutils_{}//:coreutils".format(repo_utils.platform(rctx)))
-
-def _sha256(rctx, coreutils_label, path):
-    result = rctx.execute([coreutils_label, "hashsum", "--sha256", path])
-    if result.return_code:
-        msg = "hashsum failed: \nSTDOUT:\n%s\nSTDERR:\n%s" % (result.stdout, result.stderr)
-        fail(msg)
-    return result.stdout.split(" ", 1)[0]
-
 def _oci_alias_impl(rctx):
     if rctx.attr.platforms and rctx.attr.single_platform:
         fail("Only one of 'platforms' or 'single_platform' may be set")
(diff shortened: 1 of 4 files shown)


thesayyn:
ping @alexeagle

alexeagle (commented):
Could you answer "why" in the PR description?

oci/pull.bzl
```diff
@@ -418,97 +417,86 @@ alias(
...
 
+_SINGLE_PLATFORM_IMAGE_ALIAS = """\
+alias(
+    name = "{target_name}",
+    actual = "@{original}",
+    visibility = ["//visibility:public"],
+)
+"""
+
+def _coreutils_label(rctx):
```
alexeagle, line 428: I don't understand why the delta is so large, the PR description just says you're changing an error to a warning. Is this related?

thesayyn: yes, it replaces `sha256sum` with the coreutils equivalent.

thesayyn:
> Could you answer "why" in the PR description?

done. I could revert the `doesn't stop the execution part as well.`

oci/pull.bzl
```diff
@@ -412,97 +411,78 @@ alias(
...
-jq(
-    name = "platforms",
-    srcs = ["manifest_list.json"],
-    filter = "[(.manifests // [])[] | .platform | .os + \\"/\\" + .architecture]",
-    # Print without newlines because it's too hard to indent that to fit under the generated
-    # starlark code below.
-    args = ["--compact-output"],
-)
+        # buildifier: disable=print
+        fail("""
```
alexeagle, line 449: I am really confused now, the PR description says we are making it a warning message, but here we `fail`.

Do you mind just making a new PR that is minimal and doesn't have things crossed out in the PR description?

thesayyn: I made it fail to keep the changes to the PR minimal.

PR #131, opened 2023-03-29
feat: support credStore

thesayyn:
PR also introduces auth/token caching which is needed otherwise the credential helpers are called repetitively which is not suboptimal. 

I've done my best to keep changes small.


oci/private/auth_config_locator.bzl
```diff
@@ -41,7 +41,6 @@ def _get_auth_file_path(rctx):
 
 def _oci_auth_config_locator_impl(rctx):
     config_path = _get_auth_file_path(rctx)
-    config_path = None
```
thesayyn, line 44: https://github.com/bazel-contrib/rules_oci/pull/150

oci/pull.bzl
```diff
@@ -218,6 +271,17 @@ Falling back to using `curl`. See https://github.com/bazelbuild/bazel/issues/178
 
     return manifest, len(bytes)
 
+def _create_downloader(rctx):
+    state = {
+        "config": json.decode(rctx.read(rctx.attr.config)),
```
alexeagle, line 276: should there be some error handling in case you point us to a wrong file or it isn't valid json?

alexeagle (approved):
I didn't read through the new e2e but I'm happy it's there

oci/pull.bzl
thesayyn: json.decode throws in invalid json case. for other cases, we check if the key we are looking for exists in the json file. so no for now.

PR #133, opened 2023-03-29
feat: allow specifying protocol

thesayyn:
currently, we assume that registries are always `https` which is not true especially when testing or running local registry instances.  this seems like the easiest way to do this but I am open to other ideas. i also thought of adding `insecure_registry` attribute but seemed overkill


e2e/custom_registry/WORKSPACE
```diff
@@ -27,7 +27,7 @@ load("@rules_oci//oci:pull.bzl", "oci_pull")
 
 oci_pull(
     name = "debian",
-    image = "index.docker.io/library/debian",
+    image = "https://index.docker.io/library/debian",
```
alexeagle, line 30: where do users typically get this string? I think we should optimize for easy copy-paste.

For that matter, I've been doing some rules_docker migrations, and they have separate `registry` and `repository` attributes. Maybe we should optimize for the migration case, if that API was just as good

alexeagle: agreed to have separate attributes on the underlying repository rule, but give a wrapper macro with syntax sugar for various shortforms of `image`
can be in a followup PR

alexeagle (approved):

PR #143, opened 2023-04-01
fix: replace crane fallback with curl

thesayyn:
replaces crane fallback with curl fallback.

### DIFF
--- a/oci/private/download.bzl
+++ b/oci/private/download.bzl
@@ -1,0 +1,134 @@
+"Downloader functions "
+
+def _auth_to_header(url, auth):
+    for auth_url in auth:
+        if auth_url == url:
+            auth_val = auth[auth_url]
+
+            # TODO: basic type
+            if auth_val["type"] == "pattern":
+                token = auth_val["pattern"].replace("<password>", auth_val["password"])
+                return [
+                    "--header",
+                    "Authorization: {}".format(token),
+                ]
+    return []
+
+def _debug(message):
+    # Change to true when debugging
+    if False:
+        # buildifier: disable=print
+        print(message)
+
+# Bazel-like downloader using curl
+#  Features
+#   - Can set custom headers
+#   - Support for http2 out of the box
+#  Supports
+#   - bazel repository cache. doesn't do re-fetches if repository rule gets invalidated
+#   - authorization
+#  Caveats
+#   - Doesn't support --experimental_downloader_config. (can not read the flag therefore can't support it out of the box. though can be supported by introducing an attribute.)
+#   - Doesn't support other http related bazel flags.
+#   - Support for netrc. TODO: https://curl.se/docs/manpage.html#--netrc-file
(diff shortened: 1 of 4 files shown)


oci/private/download.bzl
```diff
@@ -0,0 +1,134 @@
...
+        integrity = integrity,
+        sha256 = sha256,
+    )
+
+    return cache_it
+
+# A dummy function that uses bazel downloader.
+#  Caveats
+#   - Doesn't support setting custom headers
+def _bazel_download(
```
alexeagle, line 109: I bet this could be a lot shorter, maybe just `_bazel_download = lambda *args: rctx.download(*args)` or even `_bazel_download = rctx.download`
In those cases maybe you don't even need to export a `bazel` symbol from this file's `struct` and just let the caller do that.

alexeagle (commented):
Very good idea to remove crane, it's much heavier than we actually need for setting headers.

This might be broken on windows? I think you need a magic branch to trigger windows CI. We should change that so a magic tag or something added to the PR also triggers it.

oci/private/download.bzl
thesayyn: we need to list extra arguments here to have matching signatures.

thesayyn:
> Very good idea to remove crane, it's much heavier than we actually need for setting headers.
> 
> This might be broken on windows? I think you need a magic branch to trigger windows CI. We should change that so a magic tag or something added to the PR also triggers it.

curl is shipped with windows as well. it could fail though due to an older version of curl being used. I'll do the windows thing in a different PR.

oci/private/download.bzl
```diff
@@ -0,0 +1,121 @@
...
+        print(message)
+
+# Bazel-like downloader using curl
+#  Features
+#   - Can set custom headers
+#   - Support for http2 out of the box
+#  Supports
+#   - bazel repository cache. doesn't do re-fetches if repository rule gets invalidated
+#   - authorization
+#  Caveats
```
alexeagle, line 30: Thank you!

oci/private/download.bzl
```diff
@@ -0,0 +1,121 @@
...
+    rctx.file(headers_output_path)
+
+    result = rctx.execute(command)
+
+    _debug("""\nSTDOUT\n{}\nSTDERR\n{}""".format(result.stdout, result.stderr))
+
+    if result.return_code != 0 and allow_fail:
+        return struct(success = False)
+    elif result.return_code != 0 and not allow_fail:
+        fail("Failed to fetch {} {}".format(url, result.stderr))
```
alexeagle, line 92: Don't we need to print stderr here too?

alexeagle (approved):

oci/private/download.bzl
thesayyn: yeah. i'll fix it

thesayyn: actually, this is intentional. curl always prints errors to stderr. stdout is just empty.

PR #145, opened 2023-04-05
fix: prevent spamming on missing config file

thesayyn:
Moved the logic for finding the auth file into its own file and repository file so that the result can be shared between oci_pull targets. Also, changed `print` to `rctx.execute` to only print when the repository is not cached as opposed to running at analysis time. 

NOTE: rctx.execute has much nicer output


oci/pull.bzl
```diff
@@ -403,16 +358,8 @@ oci_pull_rule = repository_rule(
         "platform": attr.string(doc = "platform in `os/arch` format, for multi-arch images"),
         "target_name": attr.string(doc = "Name given for the image target, e.g. 'image'", mandatory = True),
         "toolchain_name": attr.string(default = "oci", doc = "Value of name attribute to the oci_register_toolchains call in the workspace."),
+        "config": attr.label(doc = "Label to a config.json file", default = "@oci_config//:config.json"),
```
alexeagle, line 361: docstring could say "internal use only" or "it should be generated by XXX" so the docs/*.md can help someone understand the layering of macros/rules

oci/pull.bzl
```diff
@@ -516,6 +463,7 @@ pin_tag = repository_rule(
         "image": attr.string(doc = "The name of the image we are fetching, e.g. `gcr.io/distroless/static`", mandatory = True),
         "tag": attr.string(doc = "The tag being used, e.g. `latest`", mandatory = True),
         "toolchain_name": attr.string(default = "oci", doc = "Value of name attribute to the oci_register_toolchains call in the workspace."),
```
alexeagle, line 465: nit: seems like it's time to extract a dict constant for the attrs that are common between alias_target and the one above

alexeagle, line 465: nit: seems like it's time to extract a dict constant for the attrs that are common between alias_target and the one above

oci/private/config.bzl
```diff
@@ -0,0 +1,66 @@
...
+    if _file_exists(rctx, config_path):
+        return config_path
+
+    return None
+
+def _oci_config_locator_impl(rctx):
+    config_path = _get_auth_file_path(rctx)
+    config_path = None
+    if not config_path:
+        # rctx.execute is cached between bazel invocations. prefer it over print
```
alexeagle, line 41: the spam problem was in a single bazel invocation, so I don't think this is actually a reason to avoid print. We were simply evaluating that starlark statement too many times in a single invocation.

I guess `echo` is still fine, though it feels unusual

oci/pull.bzl
thesayyn: i will do that as a follow-up

oci/private/config.bzl
thesayyn: output is nicer too

```WARNING: Ignoring JAVA_HOME, because it must point to a JDK, not a JRE.
Starting local Bazel server and connecting to it...
WARNING: Could not find the `$HOME/.docker/config.json` and `$XDG_RUNTIME_DIR/containers/auth.json` file. 
 Running one of `podman login`, `docker login`, `crane login` may help.
INFO: Analyzed 4 targets (51 packages loaded, 303 targets configured).
INFO: Found 4 targets...
INFO: Elapsed time: 22.658s, Critical Path: 2.06s
INFO: 7 processes: 5 internal, 2 local.
INFO: Build completed successfully, 7 total actions

```

oci/private/config.bzl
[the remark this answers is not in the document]
thesayyn: renamed to auth_config_locator

alexeagle (approved):

PR #153, opened 2023-04-13
support pulling from ghcr.io

LavaToaster:


thesayyn:
it would be great to have an oci_pull target in the WORKSPACE which we fetch as part of CI as a smoke test.  https://github.com/bazel-contrib/rules_oci/blob/eae8d198ab8f95797402f0692094d43a2a62f35f/oci/tests/BUILD.bazel#L42-L48
https://github.com/bazel-contrib/rules_oci/blob/eae8d198ab8f95797402f0692094d43a2a62f35f/WORKSPACE#L117-L134

thesayyn (approved):

PR #154, opened 2023-04-13
feat: allow stamping in repository name

### LINKED ISSUE #46: oci_push --repository and --repotags flags should accept file paths
Useful for stamping, for example

alexeagle:
This could be used to take the username of the person running the build.

Fixes #46

### DIFF
--- a/examples/push/BUILD.bazel
+++ b/examples/push/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
 load("//oci:defs.bzl", "oci_image", "oci_image_index", "oci_push")
 load(":stamp_tags.bzl", "stamp_tags")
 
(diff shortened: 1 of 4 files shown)


oci/private/push.bzl
```diff
@@ -70,12 +70,14 @@ _attrs = {
         doc = "Label to an oci_image or oci_image_index",
         mandatory = True,
     ),
-    "repository": attr.string(
+    "repository": attr.label(
         doc = """\
-        Repository URL where the image will be signed at, e.g.: `index.docker.io/<user>/image`.
+        a .txt file containing one line, which is the Repository URL where the image will be signed
```
thesayyn, line 75:
```suggestion
        a .txt file containing one line, which is the Repository URL where the image will be pushed
```

thesayyn: also, does it have to be a `.txt` file?

examples/push/BUILD.bazel
```diff
@@ -40,8 +55,8 @@ stamp_tags(
 oci_push(
     name = "push_image_index",
     image = ":image_index",
-    repository = "index.docker.io/<ORG>/image",
-    repotags = ":stamped",
+    repository_file = ":stamped_repo",
```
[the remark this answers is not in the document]
alexeagle: ah, because we can't detect on the attribute type this time. The other ones use a trick in the macro like `types.is_list` or `types.is_dict`. But a string could be a label or it could be a repository name and the macro can't tell.

oci/private/push.bzl
alexeagle: no, that's artificial, will remove

examples/push/BUILD.bazel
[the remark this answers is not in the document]
thesayyn: this should be possible now that we have  unified  `repotags` which is a list

thesayyn:
@alexeagle this is unblocked.

alexeagle:
I believe this was fixed as a side-effect of the repository name moving into the `repotags` attribute which is already stampable.

seh:
> I believe this was fixed as a side-effect of the repository name moving into the `repotags` attribute which is already stampable.

Per my https://github.com/bazel-contrib/rules_oci/issues/46#issuecomment-1718150022, we lost this ability.

PR #155, opened 2023-04-14
feat: allow stamping labels/annotations

### LINKED ISSUE #152: Stamp labels/annotations
The image label/annotations should be stampable.

For example, Datadog (monitoring) can use this information to link to a VCS revision.

In rules_docker, labels are stampable.

alexeagle:
Fixes #152

### DIFF
--- a/examples/labels/BUILD.bazel
+++ b/examples/labels/BUILD.bazel
@@ -1,20 +1,31 @@
 load("//oci:defs.bzl", "oci_image", "structure_test")
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
-load("@bazel_skylib//rules:write_file.bzl", "write_file")
 
 pkg_tar(
     name = "app",
     srcs = ["test.bash"],
 )
 
-write_file(
+# We don't actually work with JSON, but the jq rule happens to be stamp-aware so we can (ab)use it
+# to put (non-deterministic) version information in our labels/annotations.
+# You'd typically use that only when building release artifacts, not for testing, as it causes
+# action cache misses.
+jq(
     name = "labels",
+    srcs = [],
     out = "labels.txt",
-    content = [
-        # TODO(alex): show how this can be produced by a stamp-aware rule so the 0.0.0 can be replaced
-        "org.opencontainers.image.version=0.0.0",
-        "org.opencontainers.image.source=https://github.com/bazel-contrib/rules_oci",
-    ],
+    args = ["--raw-output"],
+    filter = """\
+        # See docs for jq rule about stamping:
+        # https://docs.aspect.build/rules/aspect_bazel_lib/docs/jq
+        $ARGS.named.STAMP as $stamp |
+
+        # Use the value of --embed_label under --stamp, otherwise use a deterministic constant
+        # value to ensure cache hits for actions that depend on this.
+        "org.opencontainers.image.version=" + ($stamp.BUILD_EMBED_LABEL // "0.0.0"),
+        "org.opencontainers.image.source=https://github.com/bazel-contrib/rules_oci"
+    """,
 )
 
 oci_image(
--- a/examples/labels/BUILD.bazel
+++ b/examples/labels/BUILD.bazel
@@ -1,9 +1,20 @@
 load("//oci:defs.bzl", "oci_image", "structure_test")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
 
 pkg_tar(
     name = "app",
     srcs = ["test.bash"],
+)
+
+write_file(
+    name = "labels",
+    out = "labels.txt",
+    content = [
+        # TODO(alex): show how this can be produced by a stamp-aware rule so the 0.0.0 can be replaced
+        "org.opencontainers.image.version=0.0.0",
+        "org.opencontainers.image.source=https://github.com/bazel-contrib/rules_oci",
+    ],
 )
 
 oci_image(
@@ -14,10 +25,7 @@ oci_image(
     }),
     base = "//examples:base",
     cmd = ["test.sh"],
-    labels = {
-        "org.opencontainers.image.version": "xxx",
-        "org.opencontainers.image.source": "https://github.com/bazel-contrib/rules_oci",
-    },
+    labels = ":labels",
     os = "linux",
     tars = ["app.tar"],
 )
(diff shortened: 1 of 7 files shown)


oci/private/image.sh.tpl
```diff
@@ -98,6 +98,17 @@ for ARG in "$@"; do
         (--env=*\${*}* | --env=*\$*) ENV_EXPANSIONS+=(${ARG#--env=}) ;;
         (--output=*) OUTPUT="${ARG#--output=}" ;;
         (--workdir=*) WORKDIR="${ARG#--workdir=}" ;;
+        (--labels-file=*)
+          while LANG=C IFS= read -r in || [ -n "$in" ]; do
+            echo "add $in"
```
thesayyn, line 103:
```suggestion

```

thesayyn (approved):

oci/private/image.sh.tpl
alexeagle: nice catch :)

oci/private/image.sh.tpl
```diff
@@ -98,6 +98,17 @@ for ARG in "$@"; do
         (--env=*\${*}* | --env=*\$*) ENV_EXPANSIONS+=(${ARG#--env=}) ;;
         (--output=*) OUTPUT="${ARG#--output=}" ;;
         (--workdir=*) WORKDIR="${ARG#--workdir=}" ;;
+        (--labels-file=*)
+          while LANG=C IFS= read -r in || [ -n "$in" ]; do
```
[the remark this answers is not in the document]
alexeagle: it forces some programs to understand this to be the locale, rather than some other default they might choose, which in turn can cause non-utf8 encodings. I don't think we actually need this in practice, since we are not calling any programs in here. will remove.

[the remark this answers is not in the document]
thesayyn: gotcha. we can always add it later when there's an issue that needs this.

examples/labels/BUILD.bazel
```diff
@@ -1,11 +1,33 @@
...
+        # See docs for jq rule about stamping:
+        # https://docs.aspect.build/rules/aspect_bazel_lib/docs/jq
+        $ARGS.named.STAMP as $stamp |
+
+        # Use the value of --embed_label under --stamp, otherwise use a deterministic constant
+        # value to ensure cache hits for actions that depend on this.
+        "org.opencontainers.image.version=" + ($stamp.BUILD_EMBED_LABEL // "0.0.0"),
+        "org.opencontainers.image.source=https://github.com/bazel-contrib/rules_oci"
+    """,
+)
```
seh, line 29: As an alternative to relying on _jq_, but instead relying on other shell tools like _grep_ and _touch_, I wrote the following  a few weeks ago:
```bazel
def _stamped_strings_impl(ctx):
    output_file = ctx.outputs.out
    if not output_file:
        output_file = ctx.actions.declare_file("{}_stamped.txt".format(ctx.label.name))
    seen_required_keys = dict()
    for k in ctx.attr.required_keys:
        if k in ctx.attr.optional_keys:
            fail("Required stamp key \"{}\" is also cited as optional".format(k))
        if k in seen_required_keys:
            fail("Duplicate required stamp key: \"{}\"".format(k))
        seen_required_keys[k] = None
    required_key_file = ctx.actions.declare_file("{}_required_stamp_keys".format(ctx.label.name))
    ctx.actions.write(
        output = required_key_file,
        content = "\n".join(ctx.attr.required_keys) + "\n",
    )
    optional_key_mapping_file = ctx.actions.declare_file("{}_optional_stamp_keys".format(ctx.label.name))
    ctx.actions.write(
        output = optional_key_mapping_file,
        content = "\n".join([
            "{}:{}".format(k, v)
            for k, v in ctx.attr.optional_keys.items()
        ]) + "\n",
    )
    ctx.actions.run(
        executable = ctx.executable._collect_stamped_values,
        arguments = [
            required_key_file.path,
            optional_key_mapping_file.path,
            output_file.path,
            ctx.info_file.path,
            ctx.version_file.path,
        ],
        inputs = [
            required_key_file,
            optional_key_mapping_file,
            ctx.info_file,
            ctx.version_file,
        ],
        outputs = [output_file],
        mnemonic = "CollectStampedValues",
        progress_message = "Collecting stamped values for {}".format(ctx.label.name),
    )
    return [
        DefaultInfo(
            files = depset(direct = [output_file]),
        ),
    ]

stamped_strings = rule(
    implementation = _stamped_strings_impl,
    attrs = {
        "_collect_stamped_values": attr.label(
            default = ":collect_stamped_values",
            executable = True,
            allow_files = True,
            cfg = "exec",
        ),
        "optional_keys": attr.string_dict(
            doc = """Set of stamp keys whose values to include when available,
            with an optional default value for each.

            If no value for the key is available and the default value
            is empty, don't include any corresponding value in the
            resulting stamped values.""",
        ),
        "required_keys": attr.string_list(
            doc = """Set of stamp keys whose values to include.

            If no value for the key is available, fail and emit no
            output.""",
        ),
        "out": attr.output(
            doc = "File to which to write the resulting stamped values",
        ),
    },
)
```
It relies on the _collect-stamped-values_ shell program.
<details>
<summary>collect-stamped-values</summary>

```bash
#!/usr/bin/env bash

set -e -u -o pipefail

required_key_mapping_file="$1"; shift
optional_key_mapping_file="$1"; shift
output_file="$1"; shift
input_files=("$@")

while read key; do
  if [ -z "${key}" ]; then
    continue
  fi
  if stamped_line=$(grep --max-count=1 \
                         --extended-regexp \
                         "^\Q${key}\E " \
                         "${input_files[@]}"); then
    read k v <<< $(echo "${stamped_line}")
    if [ -n "${v}" ]; then
      echo "${v}" >> "${output_file}"
    else
      echo >&2 "No stamped value is available for key \"${key}.\""
      exit 1
    fi
  else
    echo >&2 "No stamp entry is available for key \"${key}.\""
    exit 1
  fi
done < "${required_key_mapping_file}"

while read line; do
  if [ -z "${line}" ]; then
    continue
  fi
  if [[ "$line" =~ ^([^:]+):(.*)$ ]]; then
    key="${BASH_REMATCH[1]}"
    default_value="${BASH_REMATCH[2]}"
    if stamped_line=$(grep --max-count=1 \
                           --extended-regexp \
                           "^\Q${key}\E " \
                           "${input_files[@]}"); then
      read k v <<< $(echo "${stamped_line}")
      if [ -n "${v}" ]; then
        echo "${v}" >> "${output_file}"
      elif [ -n "${default_value}" ]; then
        echo "${default_value}" >> "${output_file}"
      else
        echo >&2 "No stamped value is available for optional key \"${key}.\""
        # Carry on without emitting an output value.
      fi
    elif [ -n "${default_value}" ]; then
      echo "${default_value}" >> "${output_file}"
    fi
  fi
done < "${optional_key_mapping_file}"

if [ ! -f "${output_file}" ]; then
  touch "${output_file}"
fi
```
</details>

I made that program available to preceding Bazel rule with the following target in the _BUILD.bazel_ file:
```bazel
sh_binary(
    name = "collect_stamped_values",
    srcs = ["collect-stamped-values"],
)
```

alexeagle: going to resolve so I can land this, we could chat more on a bazel-lib issue maybe?

seh: Right, I was _not_ proposing including it here. I just wanted to note that in lieu of something being built-in to _rules_oci_ like it was in _rules_docker_, I went off and wrote my own.

Whether it's good enough to publish as a library I don't know. I'd want your careful review of it. It's been working well enough for my limited use of it so far, but it's too young to be sure that I haven't missed anything wrong with it.

PR #164, opened 2023-04-17
fix: remove unused toolchain_name attribute

thesayyn:


alexeagle (approved):
right it's unused since changing to `curl` right?

thesayyn:
> right it's unused since changing to `curl` right?

yes. it was forgotten.

PR #173, opened 2023-04-21
docs: add build instruction for rust

vucong2409:
Hi guys, 
I'm using `rules_oci` and see that Rust build instructions is missing. 
So I  create this PR to add build instruction & a small example for Rust.


alexeagle (approved):
Awesome, thank you for adding this!!

PR #174, opened 2023-04-21
fix: take variant into account

### LINKED ISSUE #122: oci_pull should take variant, os.version, os.features into consideration
when selecting manifest out of platforms array

thesayyn:
fixes #122 
Also, the pinning was broken. we need a test for that.


thesayyn:
CI is acting weird. tests pass locally though

alexeagle:
>Also, the pinning was broken

could you please make a separate PR for something that can be orthogonal? easier to do a focused review on one thing at a time

oci/private/pull.bzl
```diff
@@ -479,78 +487,57 @@ oci_alias = repository_rule(
...
 cat <<EOF
 oci_pull(
     name = "{reponame}",
     digest = "sha256:{digest}",
     image = "{registry}/{repository}",
+    {optional_platforms}
+)
 EOF
-
-[[ $mediaType == "{manifestListType}" ]] && cat <<EOF
```
thesayyn, line 527: this was broken as it assumed that the manifest list type is always `application/vnd.oci.image.index.v1+json` which is not true for docker v2 images.  

also, it's overkill to make these calculations at runtime. it's easier to calculate platforms in the repository rule itself. see below.

oci/private/pull.bzl
```diff
@@ -479,78 +487,57 @@ oci_alias = repository_rule(
...
         fail(msg)
-    rctx.file("pin.sh", _pin_sh.format(
+
+    optional_platforms = ""
+    if manifest["mediaType"] in _SUPPORTED_MEDIA_TYPES["index"]:
+        # calculate platforms
+        platforms = []
+        for submanifest in manifest["manifests"]:
+            parts = [submanifest["platform"]["os"], submanifest["platform"]["architecture"]]
+            if "variant" in submanifest["platform"]:
```
thesayyn, line 527: this will also make future iterations easier for us when we add support for `os.version` and `os.features` which are a pain to deal within shell scripts.

thesayyn:
> could you please make a separate PR for something that can be orthogonal? easier to do a focused review on one thing at a time

ready. it's as small as it could get.

alexeagle (approved):

PR #182, opened 2023-04-21
chore: remove structure test

alexeagle:
The rule should just live in the repo that provides the binaries. It has no coupling at all with the code in this repo, rather it should be coupled with the container-structure-test Go program.

I'm working with @loosebazooka to see if that might be accepted upstream, then we don't need to carry this in the rules_oci public api.


thesayyn:
this is awesome.

alexeagle:
We have a preliminary okay to upstream this, so I'm going to remove it before 1.0

thesayyn:
> We have a preliminary okay to upstream this, so I'm going to remove it before 1.0

sounds good. lmk when it's ready for review

alexeagle:
@thesayyn ready for a review. We have to decide if we want to land this before we get into the upstream, or are okay with aspect-forks.

alexeagle:
upstream PR has landed, cleaning this up now.

thesayyn (approved):
🤩

PR #185, opened 2023-04-22
feat: Allow stampable repotags for oci_tarball

### LINKED ISSUE #184: oci_tarball: Support stamping of repotags similar to oci_push
Makes it a lot easier to test loading locally without needing extra steps to re-tag after loading.

josephglanville:
Fixes #184

### DIFF
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -4,8 +4,19 @@ set -o pipefail -o errexit -o nounset
 readonly YQ="{{yq}}"
 readonly IMAGE_DIR="{{image_dir}}"
 readonly BLOBS_DIR="{{blobs_dir}}"
-readonly REPOTAGS='{{repotags}}'
+readonly TAGS_FILE="{{tags}}"
 readonly TARBALL_MANIFEST_PATH="{{manifest_path}}"
+
+REPOTAGS=()
+# read repotags file as array and prepend it to REPOTAGS array.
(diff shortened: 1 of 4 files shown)


oci/private/tarball.sh.tpl
```diff
@@ -4,9 +4,20 @@ set -o pipefail -o errexit -o nounset
 readonly YQ="{{yq}}"
 readonly IMAGE_DIR="{{image_dir}}"
 readonly BLOBS_DIR="{{blobs_dir}}"
-readonly REPOTAGS='{{repotags}}'
+readonly TAGS_FILE="{{tags}}"
 readonly TARBALL_MANIFEST_PATH="{{manifest_path}}"
 
+REPOTAGS=()
+# read repotags file as array and prepend it to REPOTAGS array.
+IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
```
josephglanville, line 12: This is currently failing because `$TAGS_FILE` doesn't exist at the expected path. I'm not sure if I am not passing it in currently to the `actions.run` that executes the shell script or maybe using the wrong path attribute.

alexeagle: I changed `ctx.file.repotags.short_path` to `ctx.file.repotags.path` (the latter includes the `bazel-out/arch/bin` segments which are needed to resolve generated files)

alexeagle (approved):
thanks!

PR #190, opened 2023-04-29
Fix bug with formatting of repotags in tarball manifest.json

AttilaTheFun:
The current tarball.sh.tpl script is generating a manifest.json file which formats the repotags as a string. 

Example:
```
[
  {
    "Config": "blobs/sha256/81e06113ed0b517fd70e4933d505b5f29a117e13edc50b5bba81af88c8f0effb",
    "RepoTags": "phono:latest",
    "Layers": [
      "blobs/sha256/6b0b1527d75b1e49c2cb08f4eb8616c64c9a20a8b9d8c1179146b2675586536b.tar.gz",
      "blobs/sha256/26e3e4b0848c0e87391a7a844b11fcd2e6d223bd516be6ac8b17bac4e6808609.tar.gz",
      "blobs/sha256/6670ae47b92addccf12843ef586edfbd7764c5d360e2ddc4a64e035625ce4bb9.tar.gz"
    ]
  }
]
```

But Docker attempts to parse the manifest with an array of strings for the repotags, resulting in this error:
```
json: cannot unmarshal string into Go struct field manifestItem.RepoTags of type []string
```

This PR updates the bash script to format the repo tags as an array of strings like this:
```
[
  {
    "Config": "blobs/sha256/81e06113ed0b517fd70e4933d505b5f29a117e13edc50b5bba81af88c8f0effb",
    "RepoTags": ["phono:latest"],
    "Layers": [
      "blobs/sha256/6b0b1527d75b1e49c2cb08f4eb8616c64c9a20a8b9d8c1179146b2675586536b.tar.gz",
      "blobs/sha256/26e3e4b0848c0e87391a7a844b11fcd2e6d223bd516be6ac8b17bac4e6808609.tar.gz",
      "blobs/sha256/6670ae47b92addccf12843ef586edfbd7764c5d360e2ddc4a64e035625ce4bb9.tar.gz"
    ]
  }
]
```

I tested loading this tar with docker which was successful:
```
2023/04/29 17:01:14 Using unreleased version at commit de5aac1daa152338131d06c1bc7e309e74cd2e4f
Loading: 
Loading: 
Loading: 0 packages loaded
Analyzing: target //apps/phono:tarball (0 packages loaded, 0 targets configured)
INFO: Analyzed target //apps/phono:tarball (26 packages loaded, 1381 targets configured).
INFO: Found 1 target...
INFO: Elapsed time: 0.215s, Critical Path: 0.00s
INFO: 0 processes.
INFO: Build completed successfully, 0 total actions
f1e5859e2f15: Loading layer [==================================================>]    821kB/821kB
3f53f0ab435d: Loading layer [==================================================>]   7.09MB/7.09MB
33fc041a7618: Loading layer [==================================================>]  11.72MB/11.72MB
Loaded image: phono:latest
```

### DIFF
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -11,7 +11,6 @@ REPOTAGS=()
 # read repotags file as array and prepend it to REPOTAGS array.
 IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
 REPOTAGS=(${REPOTAGSFILE[@]+"${REPOTAGSFILE[@]}"} ${REPOTAGS[@]+"${REPOTAGS[@]}"})
-echo REPOTAGS: $REPOTAGS
 
 MANIFEST_DIGEST=$(${YQ} eval '.manifests[0].digest | sub(":"; "/")' "${IMAGE_DIR}/index.json")
 MANIFEST_BLOB_PATH="${IMAGE_DIR}/blobs/${MANIFEST_DIGEST}"
@@ -42,7 +41,6 @@ else
     done
     REPOTAGS_ARRAY=$REPOTAGS_ARRAY"]"
 fi
-echo REPOTAGS_ARRAY: $REPOTAGS_ARRAY
 
 config="blobs/${CONFIG_DIGEST}" \
 repotags="$REPOTAGS_ARRAY" \
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -11,6 +11,7 @@ REPOTAGS=()
 # read repotags file as array and prepend it to REPOTAGS array.
 IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
 REPOTAGS=(${REPOTAGSFILE[@]+"${REPOTAGSFILE[@]}"} ${REPOTAGS[@]+"${REPOTAGS[@]}"})
+echo REPOTAGS: $REPOTAGS
 
 MANIFEST_DIGEST=$(${YQ} eval '.manifests[0].digest | sub(":"; "/")' "${IMAGE_DIR}/index.json")
 MANIFEST_BLOB_PATH="${IMAGE_DIR}/blobs/${MANIFEST_DIGEST}"
@@ -27,8 +28,24 @@ for LAYER in $(${YQ} ".[]" <<< $LAYERS); do
     cp "${IMAGE_DIR}/blobs/${LAYER}" "${BLOBS_DIR}/${LAYER}.tar.gz"
 done
 
+# format the repotags as a json array:
+if [ -z "$REPOTAGS" ]
+then
+    REPOTAGS_ARRAY=[]
+else
+    REPOTAGS_ARRAY=[
+    for repotag in $REPOTAGS
+    do
+        REPOTAGS_ARRAY=$REPOTAGS_ARRAY\"
+        REPOTAGS_ARRAY=$REPOTAGS_ARRAY$repotag
+        REPOTAGS_ARRAY=$REPOTAGS_ARRAY\",
+    done
+    REPOTAGS_ARRAY=$REPOTAGS_ARRAY"]"
+fi
+echo REPOTAGS_ARRAY: $REPOTAGS_ARRAY
+
 config="blobs/${CONFIG_DIGEST}" \
-repotags="${REPOTAGS:-[]}" \
+repotags="$REPOTAGS_ARRAY" \
 layers="${LAYERS}" \
 "${YQ}" eval \
         --null-input '.[0] = {"Config": env(config), "RepoTags": env(repotags), "Layers": env(layers) | map( "blobs/" + . + ".tar.gz") }' \


aryeh-looker:
+1

oci/private/tarball.sh.tpl
```diff
@@ -27,8 +27,23 @@ for LAYER in $(${YQ} ".[]" <<< $LAYERS); do
     cp "${IMAGE_DIR}/blobs/${LAYER}" "${BLOBS_DIR}/${LAYER}.tar.gz"
 done
 
+# format the repotags as a json array:
```
alexeagle, line 30: I think you could do this inside the `yq` filter below and it would be a lot simpler? allow the `env(repotags)` to continue to read the value as it's formatted now, and use some `yq` function to convert to an array?

AttilaTheFun: Hmm so I tried:
```
repotags="$REPOTAGS" \
layers="${LAYERS}" \
"${YQ}" eval \
        --null-input '.[0] = {"Config": env(config), "RepoTags": env(repotags) | split(" "), "Layers": env(layers) | map( "blobs/" + . + ".tar.gz") }' \
        --output-format json > "${TARBALL_MANIFEST_PATH}"
```

And that worked for a single repotag but when I added multiple I got this error:
```
Error: Value for env variable 'config' not provided in env()
```

The for-loop worked though.

oci/private/tarball.sh.tpl
```diff
@@ -7,11 +7,6 @@ readonly BLOBS_DIR="{{blobs_dir}}"
 readonly TAGS_FILE="{{tags}}"
 readonly TARBALL_MANIFEST_PATH="{{manifest_path}}"
 
-REPOTAGS=()
-# read repotags file as array and prepend it to REPOTAGS array.
-IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
-REPOTAGS=(${REPOTAGSFILE[@]+"${REPOTAGSFILE[@]}"} ${REPOTAGS[@]+"${REPOTAGS[@]}"})
```
AttilaTheFun, line 13: @alexeagle @aryeh-looker While testing I also figured out that this logic did not work with multiple repotags -- it would only store the first one in the REPOTAGS variable.

Given that the tags file was already space separated, all I had to do was read it into the REPOTAGS variable directly because the IFS is space by default:
```
REPOTAGS=$(cat $TAGS_FILE)
```

This in combination with my fix below allowed me to create an image from a tarball with multiple repotags:
```
[ 
    { 
         "Config": "blobs/sha256/300fcd932ebbfab3377f4d1450e2e610ecabbbebda1dc6166ab3bc1b18414985",
         "RepoTags": [ "phono:latest", "test:latest" ], 
         "Layers": [
            "blobs/sha256/6b0b1527d75b1e49c2cb08f4eb8616c64c9a20a8b9d8c1179146b2675586536b.tar.gz",   
            "blobs/sha256/26e3e4b0848c0e87391a7a844b11fcd2e6d223bd516be6ac8b17bac4e6808609.tar.gz", 
            "blobs/sha256/336c487350cddcf412379144afef9375e4e4611df01abc41ab558e84bbcb80b8.tar.gz" 
         ] 
    } 
]
```

thesayyn: > Given that the tags file was already space separated

they are supposed to be `newline` separated

AttilaTheFun:
@aryeh-looker @alexeagle How do I re-run the CI jobs?

oci/private/tarball.sh.tpl
AttilaTheFun: Well something else is likely broken then, because when I logged the file contents they were space separated 😕

alexeagle:
sorry, until GH sees that you've contributed by landing a PR, it is configured to wait for someone to approve each CI run.

oci/private/tarball.sh.tpl
alexeagle: depends how you produced the file, but if you use the `oci_tarball` macro it puts one-per-line here https://github.com/bazel-contrib/rules_oci/blob/main/oci/defs.bzl#L102

alexeagle:
Here's the missing test case: https://github.com/bazel-contrib/rules_oci/commit/2a345dc0508f3148f3f0de0e768655a7f7244ac6

You don't permit pushes to your branch on this fork, so I can't add that myself. Having the test will make it easier for all of us to iterate on this.

AttilaTheFun:
My bad @alexeagle I used the main branch in the forked repo to create this PR -- in retrospect I should have created a branch in the fork to create the PR from. I can rebase it to pick up your test though.

oci/private/tarball.sh.tpl
AttilaTheFun: @alexeagle I verified that it is newline separated, but the bash script is not correctly handling the newlines.
I tried this locally:
```
echo "$(cat $TAGS_FILE)"

REPOTAGS=()
# read repotags file as array and prepend it to REPOTAGS array.
IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
for repotag in $REPOTAGSFILE
do
  echo "$repotag"
done
```

And this was the output:
```
phono:latest
test:latest
phono:latest
```

I.e. the file does have the newlines but the REPOTAGSFILE is losing all of the lines after the first.

AttilaTheFun:
@alexeagle @thesayyn @aryeh-looker I figured out a better way to fix this and created a new PR off of a branch instead of main here:
https://github.com/bazel-contrib/rules_oci/pull/205

I'm closing this PR in favor of that one.

PR #200, opened 2023-05-01
fix: pinning is broken

thesayyn:
prefactor https://github.com/bazel-contrib/rules_oci/issues/122


alexeagle:
File a TODO issue so we eventually add some integration testing for it?

thesayyn:
> File a TODO issue so we eventually add some integration testing for it?

https://github.com/bazel-contrib/rules_oci/issues/202

PR #205, opened 2023-05-03
Fix bug with formatting of repotags in tarball manifest.json [re-do]

AttilaTheFun:
The current tarball.sh.tpl script is generating a manifest.json file which formats the repotags as a string. 

Example:
```
[
  {
    "Config": "blobs/sha256/81e06113ed0b517fd70e4933d505b5f29a117e13edc50b5bba81af88c8f0effb",
    "RepoTags": "phono:latest",
    "Layers": [
      "blobs/sha256/6b0b1527d75b1e49c2cb08f4eb8616c64c9a20a8b9d8c1179146b2675586536b.tar.gz",
      "blobs/sha256/26e3e4b0848c0e87391a7a844b11fcd2e6d223bd516be6ac8b17bac4e6808609.tar.gz",
      "blobs/sha256/6670ae47b92addccf12843ef586edfbd7764c5d360e2ddc4a64e035625ce4bb9.tar.gz"
    ]
  }
]
```

But Docker attempts to parse the manifest with an array of strings for the repotags, resulting in this error:
```
json: cannot unmarshal string into Go struct field manifestItem.RepoTags of type []string
```

This PR updates the bash script to format the repo tags as an array of strings like this:
```
[
  {
    "Config": "blobs/sha256/81e06113ed0b517fd70e4933d505b5f29a117e13edc50b5bba81af88c8f0effb",
    "RepoTags": ["phono:latest"],
    "Layers": [
      "blobs/sha256/6b0b1527d75b1e49c2cb08f4eb8616c64c9a20a8b9d8c1179146b2675586536b.tar.gz",
      "blobs/sha256/26e3e4b0848c0e87391a7a844b11fcd2e6d223bd516be6ac8b17bac4e6808609.tar.gz",
      "blobs/sha256/6670ae47b92addccf12843ef586edfbd7764c5d360e2ddc4a64e035625ce4bb9.tar.gz"
    ]
  }
]
```

I tested loading this tar with docker which was successful:
```
2023/04/29 17:01:14 Using unreleased version at commit de5aac1daa152338131d06c1bc7e309e74cd2e4f
Loading: 
Loading: 
Loading: 0 packages loaded
Analyzing: target //apps/phono:tarball (0 packages loaded, 0 targets configured)
INFO: Analyzed target //apps/phono:tarball (26 packages loaded, 1381 targets configured).
INFO: Found 1 target...
INFO: Elapsed time: 0.215s, Critical Path: 0.00s
INFO: 0 processes.
INFO: Build completed successfully, 0 total actions
f1e5859e2f15: Loading layer [==================================================>]    821kB/821kB
3f53f0ab435d: Loading layer [==================================================>]   7.09MB/7.09MB
33fc041a7618: Loading layer [==================================================>]  11.72MB/11.72MB
Loaded image: phono:latest
```

There were actually two bugs in the script:
- The first was that the IFS was set to a newline when parsing the file (which was fine) but then the original IFS was never restored (it should normally be a space). This broke bash handling arrays later in the file, since arrays aren't really a type but an IFS separated string.
- The second was that, once I fixed that issue, the variable was still a space-separated string. I reformatted this into an array with square brackets and it worked as expected.

### DIFF
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -7,10 +7,11 @@ readonly BLOBS_DIR="{{blobs_dir}}"
 readonly TAGS_FILE="{{tags}}"
 readonly TARBALL_MANIFEST_PATH="{{manifest_path}}"
 
-REPOTAGS=()
-# read repotags file as array and prepend it to REPOTAGS array.
-IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
-REPOTAGS=(${REPOTAGSFILE[@]+"${REPOTAGSFILE[@]}"} ${REPOTAGS[@]+"${REPOTAGS[@]}"})
+# read repotags file as an array:
+SAVEIFS=$IFS   # Save current IFS (Internal Field Separator)
+IFS=$'\n'      # Change IFS to newline char
+REPOTAGS=$(cat "$TAGS_FILE")
+IFS=$SAVEIFS   # Restore original IFS
 
 MANIFEST_DIGEST=$(${YQ} eval '.manifests[0].digest | sub(":"; "/")' "${IMAGE_DIR}/index.json")
 MANIFEST_BLOB_PATH="${IMAGE_DIR}/blobs/${MANIFEST_DIGEST}"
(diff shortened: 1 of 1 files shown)


oci/private/tarball.sh.tpl
```diff
@@ -7,10 +7,27 @@ readonly BLOBS_DIR="{{blobs_dir}}"
 readonly TAGS_FILE="{{tags}}"
 readonly TARBALL_MANIFEST_PATH="{{manifest_path}}"
 
-REPOTAGS=()
-# read repotags file as array and prepend it to REPOTAGS array.
-IFS=$'\n' REPOTAGSFILE=($(cat "$TAGS_FILE"))
```
thesayyn, line 12: I think you are overdoing it here.  IFS can be set in a subshell `(  )` which only affects the subshell context where `cat` runs here. 

 you just need 
```
REPOTAGSFILE=($(IFS=$'\n' cat "$TAGS_FILE"))
```

AttilaTheFun: @thesayyn Ok I updated the PR to use the subshell. The outer parenthesis was actually breaking the zero-tags case but this worked:
```
REPOTAGS=$(IFS=$'\n' cat "$TAGS_FILE")
```

thesayyn: would you like to update the PR accordingly? I'd like to get you on our contributiors list

thesayyn (approved):

PR #207, opened 2023-05-03
refactor: drop rules_pkg dependency

### LINKED ISSUE #177: Suggest installing rules_python rather than rules_pkg_dependencies
our snippet says:
```
load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")

rules_pkg_dependencies()
```

But that thing gives a very old rules_python. We could
- stop depending on rules_pkg (?)
- fix rules_pkg to have "modern" python
- give install snippet that inlines rules_pkg_dependencies

thesayyn:
fixes #177


oci/private/tarball.sh.tpl
```diff
@@ -32,4 +33,6 @@ config="blobs/${CONFIG_DIGEST}" \
 layers="${LAYERS}" \
 "${YQ}" eval \
         --null-input '.[0] = {"Config": env(config), "RepoTags": "${repotags}" | envsubst | split("%"), "Layers": env(layers) | map( "blobs/" + . + ".tar.gz") }' \
-        --output-format json > "${TARBALL_MANIFEST_PATH}"
+        --output-format json > "${STAGING_DIR}/manifest.json"
+
+tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" manifest.json blobs
```
alexeagle, line 38: add a TODO here that this is non-hermetic and maybe has portability problems, we'd like to get a proper tar toolchain from bazel-lib.

alexeagle (approved):

PR #211, opened 2023-05-03
feat(push): restore syntax sugar for repository attribute

alexeagle:
It's just handled in the macro, the underlying rule still expects a file of repotags


alexeagle:
closing, will just revert #169 instead to restore the user interface with `oci_push#repository`

PR #213, opened 2023-05-03
Revert "refactor: merge repository into repotags (#169)"

alexeagle:
We decided the new API is significantly harder to use, and less intuitive.

This reverts commit 85b695c9b953c3a96d46c116e8fa7af3026588f1.

### DIFF
--- a/docs/push.md
+++ b/docs/push.md
@@ -7,7 +7,7 @@
 ## oci_push_rule
 
 <pre>
-oci_push_rule(<a href="#oci_push_rule-name">name</a>, <a href="#oci_push_rule-image">image</a>, <a href="#oci_push_rule-repotags">repotags</a>)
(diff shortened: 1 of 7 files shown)


thesayyn (approved):

seh:
> We decided the new API is significantly harder to use, and less intuitive.

I am happy to read this, because that was my reaction too. I had to teach my stamping rule a few new tricks to include the registry prefix in front of the tags, but the outcome felt strange. Most people don't think of the whole image name as a "tag," despite the longstanding _docker tag_ command's interface.

alexeagle:
Yeah @seh I was trying to upgrade a client to HEAD and my nice stamping stuff got gross, so we questioned whether we made the right call. Any other feedback is welcome too!

PR #214, opened 2023-05-03
test: assert oci_* rules are reproducible

### LINKED ISSUE #10: Reproducibility and assertion via CI
One of the important goals of rules_container is to be reproducible. the final image should be canonical as long as the compilation target is the same. we should assert this with our CI pipeline.

A few examples of this would be

- host `darwin/arm64`  cross-compiling to `linux/arm64`
- host `darwin/amd64`  cross-compiling to `linux/arm64`
- host `linux/armel` cross-compiling to `linux/arm64`
- host `linux/arm64` compiling to `linux/arm64`
- host `windows/amd64`  cross-compiling to `linux/arm64`

with all of the cases above, the resulting image should be canonical and have the same digest. this is a realistic goal as long as the upstream `*_binary` rules are deterministic and spit the same `runfiles` regardless of the host platform.

there are a few things to be careful about in order to ensure reproducibility.
- stamping
- `history` property in image config
- `--mtime` embedded in tar layers.
- `--sort=name` file ordering in tar layers.
- `--owner=0 --group=0 --numeric-owner` in tar layers.
- strip PAX headers from tar layers.

thesayyn:
fixes #10

### DIFF
--- a/examples/multi_arch/BUILD.bazel
+++ b/examples/multi_arch/BUILD.bazel
@@ -1,5 +1,6 @@
 load("//oci:defs.bzl", "oci_image", "oci_image_index")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
+load("@aspect_bazel_lib//lib:testing.bzl", "assert_contains")
 load(":transition.bzl", "multi_arch")
 
 pkg_tar(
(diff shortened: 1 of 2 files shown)


examples/multi_arch/BUILD.bazel
```diff
@@ -30,3 +31,17 @@ oci_image_index(
         ":images",
     ],
 )
+
+genrule(
+    name = "hash",
+    srcs = [":index"],
+    outs = ["sha256.sum"],
+    cmd = "$(JQ_BIN) -r '.manifests[0].digest' $(location :index)/index.json > $@",
```
alexeagle, line 39: nit: maybe you want `assert_json_matches` instead?

alexeagle (approved):

examples/multi_arch/BUILD.bazel
thesayyn: I wanted to but it uses diff_test from skylib which doesn't support passing individual files out of tree artifacts.

PR #222, opened 2023-05-04
refactor: rename repotags to default_tags

thesayyn:
Having `repotags` in both oci_push and oci_tarball is confusing for the users given they have different semantics.

### DIFF
--- a/docs/push.md
+++ b/docs/push.md
@@ -70,8 +70,10 @@ oci_push(
 ```
 
 When running the pusher, you can pass flags:
-- Override `repository`: `-r|--repository` flag. e.g. `bazel run //myimage:push -- --repository index.docker.io/<ORG>/image`
-- Additional `default_tags`: `-t|--tag` flag, e.g. `bazel run //myimage:push -- --tag latest`
+
+- Override `repository`; `-r|--repository` flag. e.g. `bazel run //myimage:push -- --repository index.docker.io/<ORG>/image`
+- Tags in addition to default_tags `default_tags`; `-t|--tag` flag, e.g. `bazel run //myimage:push -- --tag latest`
+
 
 
 **ATTRIBUTES**
--- a/docs/push.md
+++ b/docs/push.md
@@ -104,7 +104,7 @@ Allows the default_tags attribute to be a list of strings in addition to a text 
 | Name  | Description | Default Value |
 | :------------- | :------------- | :------------- |
 | <a id="oci_push-name"></a>name |  name of resulting oci_push_rule   |  none |
-| <a id="oci_push-default_tags"></a>default_tags |  a list of tags to apply to the image after pushing, or a label of a file containing tags one-per-line. See [stamped_tags](https://github.com/bazel-contrib/rules_oci/blob/main/examples/push/stamp_tags.bzl) as one example of a way to produce such a file.   |  <code>[]</code> |
+| <a id="oci_push-default_tags"></a>default_tags |  a list of tags to apply to the image after pushing, or a label of a file containing tags one-per-line. See [stamped_tags](https://github.com/bazel-contrib/rules_oci/blob/main/examples/push/stamp_tags.bzl) as one example of a way to produce such a file.   |  <code>None</code> |
 | <a id="oci_push-kwargs"></a>kwargs |  other named arguments to [oci_push_rule](#oci_push_rule).   |  none |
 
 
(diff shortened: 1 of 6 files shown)


seh (commented):
What does "default" mean in this context? Is there something else that could override these chosen tags?

thesayyn:
> What does "default" mean in this context? Is there something else that could override these chosen tags?

semantics is the same before. default here implies tags to apply by default in addition to ones provided via `--tag` at `bazel run //target:push -- --tag additional_tag`

seh:
> in addition to ones provided via `--tag` at `bazel run //target:push -- --tag additional_tag`

Ah, I didn't know about that flag being registered here. Thank you for explaining.

Perhaps instead of "default," either "unconditional" or "common" would be clearer.

seh:
> makes me confused and ask more questions about the condition here.

It means that these tags are always set, as opposed to any that may or may not be present on the command line via the `--tag` flag.

> suggests that it's shared with something else? unclear what common is.

There I meant to suggest that every invocation of a pushing target would bear these tags, again contrasted with the variation that could come with selective use of the `--tag` flag.

This is hard to name, because "default" leaves me wondering if using the `--tag` flag _replaces_ all the "default" values (meaning they're used only in the absence of any `--tag` flag values), or always adds more values to the default set.

thesayyn:
I agree with you that `--tag` flag makes it confusing. but it's documented properly i think. 

from docs 
> When running the pusher, you can pass flags:
> - Override `repository`: `-r|--repository` flag. e.g. `bazel run //myimage:push -- --repository index.docker.io/<ORG>/image`
> - Additional `default_tags`: `-t|--tag` flag, e.g. `bazel run //myimage:push -- --tag latest`

seh:
> it's documented properly i think.

There, instead of "Additional `default_tags`," I'd find it clearer if it said something like "Tags in addition to `default_tags`."

thesayyn:
> "Tags in addition to `default_tags`."

done

alexeagle (commented):
Last bit of bikeshedding
```
oci_push(
  default_tags = ["latest"],
  tags = ["deliverable"],
)
```
is maybe still confusing since Bazel already occupies "tags"

What about `push_tags` ?

thesayyn:
> is maybe still confusing since Bazel already occupies "tags"

idk `tags` is well known by bazel users already. At first glance looks confusing but i believe anything will be confusing at this point since Bazels use of `tags`.....

thesayyn:
> push_tags

name of the rule already implies the `push` part. a bit redundant it seems.  other proposals `remote_tags` `tags_to_apply` `tags_on_remote`

alexeagle:
I also don't love "default_tags" because it sounds like they will be overridden later by something else, rather than augmented

alexeagle:
I like `remote_tags` the best. If you `bazel build` the `oci_push` target, the tags aren't inherently "applied" in any sense to the artifact you get back. They're just stashed in a bash script.

alexeagle:
To be comprehensive, another choice is to mirror rules_docker's https://github.com/bazelbuild/rules_docker/blob/master/docs/container.md#container_push-tag

```
oci_push(
  tag = ["actually", "a", "list"],
  tags = ["no-remote", "deliverable"],
)
```

gregmagolan:
My vote is `image_tags`. Users expect `oci_push` to push images so `image_tags` would imply that it will also push those tags onto the image.

```
oci_push(
    image_tags = ["latest"],
    tags = ["no-remote"],
)
```

We should as rename `repotags` on `oci_tarball` to `repo_tags` to match the CamelCase `RepoTags` here: https://snyk.io/blog/container-image-formats/

alexeagle (approved):

alexeagle:
Decided on `remote_tags` based on more off-thread chat with @thesayyn

realtimetodie:
The changelog for v1.0.0-rc0 is misleading, as it mentions default_tags instead of remote_tags @alexeagle.

alexeagle:
Hmm thanks

alexeagle:
I don't see `default_tags` anywhere on https://github.com/bazel-contrib/rules_oci/releases/tag/v1.0.0-rc0

thesayyn:
> I don't see `default_tags` anywhere on https://github.com/bazel-contrib/rules_oci/releases/tag/v1.0.0-rc0

already fixed it.

PR #224, opened 2023-05-04
chore: rename repotags to repo_tags

alexeagle:
This reflects the CamelCase -> snake_case mapping, since the term in the manifest.json is 'RepoTags'


thesayyn:
the migration docs need a bit of update since the last two commits.

PR #225, opened 2023-05-05
Adapt to new yq version in aspect-build/bazel-lib

eikemeier:
`oci_push`does'nt work with the latest aspect-build/bazel-lib because of changes in yq, see https://github.com/aspect-build/bazel-lib/releases/tag/v1.31.0

See also https://github.com/mikefarah/yq/issues/1608

### DIFF
--- a/oci/private/push.sh.tpl
+++ b/oci/private/push.sh.tpl
@@ -37,7 +37,7 @@ while (( $# > 0 )); do
   esac
 done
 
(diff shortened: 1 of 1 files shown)


thesayyn:
yq has made a breaking change that's not semver compatible. we reverted the change bazel-lib that changed the default version for yq. this should not be necessary anymore.

PR #226, opened 2023-05-05
refactor: replace print statements with warning

thesayyn:


oci/private/pull.bzl
```diff
@@ -3,7 +3,7 @@
 load("@aspect_bazel_lib//lib:base64.bzl", "base64")
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("//oci/private:download.bzl", "download")
-load("//oci/private:util.bzl", "sha256", "warning")
+load("//oci/private:util.bzl", "sha256", "util")
```
alexeagle, line 6: what about the rest of the symbols in there? It's a little strange to have a mix of importing the "namespace" along with an individual symbol. Do you plan to put sha256 in the struct too?

alexeagle (approved):

oci/private/pull.bzl
thesayyn: handling those currently. I didn't want to mix them here.

PR #227, opened 2023-05-05
chore: remove bzlmod workaround

### LINKED ISSUE #230: "oci.pull" module extension tag's repository rule generates invalid references
Per [discussion in the "docker" channel](https://bazelbuild.slack.com/archives/CA3NW13MH/p1683306040074659) of the "Bazel" Slack workspace, using Bazel version 6.1.1 against _rules_oci_ at commit 38f819e3a51e5e2f2e05868a4fb8553ea6c5e761 and later, targets using the `oci.pull` module extension tag fail to find the _platforms_ Bazel module's repository, with a message like the following:
```
ERROR: Analysis of target '//some/tool:push_image' failed; build aborted:
  no such package '@[unknown repo 'platforms~0.0.6' requested from @rules_oci~1.0.0-rc0~oci~distroless_static_nonroot]//cpu':
    The repository '@[unknown repo 'platforms~0.0.6' requested from @rules_oci~1.0.0-rc0~oci~distroless_static_nonroot]' could not be resolved:
      No repository visible as '@platforms~0.0.6' from repository '@rules_oci~1.0.0-rc0~oci~distroless_static_nonroot'
```
Note carefully that the unknown repository name "@platforms~0.0.6" starts with a '@' character, offering a hint that the reference to the repository is malformed. Sure enough, in commit 38f819e3a51e5e2f2e05868a4fb8553ea6c5e761 (part of #219), we attempted to work around a problem in #221 by [_removing a '@' character_ from the labels referring to the _platforms_ module's repository](https://github.com/bazel-contrib/rules_oci/pull/219/files#diff-4649111173e0628fda68b8b26f4d67e2ea6ae47454f39c02e3393f41170f2709R495), leaving us with `@platforms-...` instead of `@@platforms-...`. Removing that workaround—as demonstrated in #227—allows the _BUILD_ file generated by the `oci.pull`-induced repository rule to work properly, at least for callers from _MODULE.bazel_.

alexeagle:
Bazel 6.2.0 contains a fix for https://github.com/bazelbuild/bazel/issues/17655

Fixes #230

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -490,9 +490,7 @@ buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {option
             name = rctx.attr.name,
             target_name = rctx.attr.target_name,
             platform_map = {
-                # Workaround bug in Bazel 6.1.0, see
-                # https://github.com/bazel-contrib/rules_oci/issues/221
-                str(k).replace("@@", "@"): v
+                str(k): v
                 for k, v in rctx.attr.platforms.items()
             },
         )
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,4 +1,4 @@
-6.1.0
+6.2.0
 
 # The first line of this file is used by Bazelisk and Bazel to be sure
 # the right version of Bazel is used to build and test this repo.
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,6 @@ jobs:
         bazelversion: ${{ fromJSON(needs.matrix-prep-bazelversion.outputs.bazelversions) }}
         folder:
           - .
-          - e2e/auth
           - e2e/custom_registry
           - e2e/wasm
           - e2e/smoke
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,6 +42,7 @@ jobs:
         bazelversion: ${{ fromJSON(needs.matrix-prep-bazelversion.outputs.bazelversions) }}
         folder:
           - .
+          - e2e/auth
           - e2e/custom_registry
           - e2e/wasm
           - e2e/smoke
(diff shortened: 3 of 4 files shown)


oci/private/pull.bzl
```diff
@@ -490,9 +490,7 @@ buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {option
             name = rctx.attr.name,
             target_name = rctx.attr.target_name,
             platform_map = {
```
fmeum, line 492: Didn't know this was even possible - you might want to use `repr({...})` as I don't know whether `str({...})` is guaranteed to roundtrip.

oci/private/pull.bzl
```diff
@@ -490,9 +490,7 @@ buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {option
...
-                # Workaround bug in Bazel 6.1.0, see
-                # https://github.com/bazel-contrib/rules_oci/issues/221
-                str(k).replace("@@", "@"): v
```
fmeum, line 495: The fix seems correct, even if it doesn't work here :-D. Is this running with a `WORKSPACE.bzlmod`? It could be picking up `platforms` from `WORKSPACE`.

oci/private/pull.bzl
Wyverald: either way, you have to pass the labels through `str()` -- `repr()` doesn't roundtrip labels, unfortunately...

oci/private/pull.bzl
Wyverald: IIUC this should always have been `str(k): v`. I don't understand why that causes #211 to fail... more digging is likely needed.

alexeagle: Bazel 6.2.0 contains the fix, I think it's fine to say rules_oci users need to upgrade. Thanks!

oci/private/pull.bzl
alexeagle: This seems to work now. THanks!

thesayyn (approved):

PR #231, opened 2023-05-08
fix: single image with tag is handled incorrectly

thesayyn:


oci/tests/BUILD.bazel
```diff
@@ -35,36 +31,17 @@ platform(
...
-    target_platform = _PLATFORM,
-)
-
-diff_test(
-    name = "test_java17",
-    file1 = "java17",
-    file2 = "distroless_java_amd64",
-)
-
-platform_transition_filegroup(
```
alexeagle, line 57: how come we needed the transition before and don't need it now?

oci/private/util.bzl
```diff
@@ -65,5 +65,6 @@ def _warning(rctx, message):
 
 util = struct(
     parse_image = _parse_image,
+    sha256 = _sha256,
```
alexeagle, line 68: yay consistency

alexeagle (approved):

oci/tests/BUILD.bazel
thesayyn: this was added to get past the select statement we have in oci_alias and get the concrete oci_pull which would `amd64` that is equivalent to `{name}_linux_amd64`.

thesayyn:
> Is there an issue fixed by it?

no. discovered it while going through the oci_pull code.

PR #242, opened 2023-05-11
test: add test for per registry credential helpers

thesayyn:


.bazelrc
```diff
@@ -4,6 +4,8 @@
 
 build --incompatible_strict_action_env
 build --nolegacy_external_runfiles
+build --symlink_prefix=/
```
alexeagle, line 7: why? it seems like it belongs in your personal bazelrc

alexeagle (approved):

.bazelrc
thesayyn: `bats -r .` recurses into bazel-bin causing infinite loops.  this belongs in e2e/auth actually. will fix it

thesayyn: oh, it's a symlink. can't do that. I'll remove this and put into .bazelrc.user

thesayyn: done.

PR #255, opened 2023-05-31
Add example for how to create a static content server image with nginx.

echochamber:
Seems like a common enough use case to be worth having an example for.


thesayyn:
thanks!

PR #256, opened 2023-06-01
Make sure that the `tars` list contains unique records

uhlajs:
There is no guarantee that the OCI Manifest layers record contains only unique items. For example, layers which contain no change in the docker image FS will always have a same digests. In such a case, `bazel` will complain that `srcs` for `blog` target contains duplicities. Therefore, we need to be sure that there is no such duplicities when creating `tars` list.

Example of error message:

```
ERROR: /beyond/.bazel_cache/external/test_image_single/BUILD.bazel:43:18: Label '@test_image//:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1' is duplicated in the 'srcs' attribute of rule 'blobs'
```

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -386,7 +386,8 @@ def _oci_pull_impl(rctx):
 
         # TODO: we should avoid eager-download of the layers ("shallow pull")
         downloader.download_blob(layer["digest"], hash)
-        tars.append(hash)
+        if hash not in tars:
+            tars.append(hash)
 
     # To make testing against `crane pull` simple, we take care to produce a byte-for-byte-identical
     # index.json file, which means we can't use jq (it produces a trailing newline) or starlark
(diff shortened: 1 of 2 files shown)


uhlajs:
To be honest, I do NOT fully understand the reason, why it is necessary to add `-f` option to `cp` command. The default behaviour of `cp` is that it overrides already existing files. Therefore, I would expect that the `-f` option is not needed, which is not the case. We should probably investigate this in detail before merging it.

thesayyn:
one question; are you able to pull said image with other tools such as docker pull?

uhlajs:
Yes. All operations work completely fine with such image. If necessary, I can prepare such image and upload it to some public repository.

thesayyn:
I couldn't find any information about this on oci-spec. I don't know what the behavior should be. omitting duplicate layers isn't safe as they rely on being squashed on top of each other respectively to their order.

uhlajs:
I haven't investigated the code in deep, so I cannot tell from top of my head, whether it is good or not. If there is a better solution, how to enable `rules_oci` working with such image, feel free to close this PR and prepare another.

thesayyn:
oh, actually, this is fine. `tars` is just a list of files we need to copy to oci-layout which should contain unique layers only.

uhlajs:
resolve #257

thesayyn (approved):

PR #271, opened 2023-06-15
fix: Avoid mkdir -p attempting to change permissions

anguslees:
Output directories (`ctx.actions.declare_directory` et al) are created by bazel
before invoking the action.  On 'clever' build executors, they may be symlinks
to other locations or other exotic implementation details.

A naive `mkdir -p` will try to create these directories if they don't exist _and
also_ set dir permissions to what mkdir thinks they should be.  This latter
permissions step can fail in more interesting build environments, even when the
dir already exists and `mkdir -p` should be a no-op.

This PR avoids invoking `mkdir -p` at all if the dir already exists.

### DIFF
--- a/oci/private/image_index.sh.tpl
+++ b/oci/private/image_index.sh.tpl
@@ -4,6 +4,10 @@ set -o pipefail -o errexit -o nounset
 readonly YQ="{{yq_path}}"
 readonly COREUTILS="{{coreutils_path}}"
 
+function mkdirp() {
+    test -d "$1" || "${COREUTILS}" mkdir -p "$1"
+}
+
 function add_image() {
     local image_path="$1"
     local output_path="$2"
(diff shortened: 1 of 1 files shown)


anguslees:
An alternative is to remove these `mkdir -p` statements entirely, since afaics they're unnecessary in this case (both locations are already declared as acton outputs, and thus should already exist by the time this script is invoked).

oci/private/image_index.sh.tpl
```diff
@@ -4,6 +4,10 @@ set -o pipefail -o errexit -o nounset
 readonly YQ="{{yq_path}}"
 readonly COREUTILS="{{coreutils_path}}"
 
+function mkdirp() {
```
thesayyn, line 7: can we leave some comments here?

alexeagle (approved):

PR #274, opened 2023-06-17
Fix typo in oci_push docs

Sovietaced:
I don't believe `tags` is a valid parameter name: https://github.com/bazel-contrib/rules_oci/blob/main/oci/defs.bzl#L62

It also doesn't seem to be a valid parameter of `oci_push_rule`.


thesayyn:
you need to change the source `.bzl` and then run `bazel test //docs/...` and follow instruction on the error message to update docs.

alexeagle (approved):
thanks!

PR #276, opened 2023-06-19
Replace --fail-with-body for a test that checks the status code

### LINKED ISSUE #183: Requires newer curl
I get `curl: option --fail-with-body: is unknown`
because I have `curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3` but that option is introduced in 7.76.0 

We have the usual choices:

- Curl is probably buildable under rules_foreign_cc https://everything.curl.dev/build#on-linux-and-unix-like-systems but then you own/maintain another build. I stopped myself even trying it.
- https://github.com/bazelbuild/bazel/issues/17829 is the actual thing we want - using curl is just a workaround. I wish we could put our time into the long-term fix and just skip around this curl thing.
- We could check the feasibility of not using this flag and find some way to accomplish the same thing with older curl versions.
- I'm sure there are pre-built curl binaries on the internet, but it's hard to trust them.

jamiees2:
Fixes #183 

The initial idea was to just replace curl with a script that does this check, something akin to https://stackoverflow.com/a/22508743. It seems much easier to just test the output in Starlark, since we have all the required information on-hand.

### DIFF
--- a/oci/private/download.bzl
+++ b/oci/private/download.bzl
@@ -72,7 +72,8 @@ def _download(
     command = [
         "curl",
         url,
-        "--fail-with-body",
+        "--write-out",
+        "%{http_code}",
         "--location",
         "--no-progress-meter",
         "--request",
@@ -92,12 +93,14 @@ def _download(
 
     result = rctx.execute(command)
 
-    _debug("""\nSTDOUT\n{}\nSTDERR\n{}""".format(result.stdout, result.stderr))
+    status_code = int(result.stdout.strip())
+    _debug("""\nSTATUS\n{}\nSTDOUT\n{}\nSTDERR\n{}""".format(status_code, result.stdout, result.stderr))
 
-    if result.return_code != 0 and allow_fail:
-        return struct(success = False)
-    elif result.return_code != 0 and not allow_fail:
-        fail("Failed to fetch {} {}".format(url, result.stderr))
+    if result.return_code != 0 or status_code >= 400:
+        if allow_fail:
+            return struct(success = False)
+        else:
+            fail("Failed to fetch {} {}".format(url, result.stderr))
 
     cache_it = rctx.download(
         url = "file://{}".format(output_path),


oci/private/download.bzl
```diff
@@ -72,7 +72,8 @@ def _download(
     command = [
         "curl",
         url,
-        "--fail-with-body",
+        "--write-out",
+        "%{http_code}",
```
thesayyn, line 76: hmm, should this also have something like %{http_body} so we also have the error message written to body?

jamiees2: My understanding is that this argument is only what gets written to stdout, but the actual body still gets written according to `--output` below.

Is there something you'd like to get from having the body in stdout?

thesayyn: this is different than the behavior of --fail-with-body. we'd like to get the response body written to stderr so we get meaningful errors. 

https://daniel.haxx.se/blog/2021/02/11/curl-fail-with-body/

jamiees2: I don't believe that's true, given that we provide the `--output` option. In that case, the response body always gets written to that output file, and nothing is output on stderr, whether you pass `--fail-with-body` or not. 

I tested this based on the example linked there, and got the following output:

```
$ curl --fail-with-body -o output https://example.com/404.html
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1256  100  1256    0     0   1836      0 --:--:-- --:--:-- --:--:--  1849
curl: (22) The requested URL returned error: 404

$ cat output
<!doctype html>
...
```

When we replace this with `--write-out`, the output is the following:
```
curl --write-out "%{http_code}" -o output https://example.com/404.html
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1256  100  1256    0     0    706      0  0:00:01  0:00:01 --:--:--   708
404
$ cat output
<!doctype html>
...
```

The difference here AFAICT is the exit code and the output, that now includes the status code. Does this match your understanding?

jamiees2: Oh I see, are you talking about the `curl: (22) The requested URL returned error: 404` error? Sorry, I thought we were talking about the response body.

You are right, that did get suppressed. I'm not sure how to work around that - that just seems to be dropped. We could add the status code to the `fail()`?

thesayyn: actually you are right. --output suppresses the --fail-with-body in this case. we should drop that flag.

thesayyn: i believe replacing `--fail-with-body` with `--fail` should be equivalent to this PR.

jamiees2: Hmm, so that is precisely the difference between the `--fail` and the `--fail-with-body` options. In the case of `--fail`,  then the output file is not logged, and we do not store the response body anywhere.

See the same example:
```
$ curl --fail -o output https://example.com/404.html
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0  1256    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404

$ cat output
cat: output: No such file or directory
```


I'm interested in what you prefer here - in my case it was helpful to be able to check the output of the curl command on failures, as you said, it's useful to be able to look at the body, but it also doesn't seem critical. Happy to change to `--fail`, but this PR in its current form preserves the semantics of `--fail-with-body`.

thesayyn (approved):

PR #277, opened 2023-06-19
Replace shasum with sha256sum

### LINKED ISSUE #259: Replace shasum with sha256sum
`shasum` is a perl script that doesn't appear to be available by default on some systems (Fedora & similar for one). Since you are only using SHA256, the `sha256sum` binary (which is not a perl script) could be a better choice, provided it doesn't have its own availability concerns. From a check of OS's I have quick access to, everything has `sha256sum` but not everything has `shasum`, but I don't have a Mac etc to check.

Alternatively you could support multiple potential methods and try whichever works first: e.g. `shasum` -> `sha256sum` -> `openssl sha256`.

I can send this as a PR if you like?

edit: looking at Debian as an example: `shasum` comes from the [`perl`](https://packages.debian.org/sid/amd64/perl/filelist) package, whereas `sha256sum` comes from [`coreutils`](https://packages.debian.org/sid/amd64/coreutils/filelist) which is an essential package for every install (things such as `rm`, `cp` and also `wc` which you declare as dependency too).

aaliddell:
Closes #259

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -9,7 +9,7 @@ TAG=${GITHUB_REF_NAME}
 PREFIX="rules_oci-${TAG:1}"
 ARCHIVE="rules_oci-$TAG.tar.gz"
 git archive --format=tar --prefix=${PREFIX}/ ${TAG} | gzip > $ARCHIVE
-SHA=$(sha256sum $ARCHIVE | awk '{print $1}')
+SHA=$(shasum -a 256 $ARCHIVE | awk '{print $1}')
 
 cat << EOF
 WORKSPACE snippet:
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -9,7 +9,7 @@ TAG=${GITHUB_REF_NAME}
 PREFIX="rules_oci-${TAG:1}"
 ARCHIVE="rules_oci-$TAG.tar.gz"
 git archive --format=tar --prefix=${PREFIX}/ ${TAG} | gzip > $ARCHIVE
-SHA=$(shasum -a 256 $ARCHIVE | awk '{print $1}')
+SHA=$(sha256sum $ARCHIVE | awk '{print $1}')
 
 cat << EOF
 WORKSPACE snippet:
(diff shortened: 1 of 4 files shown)


aaliddell:
I've added two levels of fallback if sha256sum fails, first with shasum and then with openssl. I don't have a MacOS machine to test this on, so if you are able to check this is suitable, that'd be useful.

.github/workflows/release_prep.sh
```diff
@@ -9,7 +9,7 @@ TAG=${GITHUB_REF_NAME}
 PREFIX="rules_oci-${TAG:1}"
 ARCHIVE="rules_oci-$TAG.tar.gz"
 git archive --format=tar --prefix=${PREFIX}/ ${TAG} | gzip > $ARCHIVE
-SHA=$(shasum -a 256 $ARCHIVE | awk '{print $1}')
+SHA=$(sha256sum $ARCHIVE | awk '{print $1}')
```
thesayyn, line 12: can we revert this change?

thesayyn: our CI runner has the shasum binary and i'd like to keep the diff small.

aaliddell: Sure, done

oci/private/util.bzl
```diff
@@ -42,15 +42,23 @@ def _parse_image(image):
...
         path: path to the file
     Returns:
         hashsum of file
     """
-    result = rctx.execute(["shasum", "-a", "256", path])
+    # Attempt to use the first viable method to calculate the SHA256 sum. sha256sum is part of
+    # coreutils on Linux, but is not available on MacOS. shasum is a perl script that is available
+    # on MacOS, but is not necessarily always available on Linux. OpenSSL is used as a final
+    # fallback if neither are available
+    result = rctx.execute(["sha256sum", path])
```
thesayyn, line 57: for performance reasons; can have the fallback strategy in this order; shasum256 is widely available on macos and linux distros, so it should be the first.

1- shasum
2- sha256sum
3- openssl

thesayyn (approved):
great. thanks!

aaliddell:
One moment, need an extra arg on openssl

Issue #273, opened 2023-06-16
building a tarball fails with more recent version of "yq"


rbielak:
I'm using yq version 4.33.3 with rules_oci. The script that creates the tar ball fails with this error:

```
bash    bazel-bin/examples/empty_base/tarball/tarball.sh 
09:25:56 initCommand [WARN] JSON file output is now JSON by default (instead of yaml). Use '-oy' or '--output-format=yaml' for yaml output
bazel-out/k8-fastbuild/bin/examples/empty_base/image/blobs/"sha256/b812c0570a7c369b2863c64e22760dc1b1dbc025a739f02db376bac62862f4cc"
Error: open bazel-out/k8-fastbuild/bin/examples/empty_base/image/blobs/"sha256/b812c0570a7c369b2863c64e22760dc1b1dbc025a739f02db376bac62862f4cc": no such file or directory
```
The fix is to remove the quotes around the "sha256/..." name in the path.

thesayyn:
See my comment https://github.com/bazel-contrib/rules_oci/pull/208#issuecomment-1603028508

alexeagle:
@thesayyn I think your analysis must have been wrong, because #278 is green without upgrading yq. So I guess at least in that one usage, there is an expression we can use that works with both yq versions.

PR #278, opened 2023-06-21
Remove '"' from path in MANIFEST_DIGEST when creating a tar file

rbielak:
Fix for this issue: https://github.com/bazel-contrib/rules_oci/issues/273

New version of "yq" will fail processing a path that includes '"'. This fix removes the extra '"' from the path in MANIFEST_DIGEST in the script template that creates the tarball.


alexeagle (approved):
It would be nice to add an e2e test that uses the newer yq to prove that this works, and doesn't regress. If you have time to add that...

PR #286, opened 2023-07-07
`cmd` and `endpoint` should expand `$(location)`

sfc-gh-ptabor:
When developing some macros on top of OCI rules, I discovered that if `cmd` (and `endpoint`) for consistency was able to resolve path to the target file (frequently transitively embedded within a tarball), than the macros could not require from user taking care of proper full paths of artifacts. Please see the attached example and go doc update.


oci/private/deps_aspect.bzl
```diff
@@ -0,0 +1,38 @@
+"""
+  An aspect that find's most transitive dependencies (target) of the given rule's attribute.
```
seh, line 2: Should "target" be pluralized? It's not clear whether you mean that dependencies are targets, vice versa, or neither.
```suggestion
  An aspect that finds most transitive dependencies (target) of the given rule's attribute.
```

sfc-gh-ptabor: Corrected.

thesayyn:
Hey @sfc-gh-ptabor.  Thanks for taking action and creating the PR.  I have a few concerns about the necessity of this but would like to write them into a discussion issue. I'd appreciate it if you could create an issue so we can have a discussion first involving other parties.

sfc-gh-ptabor:
@thesayyn: Thank you for taking the look. The issue is: https://github.com/bazel-contrib/rules_oci/issues/287

thesayyn:
> @thesayyn: Thank you for taking the look. The issue is: #287

Thanks! Ideally we should just make `entrypoint` and `cmd` stampable. See: https://github.com/bazel-contrib/rules_oci/issues/287#issuecomment-1633036519

sfc-gh-ptabor:
Closing and it got suppressed by https://github.com/bazel-contrib/rules_oci/pull/297

Issue #180, opened 2023-04-21
oci_tarball: provide easy "bazel runnable" target to load into the daemon


alexeagle:
So we don't have to put comments above each `oci_tarball` saying how to build/load/run the image.

It can print a "next, run `docker run bazel/the/image:bazel ` command for you to know what's next after loading it.

ewhauser:
@alexeagle I have a patch available for this. What are you thinking in regards to extensibility for the container engine? I'm just using docker in the script but could expose a `run_template` argument allowing it to be overridden? Longer term, it might make sense to expose a toolchain for the local container engine?

ewhauser:
I've submitted a PR with a strawman implementation and we can discuss there

thesayyn:
> I've submitted a PR with a strawman implementation and we can discuss there

looks like the PR introduces changes that runs `docker run` which isn't what we want. we only want to load it into the daemon. running it should happen outside of bazel as it becomes confusing what `runtime_args` does in this case.

ewhauser:
Understood why that would not be desired. Submitted a new PR: https://github.com/bazel-contrib/rules_oci/pull/290

sfc-gh-ptabor:
I assume it can be closed now as 'implemented'. Thank you for this feature.

PR #290, opened 2023-07-10
feat: #180 make the tarball target runnable

ewhauser:
Addresses #180

### DIFF
--- a/docs/tarball.md
+++ b/docs/tarball.md
@@ -19,7 +19,7 @@ bazel run :tarball
 docker run --rm "my-repository:latest"
 ```
 
-By default, the generated script will attempt to use Docker to load the image. If you need to use a runtime other than docker, you can replace the entire script template:
+By default, the generated script will attempt to use Docker or Podman to load the image. If you need to use a runtime other than Docker or Podman, you can replace the entire script template:
 
 ```
 oci_tarball(
--- a/docs/tarball.md
+++ b/docs/tarball.md
@@ -12,14 +12,23 @@ oci_tarball(
 )
 ```
 
-and then run it in a container like so:
+and then load and run it in a container like so:
 
 ```
-bazel build //path/to:tarball
-docker load --input $(bazel cquery --output=files //path/to:tarball)
-docker run --rm my-repository:latest
+bazel run :tarball
+docker run --rm "my-repository:latest"
 ```
 
+By default, the generated script will attempt to use Docker to load the image. If you need to use a runtime other than docker, you can replace the entire script template:
+
+```
+oci_tarball(
+    name = "tarball",
+    image = ":image",
+    repo_tags = ["my-repository:latest"],
+    run_template = "//path/to/your.tpl",
+)
+```
 
 <a id="#oci_tarball"></a>
 
--- a/docs/tarball.md
+++ b/docs/tarball.md
@@ -19,16 +19,7 @@ bazel run :tarball
 docker run --rm "my-repository:latest"
 ```
 
-By default, the generated script will attempt to use Docker or Podman to load the image. If you need to use a runtime other than Docker or Podman, you can replace the entire script template:
-
-```
-oci_tarball(
-    name = "tarball",
-    image = ":image",
-    repo_tags = ["my-repository:latest"],
-    run_template = "//path/to/your.tpl",
-)
-```
+The generated script will attempt to use Docker or Podman to load the image.
 
 <a id="#oci_tarball"></a>
 
--- a/docs/tarball.md
+++ b/docs/tarball.md
@@ -12,14 +12,13 @@ oci_tarball(
 )
 ```
 
-and then load and run it in a container like so:
+and then run it in a container like so:
 
 ```
 bazel run :tarball
-docker run --rm "my-repository:latest"
+docker run --rm my-repository:latest
 ```
 
-The generated script will attempt to use Docker or Podman to load the image.
 
 <a id="#oci_tarball"></a>
 
--- a/e2e/wasm/BUILD.bazel
+++ b/e2e/wasm/BUILD.bazel
@@ -47,7 +47,7 @@ build_test(
 # In order to run the image you need to follow instructions at https://docs.docker.com/desktop/wasm/ first.
 # then run the following;
 #  `bazel build :tarball`
-#  `docker load -i bazel-bin/tarball/tarball.tar`
+#  `bazel run :tarball``
 #  `docker run --runtime=io.containerd.wasmedge.v1 --platform=wasi/wasm32 --pull=never gcr.io/wasm:latest`
 oci_tarball(
     name = "tarball",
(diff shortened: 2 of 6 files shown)


oci/private/tarball_run.sh.tpl
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -o pipefail -o errexit -o nounset
+
+readonly IMAGE="{{image_path}}"
+docker load --input "$IMAGE"
```
alexeagle, line 5: I would like this to check if a `docker` command is on the PATH and automatically fallback to `podman` (basically do a convenience search for installed runtimes)
BUT that's possible without a breaking API change as a follow-up PR so I'm happy to get this in as-is.

docs/tarball.md
```diff
@@ -12,14 +12,23 @@ oci_tarball(
...
 
 ```
-bazel build //path/to:tarball
-docker load --input $(bazel cquery --output=files //path/to:tarball)
-docker run --rm my-repository:latest
+bazel run :tarball
+docker run --rm "my-repository:latest"
 ```
 
+By default, the generated script will attempt to use Docker to load the image. If you need to use a runtime other than docker, you can replace the entire script template:
```
alexeagle, line 22: can update this? Now that it searches for a command, most users wouldn't need `run_template` anyway, it feels like more of a "break the glass" escape hatch now

oci/private/tarball_run.sh.tpl
ewhauser: Changed the script to fallback to podman. 

Do you want me to make `run_template` private?

docs/tarball.md
```diff
@@ -12,14 +12,23 @@ oci_tarball(
...
 
-and then run it in a container like so:
+and then load and run it in a container like so:
 
 ```
-bazel build //path/to:tarball
-docker load --input $(bazel cquery --output=files //path/to:tarball)
-docker run --rm my-repository:latest
+bazel run :tarball
+docker run --rm "my-repository:latest"
```
alexeagle, line 19: I think the original FR wanted to run a container from the image too, not just load it into the daemon. But it's really questionable - I was always confused in rules_docker whether a given target was the language kind or not, and the default behavior differed. (as much as it was fun to `bazel run -- --norun` just like you "analyze" with `bazel build --nobuild`

Also you've already fixed the main ergonomic problem (the `bazel cquery`)

Pretty much convincing myself this is already fine but I wonder if you have any thoughts.

e2e/wasm/BUILD.bazel
```diff
@@ -47,7 +47,7 @@ build_test(
 # In order to run the image you need to follow instructions at https://docs.docker.com/desktop/wasm/ first.
 # then run the following;
 #  `bazel build :tarball`
```
alexeagle, line 49: can remove this build line

oci/private/tarball.bzl
```diff
@@ -14,7 +14,7 @@ and then run it in a container like so:
 
 ```
 bazel build //path/to:tarball
```
alexeagle, line 16: can remove (every one of these example command edits should be two lines replaced by one, right?)

oci/private/tarball_run.sh.tpl
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -o pipefail -o errexit -o nounset
+
+readonly IMAGE="{{image_path}}"
+if command -v docker &> /dev/null; then
+    docker load --input "$IMAGE"
+elif command -v podman &> /dev/null; then
+    podman load -i "$IMAGE"
+else
+    echo "Neither docker or podman could be found."
```
alexeagle, line 10: this would be a nice place to point out the `run_template` attribute since it's the next step for a user who hits this

alexeagle (commented):
love it!

docs/tarball.md
ewhauser: Given most people are coming from rules_docker, I thought it made sense for the behavior to be the same but @thesayyn suggested we didn't want that behavior so I removed it.

I was going to add an example to the ticket of extending oci_tarball with macro that would run the container as well in case it was wanted.

Open to other suggestions.

thesayyn: we should not get into the business of running containers with a single command. it makes the API surface for oci_tarball more confusing. originally it was designed to be an `oci_image -> docker tarball` converter but we decided that it's common enough to get its `runnable` target which `loads` into the daemon as well. and we should stop there, this is already expanding the surface, which will make it harder for us to support other tarball formats like oci layout tarball.

docs/tarball.md
```diff
@@ -12,14 +12,23 @@ oci_tarball(
...
 ```
 
+By default, the generated script will attempt to use Docker or Podman to load the image. If you need to use a runtime other than Docker or Podman, you can replace the entire script template:
+
+```
+oci_tarball(
+    name = "tarball",
+    image = ":image",
+    repo_tags = ["my-repository:latest"],
+    run_template = "//path/to/your.tpl",
```
thesayyn, line 29: I'd be happier if we didn't introduce yet another attribute here. if the default behavior doesn't serve the specific use cases, they can always wrap oci_tarball in a macro and their own runnable target.

oci/private/tarball.bzl
```diff
@@ -59,15 +66,26 @@ def _tarball_impl(ctx):
...
         executable = executable,
-        inputs = [image, ctx.file.repo_tags],
+        inputs = [image, repo_tags],
         outputs = [tarball],
         tools = [yq_bin],
         mnemonic = "OCITarball",
         progress_message = "OCI Tarball %{label}",
     )
 
+    exe = ctx.actions.declare_file(ctx.label.name + ".sh")
```
thesayyn, line 76:
```suggestion
    loader = ctx.actions.declare_file("{}.sh".format(ctx.label.name))
```

oci/private/tarball.bzl
```diff
@@ -32,6 +31,13 @@ attrs = {
             """,
         allow_single_file = [".txt"],
     ),
+    "run_template": attr.label(
```
thesayyn, line 34:
```suggestion
    "_loader_template": attr.label(
```

thesayyn: Let's keep this private.

docs/tarball.md
alexeagle: people coming from rules_docker are already used to both behaviors anyhow

https://github.com/bazelbuild/rules_docker#using-with-docker-locally

>For the lang_image targets, this will also run the container using docker run to maximize compatibility with lang_binary rules.

But plain `container_image` targets have the behavior we'll have here to just load the image with `bazel run`

docs/tarball.md
alexeagle: I made the attribute private for now, and filed #295 where we can consider what public API surface might be appropriate to provide more control.

alexeagle (approved):
thanks!

PR #293, opened 2023-07-12
fix(docs): indicate that oci_tarball#repo_tags is mandatory

### LINKED ISSUE #272: repo_tags marked as optional but rule fails if it is missing
In the `oci_tarball` rule, [repo_tags](https://github.com/bazel-contrib/rules_oci/blob/main/docs/tarball.md#oci_tarball-repo_tags) is marked as optional, however not providing it results in an error:

```
ERROR: .../BUILD: in oci_tarball rule ...:image: 
Traceback (most recent call last):
	File "/home/joguerra/.cache/bazel/_bazel_joguerra/b6e22f3e5f38bce7b2dab3eb0efc3043/external/rules_oci/oci/private/tarball.bzl", line 60, column 20, in _tarball_impl
		ctx.actions.run(
Error in run: at index 1 of inputs, got element of type NoneType, want File
```

If I add `repo_tags` to my target then it works.

```
oci_tarball(
    name = name,
    image = image,
    repo_tags = ["dummy"],
    tags = ["manual"],
)
```

alexeagle:
We always tried to read from the file, so this isn't a breaking change - it already wouldn't have worked

Fixes #272


thesayyn (approved):
nit: we should add a new oci_tarball target with no `repo_tags` attribute present to catch this case.

PR #297, opened 2023-07-13
feat: make cmd and environment attributes stampable

### LINKED ISSUE #287: `cmd` and `endpoint` should expand $(location)
Please allow to use `$(location)` type expansion in the `cmd` and `endpoint` attributes. 

It's a common need to configure cmd with:
  1. path to the binary that was build (golang)
  2. name of the deploy-jar 
  3. links to files that were deployed as part of the container. 

$(locations) [also $(rootpath(s)/execpath(s)...](https://bazel.build/reference/be/make-variables#predefined_label_variables) is a bazel mechanism to support replacing name of targets with the relative paths to these targets. This allows to automatically generate the proper path without requiring user to manually understand placement rules for files in bazel. 

It's especially important if rules_oci are used as a building blocks for generic -> language specific macros. This allows to build macros that take language-spacific artifact from user (like `go_binary` or `java_binary` deploy jar), put it into tar, wrap it into container, and generate command line that runs the artifact.

This potentially open a question whether artifacts within the container, should be placed at 'top level' path or preserve the package-location of the artifact in the bazel-tree. Both solution have pros & cons. The advantage of preserving the paths is that it guarantees lack of naming-conflicts between multiple artifacts being put in the same container and easier understanding of 'origin' (source location) of the target that produced given artifact.  The filp side is that source-code-refactorings (reorgs) are impacting the the structure within the container more.  I assume `oci_rules` should work well for both scenarios.

sfc-gh-ptabor:
Fixes #287

### DIFF
--- a/oci/private/image.sh.tpl
+++ b/oci/private/image.sh.tpl
@@ -121,6 +121,12 @@ for ARG in "$@"; do
             FIXED_ARGS+=("--annotation=$in")
           done <"${ARG#--annotations-file=}"
           ;;
+        (--cmd-file=*)
+          FIXED_ARGS+=("--cmd=$(<"${ARG#--cmd-file=}")")
+          ;;
+        (--entrypoint-file=*)
+          FIXED_ARGS+=("--entrypoint=$(<"${ARG#--entrypoint-file=}")")
+          ;;
         (*) FIXED_ARGS+=( "${ARG}" )
     esac
 done
--- a/oci/private/image.sh.tpl
+++ b/oci/private/image.sh.tpl
@@ -122,15 +122,20 @@ for ARG in "$@"; do
           done <"${ARG#--annotations-file=}"
           ;;
         (--cmd-file=*)
-          FIXED_ARGS+=("--cmd=$(<"${ARG#--cmd-file=}")")
+          while IFS= read -r in || [ -n "$in" ]; do
+            FIXED_ARGS+=("--cmd=$in")
+          done <"${ARG#--cmd-file=}"
           ;;
         (--entrypoint-file=*)
-          FIXED_ARGS+=("--entrypoint=$(<"${ARG#--entrypoint-file=}")")
+          while IFS= read -r in || [ -n "$in" ]; do
+            FIXED_ARGS+=("--entrypoint=$in")
+          done <"${ARG#--entrypoint-file=}"
           ;;
         (*) FIXED_ARGS+=( "${ARG}" )
     esac
 done
 
+echo "!!!: ${CRANE}" "${FIXED_ARGS[@]}"
 REF=$("${CRANE}" "${FIXED_ARGS[@]}")
 
 if [ ${#ENV_EXPANSIONS[@]} -ne 0 ]; then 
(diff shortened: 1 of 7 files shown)


sfc-gh-ptabor:
[ random activity to bring the issue to the 'needs attention' pool ]

thesayyn:
@sfc-gh-ptabor sorry I have been packed this week. I'll try to look into this next week, you are not forgotten.

sfc-gh-ptabor:
[missed release 1.3.0]

oci/private/image.sh.tpl
```diff
@@ -122,6 +122,12 @@ for ARG in "$@"; do
             FIXED_ARGS+=("--annotation=$in")
           done <"${ARG#--annotations-file=}"
           ;;
+        (--cmd-file=*)
+          FIXED_ARGS+=("--cmd=$(<"${ARG#--cmd-file=}")")
```
thesayyn, line 126: I am surprised that `--cmd` supports a file instead of an array of strings.

oci/private/image.sh.tpl
```diff
@@ -122,6 +122,12 @@ for ARG in "$@"; do
...
+          ;;
+        (--entrypoint-file=*)
+          FIXED_ARGS+=("--entrypoint=$(<"${ARG#--entrypoint-file=}")")
```
thesayyn, line 129: same as above. `--entrypoint` argument doesn't support files. https://github.com/google/go-containerregistry/blob/190ad0e4d556f199a07951d55124f8a394ebccd9/cmd/crane/cmd/mutate.go#L177-L178

You have to read the files line by line and create multiple `--entrypoint` and `--cmd` arguments per line.

thesayyn: like so https://github.com/bazel-contrib/rules_oci/blob/a5efb0efd19eb25e3f780b17e1bcb4d8444d6757/oci/private/image.sh.tpl#L110-L116

oci/private/image.sh.tpl
sfc-gh-ptabor: Hi, Thank you for review: 

It doesn't take a file: `$(<file)` is reading the file. Please: see: 
https://unix.stackexchange.com/questions/102894/what-does-the-expression-dir-file-mean-in-bash

It's pattern used above. 

----

Please take a look at crane docs: 

https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_mutate.md

```
      --cmd strings                 New cmd to set
      --entrypoint strings       New entrypoint to set
```
(plural strings)

AFAIR sending multiple --cmd flags caused the crane to misbehave (fail or just use the last flag).

sfc-gh-ptabor: Let me (tomorrow) expand the test to cover multiple flags scenario...  either way worth having it and I might be wrong.

oci/private/image.sh.tpl
sfc-gh-ptabor: Fixed.

thesayyn (approved):

sfc-gh-ptabor:
Thank you for review and fixing the debug line.

PR #300, opened 2023-07-18
fix: oci_tarball should support more than 2 repo_tags

sfc-gh-ptabor:
Due to usage of 'replace_one' in bash instead of 'replace_all',

the attempts to run `bazel run examples/empty_base/tarball` (with 3 tags), failed with 'Invalid reference format' error, as the line 2 and 3 kept being separated by 'explicit \n'.


sfc-gh-ptabor:
@alexeagle - could you, please take a look.

thesayyn (approved):
thanks!

PR #303, opened 2023-07-21
fix(windows): check in empty tar file rather than create dynamically

alexeagle:
Avoids a bug on Windows where the bash redirection tries to read a file descriptor with a path: Error: appending [/proc/198/fd/63]: reading layer '/proc/198/fd/63': CreateFile /proc/198/fd/63: The system cannot find the path specified.


thesayyn (approved):
awesome

PR #305, opened 2023-07-21
feat(windows): introduce batch wrapper

alexeagle:
This allows us to invoke our bash program under windows, otherwise you get the usual CreateProcess error %1 is not a valid win32 application

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -1,5 +1,7 @@
 "Implementation details for image rule"
 
+load("//oci/private:util.bzl", "util")
+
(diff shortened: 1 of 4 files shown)


oci/private/image.bzl
```diff
@@ -187,6 +191,7 @@ oci_image = rule(
     attrs = _attrs,
     doc = _DOC,
     toolchains = [
+        "@bazel_tools//tools/sh:toolchain_type",
```
thesayyn, line 194: hmm, is this a breaking change?

thesayyn (approved):
looks great!

PR #306, opened 2023-07-21
fix(windows): prevent path transformation of arguments to crane mutate

alexeagle:
Tested manually on windows.

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -170,10 +170,22 @@ def _oci_image_impl(ctx):
     output = ctx.actions.declare_directory(ctx.label.name)
     args.add(output.path, format = "--output=%s")
 
+    action_env = {}
+
+    # Windows: Don't convert arguments like --entrypoint=/some/bin to --entrypoint=C:/msys64/some/bin
+    if ctx.target_platform_has_constraint(ctx.attr._windows_constraint[platform_common.ConstraintValueInfo]):
+        # See https://www.msys2.org/wiki/Porting/:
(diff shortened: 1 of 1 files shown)


oci/private/image.bzl
```diff
@@ -174,6 +174,14 @@ def _oci_image_impl(ctx):
         inputs = depset(transitive = inputs_depsets),
         arguments = [args],
         outputs = [output],
+        # Don't convert arguments like --entrypoint=/some/bin to --entrypoint=C:/msys64/some/bin
```
thesayyn, line 177: nit: we should only set this on windows?

thesayyn (approved):

oci/private/image.bzl
alexeagle: sure, I'll make it conditional

alexeagle: actually I think I could move it into the .bat wrapper, will test that

thesayyn: that'd be awesome actually.

thesayyn: instead of being part of action cache key, it can be a runtime detail.

alexeagle: I failed to get that to work, it messes with something and causes

`ERROR: C:/users/alex/documents/github/rules_oci/e2e/smoke/BUILD.bazel:28:12: OCI Tarball //:tar failed: (Exit 1): wrap_tar.bat failed:tar.exe: could not chdir to '/tmp/tmp.Ekhtep7ycr'`

f0rmiga (approved):

PR #307, opened 2023-07-21
ci: add windows coverage

### LINKED ISSUE #97: Fix BCR presubmit on windows
See https://github.com/bazelbuild/bazel-central-registry/pull/513/files#diff-f32762d4a16f5f26043de312cfaa0b017b1f36ed0b02b52727a810e5daffd76dR4-R8

alexeagle:
Fixes #97

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -3,7 +3,7 @@ bcr_test_module:
   matrix:
     bazel: ["6.x", "7.x"]
     # TODO(#97): add windows
-    platform: ["debian10", "ubuntu2004"]
+    platform: ["debian10", "ubuntu2004", "windows"]
   tasks:
     test_linux:
       name: "Run test module"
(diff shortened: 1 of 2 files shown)


alexeagle:
Ugh! I can't reproduce this windows failure locally, even with 
`bazel clean --expunge; bazel --bazelrc=../../.github/workflows/ci.bazelrc --bazelrc=.bazelrc test --enable_bzlmod --noenable_runfiles //... --nocache_test_results`
it is still green for me. That spot that passes the `$(rlocation)` for tarball.tar so the test failure isn't making sense. I'll come back to it later maybe.

alexeagle:
@thesayyn this failure doesn't make sense to me from looking at the code, do you have any ideas?

thesayyn:
> @thesayyn this failure doesn't make sense to me from looking at the code, do you have any ideas?

it might repro if you delete docker.config at $HOME. Also it might help if you rebase as i am not sure if this branch was run against the main that has some fixes for docker config discovery.

Issue #309, opened 2023-07-23
[FR]: Easy way to produce images from apk packages


alexeagle:
apk is [Alpine package keeper](https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper)

https://github.com/chainguard-dev/apko is the obvious tool to use for this purpose

Thank you to [Chainguard](https://www.chainguard.dev/) for offering to fund this feature request!

farcop:
Its pity but it does not working on Mac M1

```
% bazel build //images/alpine:gen_apko                                  
INFO: Analyzed target //images/alpine:gen_apko (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /Users/farcop/Sources/kublr/images/alpine/BUILD.bazel:1:8: Executing genrule //images/alpine:gen_apko failed: (Exit 2): bash failed: error executing command (from target //images/alpine:gen_apko) /bin/bash -c ... (remaining 1 argument skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ℹ️            | loading config file: images/alpine/alpine.yaml
ℹ️  aarch64   | Building images for 1 architectures: [arm64]
ℹ️  aarch64   | building tags [alpine:latest]
ℹ️            | loading config file: images/alpine/alpine.yaml
ℹ️  aarch64   | build context:
ℹ️  aarch64   | build options:
{
        "withVCS": true,
        "workDir": "/var/folders/bb/_k53lmmn03vcfj7vsgxxy4_w0000gn/T/apko-2382180089/aarch64",
        "tarballPath": "/var/folders/bb/_k53lmmn03vcfj7vsgxxy4_w0000gn/T/apko-2382180089/image/apko-aarch64.tar.gz",
        "tags": [
                "alpine:latest"
        ],
        "sourceDateEpoch": "1970-01-01T00:00:00Z",
        "arch": "arm64",
        "Log": {
                "Out": {},
                "Fields": null,
                "Level": 4
        }
}
ℹ️  aarch64   | image configuration:
ℹ️  aarch64   |   contents:
ℹ️  aarch64   |     repositories: []
ℹ️  aarch64   |     keyring:      []
ℹ️  aarch64   |     packages:     [ca-certificates bash procps libc6-compat git less openssh jq unzip curl cdrkit ethtool openssl iperf3 netcat-openbsd]
ℹ️  aarch64   |   accounts:
[...]

PR #310, opened 2023-07-23
feat(apko): introduce toolchain to fetch and invoke apko executable

alexeagle:
First step in integrating that tool into rules_oci

Part of #309


alexeagle:
@thesayyn is moving this to another repo, as we determined in #309 that this doesn't belong here

PR #312, opened 2023-07-24
Ignore empty repo-tags in the file. Allow \n at the end of files.

sfc-gh-ptabor:
In particular without the change, a file with a trailing '\n' in the last line (expected) was considered to generate a new empty repo-tag.


alexeagle (approved):
Thanks for including a new test case!

PR #314, opened 2023-07-26
fix: allow passwords that contain  :

tokongs:
The way auth is currently splitting username:password does not work if the password contains colons(:). Such is the case when authenticating against the Google Artifact Registry using a Service Account.  Allowing a maximum of 1 splits will make it work.

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -91,7 +91,9 @@ def _get_auth(rctx, state, registry):
                 elif "auth" in auth_val:
                     # base64 encoded plaintext username and password
                     raw_auth = auth_val["auth"]
-                    (login, password) = base64.decode(raw_auth).split(":", 1)
+                    login, sep, password = base64.decode(raw_auth).partition(":")
+                    if not sep:
+                        fail("Auth string must be in form username:password")
                     pattern = {
                         "type": "basic",
                         "login": login,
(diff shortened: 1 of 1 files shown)


thesayyn (approved):

oci/private/pull.bzl
```diff
@@ -91,7 +91,7 @@ def _get_auth(rctx, state, registry):
                 elif "auth" in auth_val:
                     # base64 encoded plaintext username and password
                     raw_auth = auth_val["auth"]
-                    (login, password) = base64.decode(raw_auth).split(":")
+                    (login, password) = base64.decode(raw_auth).split(":", 1)
```
thirtyseven, line 94:
```suggestion
                    login, _, password = base64.decode(raw_auth).partition(":")
```
will avoid unpacking errors even if a colon is missing from the string.

thesayyn: I am not sure if that's what we want here. we except the basic auth string to contain `:`.

thirtyseven: True, but getting a "not enough values to unpack" error isn't user-friendly. Maybe:
```suggestion
                    login, sep, password = base64.decode(raw_auth).partition(":")
                    if not sep:
                        fail("Auth string must be in form username:password")
```

thesayyn: that would be a corrupted auth file in this case no?

tokongs: I think a proper error message would be nice. I was confused at first when I got the `too many values to unpack`.

PR #317, opened 2023-07-28
fix(bzlmod): print working MODULE.bazel reproducibility instructions

### LINKED ISSUE #315: buildozer command for reproducibility doesn't work with bzlmod
A `MODULE.bazel` file like
```
oci.pull(
    name = "distroless_base",
    tag = "latest",
    # digest = "sha256:ccaef5ee2f1850270d453fdf700a5392534f8d1a8ca2acda391fbb6a06b81c86",
    image = "gcr.io/distroless/base",
    platforms = [
        "linux/amd64",
        "linux/arm64",
    ],
)
```

results in 

```
WARNING: for reproducible builds, a digest is recommended.
Either set 'reproducible = False' to silence this warning,
or run the following command to change oci_pull to use a digest:

buildozer 'set digest "sha256:73deaaf6a207c1a33850257ba74e0f196bc418636cada9943a03d7abea980d6d"' 'remove tag' 'remove platforms' 'add platforms "linux/amd64" "linux/arm64" "linux/arm" "linux/s390x" "linux/ppc64le"' WORKSPACE:rules_oci~1.2.0~oci~distroless_base
```

which clearly doesn't work, as it references the WORKSPACE file.

@fmeum recently improved buildozer to support `use_repo` calls - do you have any recommendation if there's an easy way for us to print a command that works for all users?

alexeagle:
Fixes #315

thanks @fmeum

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -500,7 +500,7 @@ or run the following command to change {rule} to use a digest:
 
 buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {optional_platforms} {location}
     """.format(
-            location = "//MODULE.bazel:" + rctx.attr.name.split("~")[-1] if is_bzlmod_enabled() else "//WORKSPACE:" + rctx.attr.name,
+            location = "MODULE.bazel:" + rctx.attr.name.split("~")[-1] if is_bzlmod_enabled() else "WORKSPACE:" + rctx.attr.name,
             digest = digest,
             optional_platforms = optional_platforms,
             warning = "(make sure you use a recent buildozer release with MODULE.bazel support)" if is_bzlmod_enabled() else "",
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -1,6 +1,7 @@
 "Implementation details for oci_pull repository rules"
 
 load("@aspect_bazel_lib//lib:base64.bzl", "base64")
+load("@aspect_bazel_lib//lib:utils.bzl", "is_bzlmod_enabled")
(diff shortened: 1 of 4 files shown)


oci/private/pull.bzl
```diff
@@ -494,13 +495,16 @@ def _oci_alias_impl(rctx):
...
 Either set 'reproducible = False' to silence this warning,
-or run the following command to change oci_pull to use a digest:
+or run the following command to change {rule} to use a digest:
+{warning}
 
-buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {optional_platforms} WORKSPACE:{name}
+buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {optional_platforms} {location}
     """.format(
-            name = rctx.attr.name,
+            location = "MODULE.bazel:" + rctx.attr.name.split("~")[-1] if is_bzlmod_enabled() else "WORKSPACE:" + rctx.attr.name,
```
fmeum, line 503: Just note that a user may still have WORKSPACE repos of this type even if Bzlmod is enabled. 

Instead, you could add a new attribute that you only populated from the module extension as an indicator that the repo is coming from `MODULE.bazel`. You can also use this attribute to pass in the friendly repo name directly instead of reverse engineering it here.

thesayyn (approved):
Awesome

oci/private/pull.bzl
alexeagle: never mind, the repository name can already be found as a string argument in this location

fmeum: It looks like you could pass in this attribute unconditionally instead of only with Bzlmod, then you don't need to use the detection logic.

PR #323, opened 2023-08-04
fix: cat temporary file instead of echoing location.

nikron:
Location of output temp file is not that useful, but the contents of the file probably are.

### DIFF
--- a/oci/private/image.sh.tpl
+++ b/oci/private/image.sh.tpl
@@ -60,7 +60,7 @@ function base_from_layout() {
 
     "${CRANE}" push "${oci_layout_path}" "${registry}/oci/layout:latest" --image-refs "${refs}" > "${output}" 2>&1
 
-    echo "${output}" >&2
+    cat "${output}" >&2
 
     if grep -q "MANIFEST_INVALID" "${output}"; then
     cat >&2 << EOF


oci/private/image.sh.tpl
```diff
@@ -60,7 +60,7 @@ function base_from_layout() {
 
     "${CRANE}" push "${oci_layout_path}" "${registry}/oci/layout:latest" --image-refs "${refs}" > "${output}" 2>&1
 
-    echo "${output}" >&2
+    cat "${output}" >&2
```
thesayyn, line 63: we already print the contents of the log file when the script fails. I think this would make the output more confusing

nikron: The problem with that is crane sometimes fails before it gets to pushing to registry step. Then the log file doesn't have anything. I think we could at least cat this file if the crane push fails for some reason?

Additionally echoing the location of a temp file that gets deleted shortly doesn't help very much?

thesayyn: when this bash script exits, the contents of the log will be dumped to stderr anyway. doesn't that catch the case where crane exists early?

thesayyn:
Closing this as unnecessary.

PR #324, opened 2023-08-04
perf: make registry output to bazel-out

thesayyn:
This effectively eliminates the filesystem calls needed to copy the final image from the registry storage to the final output directory (bazel-out). 

Partially addresses: https://github.com/bazel-contrib/rules_oci/issues/365

There's still room for performance improvement for oci_image targets with `base`, which is hard linking individual blobs from the base image to the output directory directly to eliminate fs copy calls for loading base image onto the registry as well. however, i am not sure how well that would work with Sandox and RBE. I'll leave that as a follow-up.


oci/private/image.sh.tpl
```diff
@@ -6,7 +6,7 @@ set -o pipefail -o errexit -o nounset
 # NB: --output argument is an option only understood by this wrapper and will pull artifact image into a oci layout.
 
 readonly REGISTRY_LAUNCHER="{{registry_launcher_path}}"
-readonly CRANE="{{crane_path}}"
+readonly CRANE="/Users/thesayyn/Documents/go-containerregistry/main"
```
thesayyn, line 9: will revert this once i submit my patches to crane

oci/private/image.sh.tpl
```diff
@@ -33,7 +33,7 @@ function get_option() {
 
 function empty_base() {
     local registry=$1
-    local ref="$registry/oci/empty_base:latest"
+    local ref="$registry/image:latest"
```
thesayyn, line 36: nit: revert

oci/private/image.sh.tpl
```diff
@@ -58,9 +58,7 @@ function base_from_layout() {
     local oci_layout_path=$1
     local registry=$2
 
-    "${CRANE}" push "${oci_layout_path}" "${registry}/oci/layout:latest" --image-refs "${refs}" > "${output}" 2>&1
-
-    echo "${output}" >&2
+    "${CRANE}" push "${oci_layout_path}" "${registry}/image:latest" --image-refs "${refs}" > "${output}" 2>&1
```
thesayyn, line 61: nit: revert

oci/private/image.sh.tpl
```diff
@@ -151,7 +169,10 @@ if [ -n "${WORKDIR}" ]; then
 fi
 
 if [ -n "$OUTPUT" ]; then
+
     "${CRANE}" pull "${REF}" "./${OUTPUT}" --format=oci
```
thesayyn, line 173: nit: maybe convince crane people to introduce a new flag called `--prune` to remove any unreferenced blobs.

thesayyn:
this is unblocked now; https://github.com/bazel-contrib/rules_oci/pull/424

oci/private/image.sh.tpl
```diff
@@ -160,7 +158,12 @@ if [ ${#ENV_EXPANSIONS[@]} -ne 0 ]; then
 fi
 
 if [ -n "$OUTPUT" ]; then
-    "${CRANE}" pull "${REF}" "./${OUTPUT}" --format=oci
+    "${CRANE}" pull "${REF}" "./${OUTPUT}" --format=oci --annotate-ref
+    mv "${OUTPUT}/index.json" "${OUTPUT}/temp.json"
+    "${JQ}" --arg ref "${REF}" '.manifests |= map(select(.annotations["org.opencontainers.image.ref.name"] == $ref)) | del(.manifests[0].annotations)' "${OUTPUT}/temp.json" >  "${OUTPUT}/index.json"
```
thesayyn, line 163: this is needed because zot creates index.json in the storage directory and adds a new manifest entry for each preventing `crane layout gc` from garbage collecting earlier manifests. I'll work to see if i could come up with a clever trick to prevent this. 

FYI crane registry doesn't suffer from this as it never creates the index.json in the storage directory.

alexeagle (approved):
Do we have RBE testing for this repo? If not we'd better add it

Issue #280, opened 2023-07-03
oci_tarball creates an invalid manifest if more than 2 repo_tags are specified


mboulton-fathom:
BUILD file:
```bzl
load("@rules_oci//oci:defs.bzl", "oci_tarball")

oci_tarball(
    name = "my_bundle",
    image = ":my_image",
    repo_tags = ":labels.txt",
)
```
labels.txt:
```
a.f/a
a.f/b
a.f/c
a.f/d
```

manifest.json is created as:
```
{
    "Config": "blobs/sha256/81594b3eefac8dd701d4b0c9b1bb3e6837d5df55012470a07d7f33df4afe99be",
    "RepoTags": [
      "a.f/a",
      "a.f/b\na.f/c\na.f/d"
    ],
    "Layers": [
		...
	]
}
```

I'm guessing it's something to with this code: https://github.com/bazel-contrib/rules_oci/blob/ed776359d69e2a1a95d2482bb9a7aec287642b0f/oci/private/tarball.sh.tpl#L28-L36

PR #326, opened 2023-08-08
Fix invalid tarball manifest creation

mboulton-fathom:
Replacement for https://github.com/bazel-contrib/rules_oci/pull/298

Closes https://github.com/bazel-contrib/rules_oci/issues/280


alexeagle (approved):
Thanks!
Do you have time to also update some example or e2e/ to have multiple tags, so this can't regress?

Issue #253, opened 2023-05-28
rules_oci does not work on linux arm64 hosts due to jq dependency


chancila:
```
While resolving toolchains for target //bin/services/external_proxy:image: No matching toolchains found for types @aspect_bazel_lib//lib:jq_toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@aspect_bazel_lib//lib:jq_toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
```

`//bin/services/external_proxy:image` being a `oci_image` target

if this is a known issue perhaps stating compatible architectures might be handy

chancila:
I was able to work around this issue by building my own jq arm64 binary and registering a toolchain within my workspace locally:

on a mac arm64 host:
```
docker run --rm -it ubuntu:18.04
apt update
apt upgrade
apt install git build-essential make libtool autoconf
git clone https://github.com/jqlang/jq
git submodule update --init --recursive
autoreconf -fi
./configure --with-oniguruma=builtin
make -j8
```

```starlark
load("@aspect_bazel_lib//lib/private:jq_toolchain.bzl", "jq_toolchain")

jq_toolchain(
    name = "jq_linux_arm64_tools",
    bin = ":jq", # commited prebuilt jq binary locally
)

toolchain(
    name = "jq_linux_arm64_toolchain",
    exec_compatible_with = [
        "@platforms//os:linux",
        "@platforms//cpu:arm64",
    ],
    toolchain = ":jq_linux_arm64_tools",
    toolchain_type = "@aspect_bazel_lib//lib:jq_toolchain_type",
)
```

malt3:
FYI: This can soon be resolved using yq since [stamping support](https://github.com/aspect-build/bazel-lib/pull/456) is coming.

archen:
if anyone is looking for a quick patch-around, this works for my own usecases but I make no guarantees that it works with all: https://github.com/archen/rules_oci/pull/1

michaeljs1990:
Tested out your above patch and although it looks like it might break something else based on the tests it did let me keep developing on my linux/arm64 machine. I don't think I will ever purchase another arm machine 😞

archen:
[...]

PR #337, opened 2023-08-18
replace jq with yq for improved cross-platform support

archen:
opened by request: https://github.com/archen/rules_oci/pull/1#issuecomment-1674659207

The `jq` toolchain does not support as many architectures as `yq` and `yq` supports (for the most-part) a super-set of the `jq` capabilities. This change introduces the use of `yq` as a `jq` replacement for downstream users with build-worker architectures not natively supported by `jq`.

fixes: https://github.com/bazel-contrib/rules_oci/issues/253

requires: https://github.com/aspect-build/bazel-lib/releases/tag/v1.33.0

```starlark
# WORKSPACE
load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
maybe(
    http_archive,
    name = "aspect_bazel_lib",
    sha256 = "d488d8ecca98a4042442a4ae5f1ab0b614f896c0ebf6e3eafff363bcc51c6e62",
    strip_prefix = "bazel-lib-1.33.0",
    url = "https://github.com/aspect-build/bazel-lib/archive/refs/tags/v1.33.0.tar.gz",
)

maybe(
    git_repository,
    name = "rules_oci",
    remote = "https://github.com/archen/rules_oci",
    commit = "880ba021b38732cb022207af76c752b87e20deec",
    shallow_since = "1692363244 -0400",
)
```

With the updated rules loaded, you can also `stamp` images with the following modification:
```starlark
"""Helper for stamping version control info into OCI image tags

Modified from: https://github.com/bazel-contrib/rules_oci/blob/e426977dcf6bba3017cdc6bb8b1aeca7bc35b1d5/examples/push/stamp_tags.bzl
"""

load("@aspect_bazel_lib//lib:yq.bzl", "yq")
load("@bazel_skylib//lib:types.bzl", "types")

def stamp_tags(name, remote_tags, **kwargs):
    """Wrapper macro around the [yq](https://docs.aspect.build/rules/aspect_bazel_lib/docs/yq) rule.

    Produces a text file that can be used with the `remote_tags` attribute of [`oci_push`](#oci_push).

    Each entry in `remote_tags` is typically either a constant like `my-repo:latest`, or can contain a stamp expression.
    The latter can use any key from `bazel-out/stable-status.txt` or `bazel-out/volatile-status.txt`.
    See https://docs.aspect.build/rules/aspect_bazel_lib/docs/stamping/ for details.

    The yq `//` default operator is useful for returning an alternative value for unstamped builds.

    For example, if you use the expression `($stamp.BUILD_EMBED_LABEL // "0.0.0")`, this resolves to
    "0.0.0" if stamping is not enabled. When built with `--stamp --embed_label=1.2.3` it will
    resolve to `1.2.3`.

    Args:
        name: name of the resulting yq target.
        remote_tags: list of yq expressions which result in a string value, see docs above
        **kwargs: additional named parameters to the yq rule.
    """
    if not types.is_list(remote_tags):
        fail("remote_tags should be a list")

    _maybe_quote = lambda x: x if "\"" in x else "\"{}\"".format(x)
    yq(
        name = name,
        srcs = [],
        outs = ["_{}.tags.txt".format(name)],
        expression = "|".join([
            "load(strenv(STAMP)) as $stamp",
            ",".join([_maybe_quote(t) for t in remote_tags]),
        ]),
        **kwargs
    )
```


thesayyn:
I don't think YQ is a drop-in replacement for JQ but let see what CI says.

archen:
@thesayyn yeah it's possible, I've been using the included modifications in my fork for a few weeks with no issues, but I also haven't exercised every rule.

archen:
well, speak of the devil https://github.com/bazel-contrib/rules_oci/actions/runs/5903016626/job/16018454016?pr=337#step:6:93

archen:
sorry, I haven't had time to address this over the weekend. I overlooked the `image.sh.tpl` implications and never encountered failures in my own use-cases as this logic is either not exercised or failing silently in my project(s).

As far as the path forward goes... if I'm able to get the filter statements working with `yq` as equivalent with the current `jq` usage... are you opposed to having the variable substitutions loaded from env vars versus the `--arg` approach used in `jq`? I would use the `env` operators from `yq` documented here: https://mikefarah.gitbook.io/yq/operators/env-variable-operators

thesayyn:
> sorry, I haven't had time to address this over the weekend. I overlooked the `image.sh.tpl` implications and never encountered failures in my own use-cases as this logic is either not exercised or failing silently in my project(s).
> 
> As far as the path forward goes... if I'm able to get the filter statements working with `yq` as equivalent with the current `jq` usage... are you opposed to having the variable substitutions loaded from env vars versus the `--arg` approach used in `jq`? I would use the `env` operators from `yq` documented here: https://mikefarah.gitbook.io/yq/operators/env-variable-operators

That should be okay as there's no equivalent `--arg` replacement in YQ.

alexeagle:
Note, a new jq release came out that finally ships darwin-arm64 binaries so this is less important.
I'll close this for now, @kormide is working on the bazel-lib change to support newer jq.

PR #338, opened 2023-08-23
Allow passing repository for oci_push in from a file

mboulton-fathom:
As mentioned in the updated docs this is useful if you want to push to a different repository (eg, separate dev/prod repositories) based on a stamp variable.

### DIFF
--- a/examples/push/BUILD.bazel
+++ b/examples/push/BUILD.bazel
@@ -16,6 +16,13 @@ oci_push(
 )
 
 oci_push(
+    name = "push_image_repository_file",
+    image = ":image",
+    remote_tags = ["latest"],
+    repository_file = ":repository.txt",
+)
+
+oci_push(
     name = "push_image_wo_tags",
     image = ":image",
     repository = "index.docker.io/<ORG>/image",
@@ -52,11 +59,13 @@ sh_test(
         "$(LAUNCHER_WRAPPER)",
         "$(location :push_image)",
         "$(location :push_image_index)",
+        "$(location :push_image_repository_file)",
         "$(location :push_image_wo_tags)",
     ],
     data = [
         ":push_image",
         ":push_image_index",
+        ":push_image_repository_file",
         ":push_image_wo_tags",
         "@oci_crane_toolchains//:current_toolchain",
         "@oci_zot_toolchains//:current_toolchain",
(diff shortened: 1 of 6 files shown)


examples/push/BUILD.bazel
```diff
@@ -15,6 +15,13 @@ oci_push(
     repository = "index.docker.io:9899/<ORG>/image",
 )
 
+oci_push(
+    name = "push_image_repository_file",
+    image = ":image",
+    remote_tags = ["latest"],
+    repository_file = ":repository.txt",
```
alexeagle, line 22: nit: it would be a more useful example if you stamp this file (use `expand_template` from aspect_bazel_lib)

oci/private/push.sh.tpl
```diff
@@ -6,12 +6,17 @@ readonly YQ="{{yq_path}}"
...
+REPOSITORY=""
+if [ -f $REPOSITORY_FILE ] ; then
+  REPOSITORY=$(tr -d '\n' < "$REPOSITORY_FILE")
+fi
 
 # set $@ to be FIXED_ARGS+$@
 ALL_ARGS=(${FIXED_ARGS[@]} $@)
 set -- ${ALL_ARGS[@]}
 
-REPOSITORY="{{repository}}"
```
alexeagle, line 14: Looks like we no longer interpolate the `repository` variable anywhere in this template? That doesn't seem correct.

alexeagle (commented):
Thanks! I think this feature makes sense.

Ideally we'd only have the `repository_file` attribute on the rule, but that would be a breaking change so I think we do need both now.

oci/private/push.sh.tpl
mboulton-fathom: When I looked at the file it seemed like the repository variable was never stamped anyway, I assumed it was an oversight - it is currently stamped into the `fixed_args` variable instead

examples/push/BUILD.bazel
mboulton-fathom: That function isn't in the version of the aspect_bazel_lib (or bazel_skylib) specified in the dependencies, and there's a big warning in oci/dependencies.bzl saying that changing it would be a breaking change, so I'm hesitant to change that.

oci/private/push.sh.tpl
alexeagle: I see, there was no `repository` variable being interpolated to begin with.

examples/push/BUILD.bazel
alexeagle: This was just a nit, no need to hold up the merge on it.

alexeagle (approved):

PR #340, opened 2023-08-24
feat: add Azure Container Registry oci_pull support

HDYA:
* update how `_WWW_AUTH` is used to match registry
    * use placeholder `{registry}` in field `realm` and `service`
    * also check for suffix match
* update how authentication response is handled
    * check for field `access_token` 
    * fail if no token found

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -137,10 +137,17 @@ def _get_token(rctx, state, registry, repository):
         )
         auth_raw = rctx.read("www-authenticate.json")
         auth = json.decode(auth_raw)
+        token = ""
+        if "token" in auth:
+            token = auth["token"]
+        if "access_token" in auth:
+            token = auth["access_token"]
+        if token == "":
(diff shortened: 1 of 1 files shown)


oci/private/pull.bzl
```diff
@@ -40,19 +40,24 @@ _WWW_AUTH = {
...
         "scope": "repository:{repository}:pull",
-        "service": "ghcr.io/token",
+        "service": "{registry}/token",
     },
     "cgr.dev": {
-        "realm": "cgr.dev/token",
-        "service": "cgr.dev",
+        "realm": "{registry}/token",
         "scope": "repository:{repository}:pull",
+        "service": "{registry}s",
```
alexeagle, line 55: the trailing "s" on this line looks suspicious. I wonder if we have e2e tests that fetch from each registry so we know they all work? it would be good to add a simple Azure CR pull in one of the e2e/examples as well.

thesayyn: i believe that is a typo.

PR #341, opened 2023-08-28
Add `toolchain` param to the affected actions

kotlaja:
This is a step forward for the migration of Automatic Exec Groups. These actions don't use an executable/tool from a toolchain. Hence, `None` should be set.

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -194,6 +194,7 @@ def _oci_image_impl(ctx):
         tools = [crane.crane_info.binary, registry.registry_info.launcher, registry.registry_info.registry, jq.jqinfo.bin],
         mnemonic = "OCIImage",
         progress_message = "OCI Image %{label}",
+        toolchain = None,
     )
 
     return [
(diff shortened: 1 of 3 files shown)


thesayyn:
isn't this a breaking change from bazel's POV?

kotlaja:
It's not a breaking change since `toolchain` param is still not effective - Automatic Exec Groups are still not enabled in Bazel.

Still, it would be an obstacle if you're supporting Bazel version older than Bazel 6.0 since the `toolchain` parameter is visible from 6.0 onwards. Which Bazel version do you support in rules_oci?

alexeagle:
We test against Bazel 5 also https://github.com/bazel-contrib/rules_oci/blob/main/.github/workflows/ci.yaml#L27
which is supported for another 15 months https://bazel.build/release#support-matrix

thesayyn:
I will convert this to a draft and get back to it when we drop support for bazel 5.

thesayyn (approved):

PR #348, opened 2023-09-05
feat: add mnemonic to the oci_image_index action

lbcjbb:


alexeagle (approved):
@thesayyn I think we need to do a pass over the repo to add mnemonics to the other actions too, but no need to block this one on that project.

Issue #282, opened 2023-07-04
Add support for OCI artifacts


hanikesn:
`rules_oci` provides good support for OCI images already. Other OCI artifacts like helm charts and [ORAS](https://oras.land/) is becoming more and more popular. A natural extension for `rules_oci` is adding support for building and pushing generic OCI artifacts. To keep the scope limited it's sufficient to support the [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md)/[OCI Artifacts Spec](https://github.com/opencontainers/artifacts) and leave application specifics up to 3rd party rules.

If support for OCI artifacts is in scope for `rules_oci` then I'd be happy for some pointers so I can start with a POC. It certainly looks like zot [supports](https://zotregistry.io/v1.4.3/user-guides/user-guide-datapath/) generic OCI artifacts already.

alexeagle:
Interesting suggestion, I do believe based on the name of this ruleset and the scope we've indicated in the past, that we want to fully support the official OCI specs.

hanikesn:
Unfortunately I didn't have time to look into this yet, but wanted to highlight argocd's initative to support oci artifacts:
- https://github.com/argoproj/argo-cd/blob/master/docs/proposals/native-oci-support.md
- https://github.com/argoproj/argo-cd/discussions/8799

thesayyn:
With 2.0, we are going to start using https://github.com/regclient/regclient/blob/main/docs/regctl.md which supports OCI artifacts. 

We might consider supporting it after 2.0.

hanikesn:
For anyone curious initial [ArgoCD OCI](https://github.com/argoproj/argo-cd/pull/18646) support just got merged, so reproducible infrastructure is in finally in reach!

PR #349, opened 2023-09-05
feat: oci_pull supports loading Helm charts

sobotklp:
This MR shows an example of loading a Helm chart using `oci_pull` from a repository (for example, ECR)

For example, in `WORKSPACE`:
```
oci_pull(
    name = "dependency-helm-chart",
    digest = "sha256:865647e7904a1e4b4e130b423d20b34dcc89a5a37b7bc82e30ff100c30cfb86f",
    registry = "111222333444.dkr.ecr.us-east-2.amazonaws.com",
    repository = "helm/dependency-charts",
)
```

This is related to issue #282

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -183,6 +183,7 @@ _SUPPORTED_MEDIA_TYPES = {
     "manifest": [
         "application/vnd.docker.distribution.manifest.v2+json",
         "application/vnd.oci.image.manifest.v1+json",
+        "application/vnd.cncf.helm.config.v1+json",
     ],
 }
 
@@ -358,7 +359,12 @@ def _oci_pull_impl(rctx):
 
     mf, mf_len = downloader.download_manifest(rctx.attr.identifier, "manifest.json")
 
-    if mf["mediaType"] in _SUPPORTED_MEDIA_TYPES["manifest"]:
+    if "mediaType" in mf:
+        media_type = mf["mediaType"]
+    else:
+        media_type = mf["config"]["mediaType"]
+
+    if media_type in _SUPPORTED_MEDIA_TYPES["manifest"]:
         if rctx.attr.platform:
             fail("{}/{} is a single-architecture image, so attribute 'platforms' should not be set.".format(rctx.attr.registry, rctx.attr.repository))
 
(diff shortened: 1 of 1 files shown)


oci/private/pull.bzl
```diff
@@ -358,7 +359,12 @@ def _oci_pull_impl(rctx):
 
     mf, mf_len = downloader.download_manifest(rctx.attr.identifier, "manifest.json")
 
-    if mf["mediaType"] in _SUPPORTED_MEDIA_TYPES["manifest"]:
+    if "mediaType" in mf:
+        media_type = mf["mediaType"]
+    else:
+        media_type = mf["config"]["mediaType"]
```
alexeagle, line 365: it feels like this now needs a final else clause when the media_type couldn't be found in either of these two paths within the data structure

oci/private/pull.bzl
```diff
@@ -391,6 +397,11 @@ def _oci_pull_impl(rctx):
         if hash not in tars:
             tars.append(hash)
 
+    if "mediaType" in image_mf:
```
alexeagle, line 400: could you extract a function here so we only need to have this logic once?

oci/private/pull.bzl
```diff
@@ -183,6 +183,7 @@ _SUPPORTED_MEDIA_TYPES = {
     "manifest": [
         "application/vnd.docker.distribution.manifest.v2+json",
         "application/vnd.oci.image.manifest.v1+json",
+        "application/vnd.cncf.helm.config.v1+json",
```
alexeagle, line 186: can you come up with an example we could add to /WORKSPACE so there's a local reproduction we can use to test? I tried the literal helm chart you put in the PR description but get a 401 fetching that, I assume it's in a private registry.

alexeagle (commented):
Thanks for the PR!!

Could you also try updating a documentation spot on oci_pull so it's easier for users to discover this ability?

sobotklp:
@alexeagle no worries. I will spend some time adding more documentation and test cases as well. 

I'm also having trouble coming up with a general use case for someone (besides me) who wants to use this functionality. Downloading the OCI helm chart is easy enough, but what to do with it afterwards? Any guidance you could give here would be appreciated.

thesayyn:
That's the rest of the iceberg, aka OCI artifacts, which we are not yet ready to support. Neither rules_oci nor crane is able to do anything with artifacts yet. 

On top of my head, I can think of using parts of a pulled helm chart to create a new one and push it with `oci_push`. 

Currently, rules_oci only conforms to the non-artifact part of https://github.com/opencontainers/image-spec, but we are considering adding full support for artifacts in the future.

anguslees:
> Downloading the OCI helm chart is easy enough, but what to do with it afterwards?

I want this, fwiw!  I have my own bazel code that can do `helm template $chart.tgz`, but I would like to use rules_oci to actually fetch the chart file from an OCI repo.

thesayyn:
I don't think we can accept this now, as the whole story for OCI artifacts is missing here. We need to work on https://github.com/bazel-contrib/rules_oci/issues/282 first before we can add this.

PR #350, opened 2023-09-07
fix: propagate tags for macros in `defs.bzl`

ajaypbrt:
Tags were not being propagated to targets declared by the macros in `defs.bzl`. This fixes that by using [propagate_well_known_tags](https://github.com/aspect-build/bazel-lib/blob/main/docs/utils.md#propagate_well_known_tags) to propagate a relevant subset of tags to all targets.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -90,7 +90,7 @@ oci_image(
 ## oci_image
 
 <pre>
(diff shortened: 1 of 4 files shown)


alexeagle (approved):

ewhauser:
@ajaypbrt @alexeagle: We were passing the "containers" tag to `rules_oci` targets so that the team could selectively do:

```
 build --build_tag_filters=-containers
 test --test_tag_filters=-containers
```

locally to skip long tarball times. I think this PR broke the ability to do this?

ajaypbrt:
Yes, it appears so, sorry about that. @alexeagle it looks like some of the targets were passing `**kwargs` on, so before would have gotten the full set of provided tags. I wonder if we should just pass all tags through to all targets instead of filtering by `propagate_well_known_tags`?

PR #352, opened 2023-09-11
fix: remove tag argument from crane image mutate call

### LINKED ISSUE #351: oci_image target name cannot contain capital letters
When using the oci_image with a target that contains a capital letter (such as an acronym) in its name, the build will fail.

e.g.

```starlark
oci_image(
    name = "name_OCI",
    ...
)
``` 

```
INFO: Analyzed 26 targets (3 packages loaded, 125 targets configured).
ERROR: <snip>: OCI Image //<snip>:name_OCI failed: (Exit 1): image_name_OCI.sh failed: error executing command (from target //<snip>:name_OCI) bazel-out/k8-opt/bin/<snip>/image_name_OCI.sh mutate oci:layout/bazel-out/k8-opt/bin/<snip> --tag oci:registry/name_OCI ... (remaining 6 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
2023/09/11 15:44:51 serving on port 37969
2023/09/11 15:44:51 GET /v2/
2023/09/11 15:44:51 HEAD /v2/oci/layout/blobs/sha256:<snip> 404 BLOB_UNKNOWN Unknown blob
2023/09/11 15:44:51 HEAD /v2/oci/layout/blobs/sha256:<snip> 404 BLOB_UNKNOWN Unknown blob
2023/09/11 15:44:51 HEAD /v2/oci/layout/blobs/sha256:<snip> 404 BLOB_UNKNOWN Unknown blob
2023/09/11 15:44:51 POST /v2/oci/layout/blobs/uploads/
2023/09/11 15:44:51 POST /v2/oci/layout/blobs/uploads/
2023/09/11 15:44:51 POST /v2/oci/layout/blobs/uploads/
2023/09/11 15:44:51 PATCH /v2/oci/layout/blobs/uploads/<snip>
2023/09/11 15:44:51 PATCH /v2/oci/layout/blobs/uploads/<snip>
2023/09/11 15:44:51 PATCH /v2/oci/layout/blobs/uploads/<snip>
2023/09/11 15:44:52 PUT /v2/oci/layout/blobs/uploads/<snip>?digest=sha256%3A<snip>
2023/09/11 15:44:52 PUT /v2/oci/layout/blobs/uploads/<snip>?digest=sha256%3A<snip>
2023/09/11 15:44:52 PUT /v2/oci/layout/blobs/uploads/<snip>?digest=sha256%3A<snip>
2023/09/11 15:44:52 PUT /v2/oci/layout/manifests/latest
/tmp/tmp.pmoV8XgJwE
2023/09/11 15:44:52 GET /v2/
2023/09/11 15:44:52 GET /v2/oci/layout/manifests/sha256:<snip>
2023/09/11 15:44:52 GET /v2/oci/layout/blobs/sha256:<snip>
Error: parsing 127.0.0.1:37969/name_OCI: could not parse reference: 127.0.0.1:37969/name_OCI
INFO: Elapsed time: 14.974s, Critical Path: 11.56s
INFO: 18 processes: 16 internal, 2 linux-sandbox.
FAILED: Build did NOT complete successfully
```

This occurs because docker image names cannot contain capital letters, but I don't see why this naming constraint needs to be propagated to the Bazel name of the target.

A possible solution would be to lowercase the target name when talking to registry, which is possibly as simple as changing a line here with `.lower()` (not tested!):

https://github.com/bazel-contrib/rules_oci/blob/ea315e1194b45878365c0fa81fd6172e42fb319d/oci/private/image.bzl#L138

Although stepping back a bit, I'm not sure why the target name needs to be used in this action at all, as you could instead just pass a known sentinel into crane (e.g. `oci:target`)? It seems it just needs _some_ name to link up the `mutate` and `pull` commands and that name needn't depend on the target name. Stepping back even further, I'm not convinced the `--tag` arg is actually doing anything useful here, since the `REF` used to link up those commands is not the tag anyway but rather the output of crane itself...

aaliddell:
This argument is not providing a useful purpose and breaks when oci_image targets have a capital letter in their name. Removing this does not break the image build since the ref used between the mutate and pull is sourced directly from the output of crane

Closes #351 

Within the image.sh script, this slightly changes the value of the `$REF` variable used from the initial mutate call. Prior to this change, these look like:

`127.0.0.1:42477/<target name>@sha256:<some sha hash>`

Now these inherit the source name for the mutation and look like:

`127.0.0.1:42477/oci/layout@sha256:<some sha hash>` or `127.0.0.1:42477/oci/empty_base@sha256:<some sha hash>`

If you would prefer, the alternate fix is to use a static sentinel value for tag (e.g. `--tag oci:registry/target`), which produces refs like so:

`127.0.0.1:42477/target@sha256:<some sha hash>`

The final output is the same regardless of the value of `$REF` used internally

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -134,8 +134,6 @@ def _oci_image_impl(ctx):
     args.add_all([
         "mutate",
         base,
-        "--tag",
-        "oci:registry/{}".format(ctx.label.name),
     ])
 
     # add platform


alexeagle:
Thanks for the detailed explanation. I think this looks right but want 👀 from @thesayyn before merging

thesayyn:
can we add an example for this to prevent us from breaking it again in the future?

thesayyn:
> can we add an example for this to prevent us from breaking it again in the future?

I'll do that as a follow-up

thesayyn (approved):

PR #356, opened 2023-09-14
example: make sure capital letter targets work

thesayyn:


alexeagle (approved):
"various" 🤨

thesayyn:
> "various" 🤨

feel like we have some many directories for simple cases like this. I'll dump anything small into this folder from now on.

PR #357, opened 2023-09-15
feat: add Gitlab Container Registry oci_pull support

avdv:
This adds www-authenticate information for registry.gitlab.com.

### DIFF
--- a/fetch.bzl
+++ b/fetch.bzl
@@ -144,6 +144,13 @@ def fetch_images():
         reproducible = False,
     )
 
+    oci_pull(
+        name = "gitlab_assets_ce",
+        # tag = "v15-11-0-ee",
+        image = "registry.gitlab.com/gitlab-org/gitlab/gitlab-assets-ce",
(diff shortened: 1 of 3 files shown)


thesayyn:
could you add an oci_pull target to `fetch.bzl` and add it to [this target](https://github.com/bazel-contrib/rules_oci/blob/2257c18c364c72b3bfa4bd32795d7902bef6ff81/oci/tests/BUILD.bazel#L50-L62) so that we assert we can fetch from these registries?

avdv:
> could you add an oci_pull target to `fetch.bzl` and add it to [this target](https://github.com/bazel-contrib/rules_oci/blob/2257c18c364c72b3bfa4bd32795d7902bef6ff81/oci/tests/BUILD.bazel#L50-L62) so that we assert we can fetch from these registries?

Done. (I tried to find an image which is publicly available and not too large, this one is around 250MiB)

thesayyn (approved):

Issue #248, opened 2023-05-19
How to push image to multiple repositories?


prestonvanloon:
I am trying to replicate logic from rules_docker where I can have a `container_bundle` given to a `docker_push`.

```python
container_bundle(
    name = "image_bundle",
    images = {
        "gcr.io/prysmaticlabs/prysm/beacon-chain:latest": ":image_with_creation_time",
        "index.docker.io/prysmaticlabs/prysm-beacon-chain:latest": ":image_with_creation_time",
    },
    tags = ["manual"],
    visibility = ["//beacon-chain:__pkg__"],
)

docker_push(
    name = "push_images",
    bundle = ":image_bundle",
    tags = ["manual"],
    visibility = ["//beacon-chain:__pkg__"],
)
```

The end result was that I pushed an image to multiple repositories with a single target.

thesayyn:
```
oci_push(
    name = "push",
    image = ":image_with_creation_time",
    repository = "gcr.io/prysmaticlabs/prysm/beacon-chain",
    remote_tags = ["latest"]
)

oci_push(
    name = "push",
    image = ":image_with_creation_time",
    repository = "index.docker.io/prysmaticlabs/prysm-beacon-chain",
    remote_tags = ["latest"]
)
```

this can be done by simply having two oci_push targets.

prestonvanloon:
> this can be done by simply having two oci_push targets.

I understand that, but it doesn't scale well. In our case, we have 4 repository / tag variations per image.  Even with macros to expand to multiple targets, it is not possible push all from one target. (https://github.com/bazelbuild/bazel/issues/10855)

I'm looking for feature parity with the functionality of `docker_push` from rules_docker.

thesayyn:
> Even with macros to expand to multiple targets, it is not possible to push all from one target / bazel command.

this could be done by running all oci_push targets in a sh_binary target. 

> I'm looking for feature parity with the functionality of `docker_push` from rules_docker.

this is essentially what container_bundle does as I said above. 

Unfortunately, we are `-rc` so we can not introduce breaking changes.

alexeagle:
You can use https://github.com/keith/rules_multirun to make a single `bazel run`nable target.

You could write a macro that emulates `container_bundle` or even spell it out in a BUILD file:

```
load("@rules_oci//oci:defs.bzl", "oci_image", "oci_push")
load("@rules_multirun//:defs.bzl", "command", "multirun")

oci_image(
    name = "image",
[...]

PR #360, opened 2023-09-17
Draft: feat: oci_bundle

hunshcn:
similar to crane pull --format=oci image1 image2 output-dir, but locally

Preparation of https://github.com/bazel-contrib/rules_oci/issues/248

( I create https://github.com/google/go-containerregistry/pull/1787 to support push bundle. )

but we should support bundle first.

### DIFF
--- a/examples/bundle/BUILD.bazel
+++ b/examples/bundle/BUILD.bazel
@@ -1,0 +1,48 @@
+load("//oci:defs.bzl", "oci_bundle", "oci_image", "oci_tarball")
+load("@aspect_bazel_lib//lib:testing.bzl", "assert_json_matches")
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@container_structure_test//:defs.bzl", "container_structure_test")
+
+oci_image(
+    name = "image_arm64",
+    architecture = "arm64",
+    entrypoint = ["/custom_bin"],
+    os = "linux",
+)
+
+oci_image(
+    name = "image_amd64",
+    architecture = "amd64",
+    entrypoint = ["/custom_bin"],
+    os = "linux",
+)
+
+images = {
+    "app:arm64": ":image_arm64",
+    "app:amd64": ":image_amd64",
+}
+
+oci_bundle(
+    name = "bundle",
+    images = images,
+)
+
+genrule(
+    name = "bundle_manifest",
+    srcs = [":bundle"],
+    outs = ["index.json"],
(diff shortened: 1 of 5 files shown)


hunshcn:
use ln instead of cp, in https://github.com/bazel-contrib/rules_oci/pull/359

hunshcn:
The conten of this pr has been completed, but I can't do codegen.

> bazel run //:gazelle

```plain
INFO: Analyzed target //:gazelle (2 packages loaded, 63 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 0.270s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/gazelle
grep: MANIFEST: No such file or directory
/root/.cache/bazel/_bazel_root/37ac864eff3240afc36de450593d425f/execroot/rules_oci/bazel-out/k8-fastbuild/bin/gazelle: warning: could not locate GOROOT used by rules_go
gazelle: /build/rules_oci/e2e/auth/BUILD.bazel: unknown directive: gazelle:prefix
```

any help?

thesayyn:
Hey. thanks for the PR. we already have an oci_image_index rule that does basically what this does. About pushing to multiple repositories, you can always use multiple oci_push targets.  We will be happy to see the oci_push feature that supports pushing to multiple repositories efficiently.

thesayyn:
Expanding the Public API is the last thing we want to do here. Please see our motivation behind this https://github.com/bazel-contrib/rules_oci#bazel-rules-for-oci-containers

hunshcn:
In some offline deployment scenarios, we need to bundle all the images, copy them to the machine through mobile media, and then push them to registry through tools such as crane.

Do you think this should be a private rule because of low frequency of use?

hunshcn:
I think it is reasonable to provide a similar ability to `crane pull multiimag...  bundle-dir  .

thesayyn:
> Do you think this should be a private rule because of low frequency of use

Not really. I was implying that this should not go into rules_oci at all. Once it is in our public API, we are on the hook to maintain it forever.  Given that we are not trying to be a drop-in replacement for rules_docker it's less likely that we will have a rule similar to container_bundle.  I would recommend

hunshcn:
move to https://github.com/hunshcn/rule_oci_bundle

PR #364, opened 2023-09-18
feat: add custom no_match_error to oci_alias

### LINKED ISSUE #178: oci_pull - allow you to list the platforms of a multi-arch image
Something related to reproducible = True - we give an *_unpinned//:pin target

thesayyn:
Now when there is no matching platform in oci_alias, users will see an error with list of available platforms from the image. 

```
alias(
    name = "chainguard_static",
    actual = select(
        {
           "@platforms//cpu:x86_64": "@chainguard_static_linux_amd64", 
           "@platforms//cpu:arm64": "@chainguard_static_linux_arm64", 
           "@platforms//cpu:ppc": "@chainguard_static_linux_ppc64le", 
           "@platforms//cpu:s390x": "@chainguard_static_linux_s390x"
         },
        no_match_error = """could not find an image matching the target platform. \navailable platforms are "linux/386", "linux/amd64", "linux/arm/v6", "linux/arm/v7", "linux/arm64", "linux/ppc64le", "linux/s390x" """,
    ),
)
```

fixes #178

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -444,7 +444,8 @@ _MULTI_PLATFORM_IMAGE_ALIAS_TMPL = """\
 alias(
     name = "{target_name}",
     actual = select(
-        {platform_map}
+        {platform_map},
+        no_match_error = \"\"\"could not find an image matching the target platform. \\navailable platforms are {available_platforms} \"\"\",
     ),
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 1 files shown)


oci/private/pull.bzl
```diff
@@ -493,22 +494,25 @@ def _oci_alias_impl(rctx):
...
-            for submanifest in manifest["manifests"]:
-                parts = [submanifest["platform"]["os"], submanifest["platform"]["architecture"]]
-                if "variant" in submanifest["platform"]:
-                    parts.append(submanifest["platform"]["variant"])
-                platforms.append('"{}"'.format("/".join(parts)))
-            optional_platforms = "'add platforms {}'".format(" ".join(platforms))
+        if len(available_platforms):
+            optional_platforms = "'add platforms {}'".format(" ".join(available_platforms))
 
-        is_bzlmod = hasattr(rctx.attr, "bzlmod_repository") and rctx.attr.bzlmod_repository
```
alexeagle, line 511: how come we only set this variable conditionally on `_is_tag and reproducible` now, when we did it always before?

oci/private/pull.bzl
```diff
@@ -529,6 +533,7 @@ buildozer 'set digest "sha256:{digest}"' 'remove tag' 'remove platforms' {option
         build = _MULTI_PLATFORM_IMAGE_ALIAS_TMPL.format(
             name = rctx.attr.name,
             target_name = rctx.attr.target_name,
+            available_platforms = ", ".join(available_platforms),
```
alexeagle, line 536: I think you should do the quote wrapping here as well, so all the "presentation" is together. 

The `'"{}"'.format` above is pretty awkward syntax too

oci/private/pull.bzl
thesayyn: no, we didn't do it always before. it was only set when the non-reproducible warning was printed which would have been printed only when `_is_tag and reproducible`

thesayyn: now we fetch the manifest to know which platforms are available so that we can put it into build file.

oci/private/pull.bzl
thesayyn: I am not sure I follow.

thesayyn: `'"{}"'.format` is for building the buildozer command, so I wanted to use the same array

thesayyn: that's why I have triple quotes in the template.

alexeagle: ah okay I didn't see that the quoted version is used in another place

PR #366, opened 2023-09-19
fix: Propagate all tags, not just well known ones

ajaypbrt:
In trying to fix tag propagation, #350 broke tag propagation behavior for some targets which did have tags forwarded properly. The reason the behavior changed is because while we are still propagating tags in #350, we started passing them through [propagate_well_known_tags](https://github.com/aspect-build/bazel-lib/blob/main/docs/utils.md#propagate_well_known_tags) which caused some issues, e.g. https://github.com/bazel-contrib/rules_oci/pull/350#issuecomment-1724530364.

This restores the old behavior for the targets which did have tags propagated, while also propagating all tags for the new targets.


ajaypbrt:
cc: @alexeagle @ewhauser

ewhauser:
Thank you @ajaypbrt!

thesayyn (approved):
I think we should also support passing common attributes. but this is good enough for this PR.

PR #362, opened 2023-09-18
refactor: clean up variables and messages

thesayyn:
Rename variables to remove unnecessary reassigments

Issue #138, opened 2023-03-30
refactor oci_pull to be starlark testable


thesayyn:
See: https://github.com/bazel-contrib/rules_oci/pull/136#discussion_r1153496913

thesayyn:
We have already done this through many iterations to oci_pull.

PR #368, opened 2023-09-19
fix: add variant to manifest entry 

thesayyn:
part of #362

### DIFF
--- a/oci/tests/BUILD.bazel
+++ b/oci/tests/BUILD.bazel
@@ -16,7 +16,12 @@ IMAGES_TO_TEST = {
     genrule(
         name = "pull_{}".format(repo_name),
         outs = [repo_name],
-        cmd = "$(CRANE_BIN) pull {reference} $@ --format=oci --platform={platform}".format(
+        cmd = """
+$(CRANE_BIN) pull {reference} $@ --format=oci --platform={platform}
+cat "$@/index.json" | $(JQ_BIN) --sort-keys -j '.' > "$@/index-tmp.json"
+rm "$@/index.json"
+mv "$@/index-tmp.json" "$@/index.json"
+""".format(
             platform = _PLATFORM,
             reference = reference,
         ),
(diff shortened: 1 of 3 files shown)


oci/tests/BUILD.bazel
```diff
@@ -16,7 +16,12 @@ IMAGES_TO_TEST = {
     genrule(
         name = "pull_{}".format(repo_name),
         outs = [repo_name],
-        cmd = "$(CRANE_BIN) pull {reference} $@ --format=oci --platform={platform}".format(
+        cmd = """
+$(CRANE_BIN) pull {reference} $@ --format=oci --platform={platform}
+cat "$@/index.json" | $(JQ_BIN) --sort-keys -j '.' > "$@/index-tmp.json"
```
alexeagle, line 21: needs some comments about doing this to make it match the json.encode_indent output that starlark is making

alexeagle (commented):
Could you help me understand why it's an improvement? It feels approx the same performance and complexity overall

thesayyn:
> Could you help me understand why it's an improvement? It feels approx the same performance and complexity overall

the `json.encode_ident` does not provide any performance improvement over the templated json, it is purely nice to have to allow more properties to be added to the platform sub-object conditionally without templating tricks. https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions
 I can revert that. 
 
That said the use of write_file is unnecessary here and adds some performance overhead. eg: spawning some actions to write a file which is pure overhead given that we already write the same content to the BUILD file.

thesayyn:
actually, now that I think about it more, I will keep that part the same just make it more ergonomic. eg: instead of two templates, just use one.

oci/tests/BUILD.bazel
thesayyn: not relevant anymore.

thesayyn:
@alexeagle i took back the templated strings back and made an util function that can add `variant` as well. now we match `crane pull` 100% of the cases.  

now we have the ability to write a starlark unit test for the util function.  see #138

alexeagle (approved):

Issue #369, opened 2023-09-20
[1.35] Breaks testonly oci_images


hanikesn:
After upgrading from 1.31 to 1.35 our `oci_images` marked with `testonly` stopped working:
```
ERROR: /__w/BUILD:31:30: in directory_path rule //X:X-e2e-tests_image_index_json: non-test target '//X:X-e2e-tests_image_index_json' depends on testonly target '//X:X-e2e-tests_image' and doesn't have testonly attribute set
```
I suspect #346 to introduce the breaking change.

thesayyn:
This is fixed by #370

PR #370, opened 2023-09-20
fix: forward well known attributes to macro expanded targets

reltuk:
As per #369, trying to put a test in an `oci_image` target by making it `testonly = True` is currently broken.

This PR just adds the common rule attribute`testonly` to the `oci_image` and `oci_tarball` macros and forwards its value along to all the rules they instantiate.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -118,6 +118,6 @@ This is similar to the same-named target created by rules_docker's `container_im
 | <a id="oci_image-env"></a>env |  Environment variables provisioned by default to the running container. See documentation above.   |  <code>None</code> |
 | <a id="oci_image-cmd"></a>cmd |  Command & argument configured by default in the running container. See documentation above.   |  <code>None</code> |
 | <a id="oci_image-entrypoint"></a>entrypoint |  Entrypoint configured by default in the running container. See documentation above.   |  <code>None</code> |
-| <a id="oci_image-kwargs"></a>kwargs |  other named arguments to [oci_image_rule](#oci_image_rule) and [common rule attribtes](https://bazel.build/reference/be/common-definitions#common-attributes).   |  none |
+| <a id="oci_image-kwargs"></a>kwargs |  other named arguments to [oci_image_rule](#oci_image_rule) and [common rule attributes](https://bazel.build/reference/be/common-definitions#common-attributes).   |  none |
 
 
--- a/docs/image.md
+++ b/docs/image.md
@@ -90,7 +90,7 @@ oci_image(
 ## oci_image
 
 <pre>
-oci_image(<a href="#oci_image-name">name</a>, <a href="#oci_image-labels">labels</a>, <a href="#oci_image-annotations">annotations</a>, <a href="#oci_image-env">env</a>, <a href="#oci_image-cmd">cmd</a>, <a href="#oci_image-entrypoint">entrypoint</a>, <a href="#oci_image-tags">tags</a>, <a href="#oci_image-kwargs">kwargs</a>)
+oci_image(<a href="#oci_image-name">name</a>, <a href="#oci_image-labels">labels</a>, <a href="#oci_image-annotations">annotations</a>, <a href="#oci_image-env">env</a>, <a href="#oci_image-cmd">cmd</a>, <a href="#oci_image-entrypoint">entrypoint</a>, <a href="#oci_image-kwargs">kwargs</a>)
 </pre>
 
 Macro wrapper around [oci_image_rule](#oci_image_rule).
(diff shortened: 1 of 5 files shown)


thesayyn:
I feel like we shouldn't have to do this one by one. I'll check if there's a macro that could return `well-known/common kwargs` so that we can forward them to all targets.

alexeagle:
https://github.com/bazel-contrib/rules_oci/pull/350 is an example

thesayyn:
> Is the proposal to make a `propagate_common_rule_attributes(kwargs)` and then make all the rule invocations inside the macros take `..., **propagate_common_rule_attributes(kwargs)` as their last parameter argument?

exactly, but I believe this should go into bazel-lib. https://github.com/aspect-build/bazel-lib/tree/main

reltuk:
Thanks for the feedback. I took a pass at something which may be helpful here: https://github.com/aspect-build/bazel-lib/pull/553.

The changes here end up looking something like:

```diff
@@ -21,7 +22,7 @@ oci_image_rule = _oci_image
 oci_image_index = _oci_image_index
 oci_push_rule = _oci_push
 
-def oci_image(name, labels = None, annotations = None, env = None, cmd = None, entrypoint = None, tags = [], testonly = False, **kwargs):
+def oci_image(name, labels = None, annotations = None, env = None, cmd = None, entrypoint = None, tags = [], **kwargs):
     """Macro wrapper around [oci_image_rule](#oci_image_rule).
 
     Allows labels and annotations to be provided as a dictionary, in addition to a text file.
@@ -45,6 +46,9 @@ def oci_image(name, labels = None, annotations = None, env = None, cmd = None, e
         tags: Tags to propagate to targets declared by this macro.
         **kwargs: other named arguments to [oci_image_rule](#oci_image_rule)
     """
+    forwarded_kwargs = propagate_common_rule_attributes(kwargs)
+    forwarded_kwargs.pop("tags", None)
+
     if types.is_dict(annotations):
         annotations_label = "_{}_write_annotations".format(name)
         write_file(
@@ -52,7 +56,7 @@ def oci_image(name, labels = None, annotations = None, env = None, cmd = None, e
             out = "_{}.annotations.txt".format(name),
             content = ["{}={}".format(key, value) for (key, value) in annotations.items()],
             tags = tags,
-            testonly = testonly,
+            **forwarded_kwargs,
         )
         annotations = annotations_label
 
```

alexeagle:
@reltuk we released bazel-lib yesterday including your change, so this should be unblocked now ;)

reltuk:
@alexeagle Thanks for the ping and the quick response! I took a pass at the dep bump + the change to the macros. Happy to split it up into separate PRs as well, if that's preferred.

oci/defs.bzl
```diff
@@ -45,13 +46,16 @@ def oci_image(name, labels = None, annotations = None, env = None, cmd = None, e
...
     """
+    forwarded_kwargs = propagate_common_rule_attributes(kwargs)
+
     if types.is_dict(annotations):
         annotations_label = "_{}_write_annotations".format(name)
         write_file(
             name = annotations_label,
             out = "_{}.annotations.txt".format(name),
             content = ["{}={}".format(key, value) for (key, value) in annotations.items()],
             tags = tags,
```
thesayyn, line 57: let's remove tags from arguments so that it's included in tags.

reltuk:
Took a pass at removing tags and also added forwarding common attributes for `oci_push`. I originally omitted it, since testonly didn't seem to be directly relevant, but there are use cases where it might be applicable.

thesayyn (approved):

reltuk:
`bazel run //docs:update` for me is a no-op on this branch. Is there something I should do to get this test passing?

Thanks for your help :).

thesayyn:
> `bazel run //docs:update` for me is a no-op on this branch. Is there something I should do to get this test passing?
> 
> Thanks for your help :).

I think you need to rebase and then run the command.

reltuk:
Unfortunately, none of the following produced any updates:

1) Rebasing against `github.com/bazel-contrib/rules_oci:main`

2) Running on Linux instead of macOS.

3) Running with 6.2.0 instead of 6.3.2.

I may get a chance to poke at it further in a bit :).

docs/image.md
```diff
@@ -118,7 +118,6 @@ This is similar to the same-named target created by rules_docker's `container_im
 | <a id="oci_image-env"></a>env |  Environment variables provisioned by default to the running container. See documentation above.   |  <code>None</code> |
 | <a id="oci_image-cmd"></a>cmd |  Command & argument configured by default in the running container. See documentation above.   |  <code>None</code> |
 | <a id="oci_image-entrypoint"></a>entrypoint |  Entrypoint configured by default in the running container. See documentation above.   |  <code>None</code> |
-| <a id="oci_image-tags"></a>tags |  Tags to propagate to targets declared by this macro.   |  <code>[]</code> |
-| <a id="oci_image-kwargs"></a>kwargs |  other named arguments to [oci_image_rule](#oci_image_rule)   |  none |
+| <a id="oci_image-kwargs"></a>kwargs |  other named arguments to [oci_image_rule](#oci_image_rule) and [common rule attribtes](https://bazel.build/reference/be/common-definitions#common-attributes).   |  none |
```
adzenith, line 121: looks like there's a typo in "attributes" (here and in four other locations in the PR)

PR #371, opened 2023-09-20
refactor: stop using jq for stamping

alexeagle:
bazel-lib 2.0 has a breaking change for this, and expand_template is simpler


docs/push.md
```diff
@@ -91,10 +91,22 @@ oci_image_index(
...
+    ],
+)
+
+# Use the value of --embed_label under --stamp, otherwise use a deterministic constant
+# value to ensure cache hits for actions that depend on this.
+expand_template(
     name = "stamped",
-    remote_tags = ["""($stamp.BUILD_EMBED_LABEL // "0.0.0")"""],
+    out = "_stamped.tags.txt",
+    template = "tags_tmpl",
```
alexeagle, line 107: note: I think we should change expand_template to a macro that accept a list of strings here, that would create the write_file target for you

alexeagle: https://github.com/aspect-build/bazel-lib/pull/533

Not sure if I want to land/release that first, then come back to this PR and force users onto newest bazel-lib. Probably... WDYT?

thesayyn: given that this is just docs, I'd say we shouldn't wait for it.

thesayyn (approved):

Issue #328, opened 2023-08-11
oci_tarball does not produce consistent tars for the same inputs


chancila:
tarballs generated with oci_tarball keep username and unstable timestamps making oci_tarballs not consistent

```
tar --list --file tarball.tar -v
-rw-r--r--  0 chancila staff    1182 Aug 10 18:27 manifest.json
drwxr-xr-x  0 chancila staff       0 Aug 10 18:27 blobs/
drwxr-xr-x  0 chancila staff       0 Aug 10 18:27 blobs/sha256/
-r-xr-xr-x  0 chancila staff  822203 Aug 10 18:27 blobs/sha256/8a79529e357e51ee5f147b1d870d38b60e48a920532fc2b8185da87f52bb5d2c.tar.gz
-r-xr-xr-x  0 chancila staff   21202 Aug 10 18:27 blobs/sha256/fe5ca62666f04366c8e7f605aa82997d71320183e99962fa76b3209fdfbb8b58.tar.gz
-r-xr-xr-x  0 chancila staff  716491 Aug 10 18:27 blobs/sha256/b02a7525f878e61fc1ef8a7405a2cc17f866e8de222c1c98fd6681aff6e509db.tar.gz
-r-xr-xr-x  0 chancila staff  103732 Aug 10 18:27 blobs/sha256/0b41f743fd4d78cb50ba86dd3b951b51458744109e1f5063a76bc5a792c3d8e7.tar.gz
-r-xr-xr-x  0 chancila staff     355 Aug 10 18:27 blobs/sha256/7c881f9ab25e0d86562a123b5fb56aebf8aa0ddd7d48ef602faf8d1e7cf43d8c.tar.gz
-r-xr-xr-x  0 chancila staff     385 Aug 10 18:27 blobs/sha256/4aa0ea1413d37a58615488592a0b827ea4b2e48fa5a77cf707d0e35f025e613f.tar.gz
-r-xr-xr-x  0 chancila staff     113 Aug 10 18:27 blobs/sha256/1e3d9b7d145208fa8fa3ee1c9612d0adaac7255f1bbc9ddea7e461e0b317805c.tar.gz
-r-xr-xr-x  0 chancila staff 18771270 Aug 10 18:27 blobs/sha256/b39c866052b17eaa1e92cee24f2bcf63f1d9bc8ec33fa438ec5c61e95b7b32e3.tar.gz
[...]

PR #380, opened 2023-09-29
oci_tarball: fix file time to 2000-01-01 for files inside the tar

gergelyfabian:
This way we can generate reproducible tars and avoid sha256sums changing for the tar output of the oci_tarball rule.

### DIFF
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -35,4 +35,4 @@ layers="${LAYERS}" \
         --output-format json > "${STAGING_DIR}/manifest.json"
 
 # TODO: https://github.com/bazel-contrib/rules_oci/issues/217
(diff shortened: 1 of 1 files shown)


gergelyfabian:
Potential fix for #328.
oci_tarball is now generating a reproducible tar.

thesayyn (approved):

thesayyn:
this is possibly one of the many non-hermetic input here.

chancila:
was this PR tested on macs? macos (seems like bsd in general) tar does not have --mtime in its man page

chancila:
verified this does not work on macos

alexeagle:
Yup, reverting.

gergelyfabian:
@alexeagle , thanks for the revert. Yes, indeed I diid not test this on mac (I have none).

PR #385, opened 2023-10-04
refactor: use hermetic tar

alexeagle:
Note, this shouldn't land until bazel-lib 2.0 is stable (RC or ideally final)

### DIFF
--- a/oci/private/tarball.bzl
+++ b/oci/private/tarball.bzl
@@ -76,12 +76,14 @@ def _tarball_impl(ctx):
 
     image = ctx.file.image
     tarball = ctx.actions.declare_file("{}/tarball.tar".format(ctx.label.name))
+    bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
     executable = ctx.actions.declare_file("{}/tarball.sh".format(ctx.label.name))
     repo_tags = ctx.file.repo_tags
 
     substitutions = {
         "{{format}}": ctx.attr.format,
         "{{jq_path}}": jq.bin.path,
+        "{{tar}}": bsdtar.tarinfo.binary.path,
         "{{image_dir}}": image.path,
         "{{tarball_path}}": tarball.path,
     }
@@ -96,9 +98,15 @@ def _tarball_impl(ctx):
         substitutions = substitutions,
     )
 
+    # TODO(2.0): this oci_tarball rule should just produce an mtree manifest instead,
+    # and then the tar rule can be composed in the oci_tarball macro in defs.bzl.
+    # To make it a non-breaking change, call the tar program from within this action instead.
     ctx.actions.run(
         executable = util.maybe_wrap_launcher_for_windows(ctx, executable),
-        inputs = [image, repo_tags, executable],
+        inputs = depset(
+            direct = [image, repo_tags, executable],
+            transitive = [bsdtar.default.files],
+        ),
         outputs = [tarball],
         tools = [jq.bin],
         mnemonic = "OCITarball",
@@ -131,6 +139,7 @@ oci_tarball = rule(
     toolchains = [
         "@bazel_tools//tools/sh:toolchain_type",
         "@aspect_bazel_lib//lib:jq_toolchain_type",
+        "@aspect_bazel_lib//lib:tar_toolchain_type",
     ],
     executable = True,
 )
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -112,4 +112,4 @@ repotags="${REPOTAGS[@]+"${REPOTAGS[@]}"}"
 add_to_tar "${manifest_json}" "manifest.json"
 
 # We've created the manifest, now hand it off to tar to create our final output
-"${TAR}" --create --file "${TARBALL_PATH}" "@${mtree}"
+"${TAR}" --create --no-xattr --no-mac-metadata --file "${TARBALL_PATH}" "@${mtree}"
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -2,10 +2,9 @@
 set -o pipefail -o errexit -o nounset
 
 readonly FORMAT="{{format}}"
-readonly STAGING_DIR=$(mktemp -d)
 readonly JQ="{{jq_path}}"
+readonly TAR="{{tar}}"
 readonly IMAGE_DIR="{{image_dir}}"
-readonly BLOBS_DIR="${STAGING_DIR}/blobs"
 readonly TARBALL_PATH="{{tarball_path}}"
 readonly REPOTAGS=($(cat "{{tags}}"))
 readonly INDEX_FILE="${IMAGE_DIR}/index.json"
@@ -84,6 +83,7 @@ MANIFEST_BLOB_PATH="${IMAGE_DIR}/blobs/${MANIFEST_DIGEST}"
 
 CONFIG_DIGEST=$(${JQ} -r '.config.digest  | sub(":"; "/")' ${MANIFEST_BLOB_PATH})
 CONFIG_BLOB_PATH="${IMAGE_DIR}/blobs/${CONFIG_DIGEST}"
+add_to_tar "${CONFIG_BLOB_PATH}" "blobs/${CONFIG_DIGEST}"
 
 LAYERS=$(${JQ} -cr '.layers | map(.digest | sub(":"; "/"))' ${MANIFEST_BLOB_PATH})
 
@@ -100,5 +100,7 @@ repotags="${REPOTAGS[@]+"${REPOTAGS[@]}"}"
         --arg config "blobs/${CONFIG_DIGEST}" \
         --argjson layers "${LAYERS}" > "${STAGING_DIR}/manifest.json"
 
-# TODO: https://github.com/bazel-contrib/rules_oci/issues/217
-tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" manifest.json blobs
+add_to_tar "${manifest_json}" "manifest.json"
+
+# We've created the manifest, now hand it off to tar to create our final output
+"${TAR}" --create --file "${TARBALL_PATH}" "@${mtree}"
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -9,11 +9,14 @@ readonly TARBALL_PATH="{{tarball_path}}"
 readonly REPOTAGS=($(cat "{{tags}}"))
 readonly INDEX_FILE="${IMAGE_DIR}/index.json"
 
-cp_f_with_mkdir() {
-  SRC="$1"
-  DST="$2"
-  mkdir -p "$(dirname "${DST}")"
-  cp -f "${SRC}" "${DST}"
+# Write tar manifest in mtree format
+# https://man.freebsd.org/cgi/man.cgi?mtree(8)
+# so that tar produces a deterministic output.
+mtree=$(mktemp)
+function add_to_tar() {
+    content=$1
+    tar_path=$2
+    echo >>"${mtree}" "${tar_path} uid=0 gid=0 mode=0755 time=1672560000 type=file content=${content}"
 }
 
 MANIFEST_DIGEST=$(${JQ} -r '.manifests[0].digest | sub(":"; "/")' "${INDEX_FILE}" | tr  -d '"')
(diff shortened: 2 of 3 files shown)


alexeagle:
hitting a bug in bazel-lib, sent https://github.com/aspect-build/bazel-lib/pull/574

oci/private/tarball.bzl
```diff
@@ -49,13 +49,16 @@ def _tarball_impl(ctx):
...
     tarball = ctx.actions.declare_file("{}/tarball.tar".format(ctx.label.name))
     yq_bin = ctx.toolchains["@aspect_bazel_lib//lib:yq_toolchain_type"].yqinfo.bin
+    bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
     executable = ctx.actions.declare_file("{}/tarball.sh".format(ctx.label.name))
+    mtree = ctx.actions.declare_file("_{}.mtree".format(ctx.label.name))
     repo_tags = ctx.file.repo_tags
 
     substitutions = {
         "{{yq}}": yq_bin.path,
+        "{{tar}}": bsdtar.tarinfo.binary.path,
```
gergelyfabian, line 59: Do you still need this substitution?

oci/private/tarball.bzl
```diff
@@ -71,12 +74,23 @@ def _tarball_impl(ctx):
     ctx.actions.run(
         executable = util.maybe_wrap_launcher_for_windows(ctx, executable),
         inputs = [image, repo_tags, executable],
-        outputs = [tarball],
+        outputs = [mtree],
         tools = [yq_bin],
         mnemonic = "OCITarball",
```
gergelyfabian, line 79: I would move this mnemonic and progress message to the new step (that calls `bsdtar.tarinfo.binary`) as that is the part that is effectively creating the Tarball. This one could be named differently.

oci/private/tarball.sh.tpl
```diff
@@ -3,9 +3,10 @@ set -o pipefail -o errexit -o nounset
 
 readonly STAGING_DIR=$(mktemp -d)
 readonly YQ="{{yq}}"
+readonly TAR="{{tar}}"
```
gergelyfabian, line 6: I think this is not any more needed.

oci/private/tarball.sh.tpl
```diff
@@ -34,5 +35,8 @@ layers="${LAYERS}" \
         --null-input '.[0] = {"Config": env(config), "RepoTags": "${repo_tags}" | envsubst | split("%") | map(select(. != "")) , "Layers": env(layers) | map( "blobs/" + . + ".tar.gz") }' \
         --output-format json > "${STAGING_DIR}/manifest.json"
 
-# TODO: https://github.com/bazel-contrib/rules_oci/issues/217
-tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" manifest.json blobs
+# Write output in mtree format
+# https://man.freebsd.org/cgi/man.cgi?mtree(8)
+# so that tar produces a deterministic output.
+echo "manifest.json uid=0 gid=0 mode=0755 type=file content=${STAGING_DIR}/manifest.json">>${OUT}
```
gergelyfabian, line 41: Do we want to set also the time for manifest.json and blobs?

gergelyfabian: It would be nice to set the time to something like 2000-01-01 (a second equivalent of that), to ensure that when unpacking the tar we won't get some warnings (for the too old time value).

oci/private/tarball.bzl
alexeagle: yeah, sorry this was a broken attempt (putting it in two actions doesn't work since the artifacts produced by the first weren't plumbed to the second)

PTAL

oci/private/tarball.bzl
alexeagle: I had to revert to a single action, PTAL

oci/private/tarball.sh.tpl
alexeagle: Good idea, output is now

```
$ bazel build examples/repo_tags:tarball
$ tar -tvf bazel-bin/examples/repo_tags/tarball/tarball.tar
-rwxr-xr-x 0/0             123 2023-01-01 00:00 blobs/sha256/dbe00f7d7b6cdecf0837ab201ce929b9cb08fd6960e1fd4e27b076ecbe3e6496
-rw-r--r-- 0/0             234 2023-01-01 00:00 manifest.json
```

oci/private/tarball.sh.tpl
```diff
@@ -3,24 +3,26 @@ set -o pipefail -o errexit -o nounset
...
+# https://man.freebsd.org/cgi/man.cgi?mtree(8)
+# so that tar produces a deterministic output.
+mtree=$(mktemp)
+
 MANIFEST_DIGEST=$(${YQ} eval '.manifests[0].digest | sub(":"; "/")' "${IMAGE_DIR}/index.json" | tr  -d '"')
 MANIFEST_BLOB_PATH="${IMAGE_DIR}/blobs/${MANIFEST_DIGEST}"
 
 CONFIG_DIGEST=$(${YQ} eval '.config.digest  | sub(":"; "/")' ${MANIFEST_BLOB_PATH})
 CONFIG_BLOB_PATH="${IMAGE_DIR}/blobs/${CONFIG_DIGEST}"
+echo >>"${mtree}" "blobs/${CONFIG_DIGEST} uid=0 gid=0 mode=0755 time=1672560000 type=file content=${CONFIG_BLOB_PATH}"
```
gergelyfabian, line 21: Maybe better move the time to a bash variable and use it in all 3 places where we want to set it? Now one is set to 0, the other two are at 1672560000...

alexeagle: was already doing that! :)

oci/private/tarball.sh.tpl
```diff
@@ -3,24 +3,32 @@ set -o pipefail -o errexit -o nounset
...
 
+# Write tar manifest in mtree format
+# https://man.freebsd.org/cgi/man.cgi?mtree(8)
+# so that tar produces a deterministic output.
+mtree=$(mktemp)
+function add_to_tar() {
+    tar_path=$1
+    content=$2
+    readonly mtree_flags="uid=0 gid=0 mode=0755 time=1672560000 type=file"
+    echo >>"${mtree}" "blobs/${LAYER}.tar.gz ${mtree_flags} content=${IMAGE_DIR}/blobs/${LAYER}"
```
gergelyfabian, line 19: You could use $tar_path and $content, I guess.

alexeagle: yup sorry I've been pushing before running tests locally,  don't normally have such an active volunteer reviewer 😅

oci/private/tarball.sh.tpl
```diff
@@ -34,5 +42,7 @@ layers="${LAYERS}" \
         --null-input '.[0] = {"Config": env(config), "RepoTags": "${repo_tags}" | envsubst | split("%") | map(select(. != "")) , "Layers": env(layers) | map( "blobs/" + . + ".tar.gz") }' \
         --output-format json > "${STAGING_DIR}/manifest.json"
 
-# TODO: https://github.com/bazel-contrib/rules_oci/issues/217
-tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" manifest.json blobs
+add_to_tar "${STAGING_DIR}/manifest.json" "manifest.json"
```
thesayyn, line 45: STAGING_DIR is not needed anymore. we can use a `mktemp` for the manifest directly.

gergelyfabian:
Thanks @alexeagle , I checked the latest version, works good for my usecase.

oci/private/tarball.sh.tpl
```diff
@@ -1,38 +1,48 @@
...
 readonly TAGS_FILE="{{tags}}"
 
+# Write tar manifest in mtree format
+# https://man.freebsd.org/cgi/man.cgi?mtree(8)
+# so that tar produces a deterministic output.
+mtree=$(mktemp)
+function add_to_tar() {
+    content=$1
+    tar_path=$2
+    local mtree_flags="uid=0 gid=0 mode=0755 time=1672560000 type=file"
```
thesayyn, line 17: one nit: we should create parent directories recursively, for once, for any file to match the behavior of `tar -cvf DIR OUTPUT`.

gergelyfabian:
Are there any automated tests that could be extended with checks for this new feature (e.g. gid, uid, time)?

oci/private/tarball.sh.tpl
alexeagle: okay, maybe that's a helper starlark function that a bazel-lib 2.x release could include, then we can reuse that logic? I assume we'll need to create parent directories there too.

thesayyn: we created this functionality there but only for automatically generated mtrees. we need to this in bash here however, which can't be shared with bazel-lib as we don't use bash there to generate the mtree

rickvanprim:
Looks like `aspect_bazel_lib` 2.0.0 released in Nov.  Can this work move forward now?

alexeagle:
> Can this work move forward now?

Forcing users to upgrade to bazel-lib 2.0 is a breaking change, so we're planning to start a 2.x branch here where this can land.

oci/private/tarball.sh.tpl
```diff
@@ -1,38 +1,47 @@
...
         --null-input '.[0] = {"Config": env(config), "RepoTags": "${repo_tags}" | envsubst | split("%") | map(select(. != "")) , "Layers": env(layers) | map( "blobs/" + . + ".tar.gz") }' \
-        --output-format json > "${STAGING_DIR}/manifest.json"
+        --output-format json > "${manifest_json}"
+
+add_to_tar "${manifest_json}" "manifest.json"
 
-# TODO: https://github.com/bazel-contrib/rules_oci/issues/217
-tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" manifest.json blobs
+# We've created the manifest, now hand it off to tar to create our final output
+"${TAR}" --create --file "${TARBALL_PATH}" "@${mtree}"
```
Strum355, line 47:
```suggestion
"${TAR}" --create --no-xattr --file "${TARBALL_PATH}" "@${mtree}"
```

Proposing to add this due to issues that are cropping up on macOS 
https://github.com/bazel-contrib/rules_oci/issues/509 and https://github.com/moby/moby/issues/47517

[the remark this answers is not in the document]
Strum355: The manpage for bsdtar seems to indicate that this flag is only available when extracting, although the rest of the comment for this flag could also be interpreted as also being possible when creating an archive, it starts off with "x mode only"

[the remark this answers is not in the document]
honnix: The manpage is indeed confusing, but from our test, having this option enabled got rid of `tar: Could not open extended attribute file: Operation not permitted` error.

alexeagle: thanks!

oci/private/tarball.sh.tpl
```diff
@@ -84,21 +91,25 @@ MANIFEST_BLOB_PATH="${IMAGE_DIR}/blobs/${MANIFEST_DIGEST}"
...
+  --arg repo_tags "$repotags" \
+  --arg config "blobs/${CONFIG_DIGEST}" \
+  --argjson layers "${LAYERS}"
+
+add_to_tar "${manifest_json}" "manifest.json"
 
-# TODO: https://github.com/bazel-contrib/rules_oci/issues/217
-tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" manifest.json blobs
+# We've created the manifest, now hand it off to tar to create our final output
+"${TAR}" --create --file "${TARBALL_PATH}" "@${mtree}"
```
thesayyn, line 115: Also add no-xattr and other flags here.

thesayyn (approved):

oci/private/tarball.sh.tpl
```diff
@@ -46,36 +48,41 @@ if [[ "${FORMAT}" == "oci" ]]; then
...
-  "${JQ}" -r --arg repo_tags "$repotags" \
-   '.manifests[0] as $manifest | .manifests = ($repo_tags | split(" ") | map($manifest * {annotations:{"org.opencontainers.image.ref.name":.}}))' "${INDEX_FILE}" > "${STAGING_DIR}/index.json"
+  index_json=$(mktemp)
+  "${JQ}" >"$index_json" \
+    -r --arg repo_tags "$repotags" \
+    '.manifests[0] as $manifest | .manifests = ($repo_tags | split(" ") | map($manifest * {annotations:{"org.opencontainers.image.ref.name":.}}))' "${INDEX_FILE}"
+  add_to_tar "$index_json" index.json
 
-  tar -C "${STAGING_DIR}" -cf "${TARBALL_PATH}" index.json blobs oci-layout
+  ${TAR} --create --no-xattr --no-mac-metadata --file "${TARBALL_PATH}" "@${mtree}"
```
rmohr, line 85: If I have a custom rule which allows adding capabilities (xattrs) to a binary to files, would this change remove these xattrs alter on when building the OCI image? I don't fully understand the context of this CL, so it may be unrelated, but we have something like this in https://github.com/rmohr/bazeldnf?tab=readme-ov-file#rpmtree and I wonder if your work would remove whatever we set there under `capabilities`.

manuelnaranjo: @rmohr without even looking at rules_oci code I can say it will not, we pass from bazeldnf tar files, that's our output, and that gets bundled unmodified into the image through the manifest that rules_oci provides, that's all rules_oci does tbh, if you give it tar files, it assumes those are layers and that the provider knows what they're doing. But please Alex and team correct me if I'm wrong

PR #389, opened 2023-10-08
Add flatten to oci_image

JohnAmican:
Context: https://bazelbuild.slack.com/archives/CA3NW13MH/p1696348792039019

### DIFF
--- a/oci/private/image.sh.tpl
+++ b/oci/private/image.sh.tpl
@@ -54,29 +54,31 @@ function empty_base() {
 function base_from_layout() {
     # TODO: https://github.com/google/go-containerregistry/issues/1514
     local refs=$(mktemp)
+    local flattened_refs=$(mktemp)
     local output=$(mktemp)
     local oci_layout_path=$1
     local registry=$2
 
     "${CRANE}" push "${oci_layout_path}" "${registry}/oci/layout:latest" --image-refs "${refs}" > "${output}" 2>&1
+    "${CRANE}" flatten $(cat $refs) > "${flattened_refs}"
 
     echo "${output}" >&2
 
     if grep -q "MANIFEST_INVALID" "${output}"; then
     cat >&2 << EOF
 
-zot registry does not support docker manifests. 
+zot registry does not support docker manifests.
 
 crane registry does support both oci and docker images, but is more memory hungry.
 
-If you want to use the crane registry, remove "zot_version" from "oci_register_toolchains". 
+If you want to use the crane registry, remove "zot_version" from "oci_register_toolchains".
 
 EOF
 
         exit 1
     fi
 
-    cat "${refs}"
+    cat "${flattened_refs}"
 }
 
 # this will redirect stderr(2) to stderr file.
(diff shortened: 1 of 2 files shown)


oci/private/image.sh.tpl
```diff
@@ -131,6 +135,11 @@ for ARG in "$@"; do
             FIXED_ARGS+=("--entrypoint=$in")
           done <"${ARG#--entrypoint-file=}"
           ;;
+        (--flatten=*)
```
alexeagle, line 138: I am torn about this.

One the one hand, this change doesn't add complexity to rules_oci it seems, just more faithfully presents the underling CLI of the tool we wrap.

OTOH I think this will result in users having massive outputs from actions, and then having to take special care that these aren't fetched back and forth from a remote cache to a local bazel server. I think the lack of a flattening ability can actually be considered a feature of rules_oci, that you can't recreate the bad behavior you get under rules_docker.

I think we need to find a way to solve the underlying problem (very slow load into docker daemon) either without needing flattening, or by putting some guardrails and warnings on this feature so that users understand the risks.

JohnAmican: > I am torn about this.

Yeah me too. I left it in draft because I don't have enough justification to put it up for review yet

> I think we need to find a way to solve the underlying problem (very slow load into docker daemon)

I profiled `docker buildx --builder=container build --load - < Dockerfile.localstack` on my local machine OSX/amd64

`cat Dockerfile.localstack`:
```
FROM localstack/localstack:2.2.0
```

`docker --debug buildx --builder=container inspect`:
```
Name:          container
Driver:        docker-container
Last Activity: 2023-10-09 01:09:41 +0000 UTC

Nodes:
Name:           container0
Endpoint:       desktop-linux
Driver Options: env.JAEGER_TRACE="localhost:6831" network="host"
Status:         running
Buildkit:       v0.12.2
Platforms:      linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/mips64le, linux/mips64, linux/arm/v7, linux/arm/v6
Features:
 Cache export:       true
 Docker exporter:    true
 Multiple platforms: true
 OCI exporter:       true
Labels:
 org.mobyproject.buildkit.worker.executor:         oci
 org.mobyproject.buildkit.worker.hostname:         docker-desktop
 org.mobyproject.buildkit.worker.network:          host
 org.mobyproject.buildkit.worker.oci.process-mode: sandbox
 org.mobyproject.buildkit.worker.selinux.enabled:  false
 org.mobyproject.buildkit.worker.snapshotter:      overlayfs
GC Policy rule#0:
 All:           false
 Filters:       type==source.local,type==exec.cachemount,type==source.git.checkout
 Keep Duration: 48h0m0s
 Keep Bytes:    488.3MiB
GC Policy rule#1:
 All:           false
 Keep Duration: 1440h0m0s
 Keep Bytes:    27.94GiB
GC Policy rule#2:
 All:        false
 Keep Bytes: 27.94GiB
GC Policy rule#3:
 All:        true
 Keep Bytes: 27.94GiB
```
Trace in Jaeger (GitHub won't let me attach the trace .json):
Screenshot 2023-10-09 at 9 53 45 AM

The only thing that sticks out is the number of logs that happen almost simultaneously:
Screenshot 2023-10-09 at 10 08 36 AM

These logs start at 5.37s and end at 5.42s. So it's like the actual streaming happened really fast, but whatever happens after that is slow and unmeasured?

Anyway, like I said I'm not advocating for merging this PR. At least not until I understand what the slow part actually is

thesayyn: My take on this PR is that this can be done in genrule, and generally there's little to no benefit for rules_oci to maintain this long term. A flattening rule like this will put a significant strain on the remote cache when there is more than one oci_image that flattens. 

It don't think we can accept this contribution.

thesayyn:
I am going to close this as not planned.

PR #390, opened 2023-10-08
chore: enable some CI on macos again

alexeagle:
We need at least some coverage, even if we don't have a container runtime there


thesayyn (approved):

thesayyn:
@alexeagle PTAL

alexeagle:
Nice! LGTM and thanks for figuring it out.
I'm the author so you have to be the approver :)

thesayyn:
> I'm the author so you have to be the approver :)

LGTM

PR #393, opened 2023-10-16
refactor: remove workdir workaround

thesayyn:


MODULE.bazel
```diff
@@ -11,7 +11,7 @@ bazel_dep(name = "bazel_skylib", version = "1.4.1")
 bazel_dep(name = "platforms", version = "0.0.5")
 
 oci = use_extension("//oci:extensions.bzl", "oci")
-oci.toolchains(crane_version = "v0.14.0")
+oci.toolchains(crane_version = "v0.15.1")
```
alexeagle, line 14: are users forced to upgrade as well? (maybe we're forcing them here, if our version takes precedence over one that they register?

oci/private/image.sh.tpl
```diff
@@ -79,13 +79,16 @@ EOF
     cat "${refs}"
 }
 
+function history() {
```
alexeagle, line 82: this seems to be unused

MODULE.bazel
thesayyn: yes, this is a breaking change if they register one of their own toolchain.

thesayyn: We have already updated crane in another PR. While technically this is a breaking change, i don't expect any side effects from upgrading crane.

alexeagle (approved):

PR #404, opened 2023-10-25
feat: add exposed_ports to oci_image

### LINKED ISSUE #220: oci_image should allow setting ports
See: https://github.com/google/go-containerregistry/pull/1677

https://bazelbuild.slack.com/archives/CA3NW13MH/p1682514177313279

gzm0:
Fixes #220.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -13,8 +13,8 @@ load("@rules_oci//oci:defs.bzl", ...)
 ## oci_image_rule
 
 <pre>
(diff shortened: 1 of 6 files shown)


gzm0:
Seems this fails with bazelmod. I'm investigating.

gzm0:
Seems like under bzlmod, we're not using `LATEST_CRANE_VERSION`:

https://github.com/bazel-contrib/rules_oci/blob/1fce184ebb51fbd7a51268007473e33b9fd3d75f/MODULE.bazel#L14

Is this deliberate?

gzm0:
I have bumped the crane version in the MODULE.bazel of the e2e test. The test passes locally. Let's see what CI says.

If you are concerned with the test not testing with the lowest supported crane version anymore, I'm happy to split it out.

This is a separate discussion, but I find the choice of toolchain version management under bzlmod odd: I'd expect that by default, I'd get the latest known versions (but the ability to *downgrade* the versions if necessary).

trym-b:
Thank you for adding this feature.

Is there anything blocking this PR?
I would like to get my hands on this feature as I have some container images that I need to convert to `bazel` that require exposing some ports (more specifically [this](https://github.com/nlnwa/veidemann-dns-resolver/blob/master/Dockerfile#L18)).

thesayyn:
We'll unblock this shortly this week. We need to upgrade the crane version which is an undocumented public API contract we never intended to expose. We'll have to document it and tell users to register their own toolchains if they rely on a specific version of crane.

thesayyn (approved):

salrashid123:
I don't think this in included in the latest  `1.4.3` release  but its cited in the docs as supported already.

When is the next release going out?

PR #405, opened 2023-10-25
fix: remove workaround for missing crane workdir support

gzm0:
Support got added
https://github.com/google/go-containerregistry/pull/1615

Discovered while working on #220.


gzm0:
I have not vetted this for backwards compatibility concerns w.r.t. old crane versions. What is the general policy for this?

thesayyn:
i have a PR for this already, https://github.com/bazel-contrib/rules_oci/pull/393. shall we close this in favor of that one?

gzm0:
Ah, yes. Of course.

PR #407, opened 2023-10-25
Draft: Implemented stamping created time

### LINKED ISSUE #49: oci_image should support stamping and creation time.
currently, the creation time of the images is set to a fixed date for reproducibility.  this should be overridable with stamping.

liningpan:
Implemented stamping container created time with `BUILD_TIMESTAMP`

Closes #49

### DIFF
--- a/examples/created_timestamp/BUILD
+++ b/examples/created_timestamp/BUILD
@@ -1,6 +1,4 @@
-load("@aspect_bazel_lib//lib:expand_template.bzl", "expand_template")
 load("@aspect_bazel_lib//lib:jq.bzl", "jq")
-load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 load("//oci:defs.bzl", "oci_image")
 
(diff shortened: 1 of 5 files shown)


thesayyn:
this goes against our pattern of provide us a file, which may or may not be stamped and we will used that file instead of teaching every attribute stamping.

liningpan:
So, does it mean you would prefer to have the created time provided in a file?

thesayyn:
yes that's the approach we've taken.

liningpan:
Okay, we can set the created timestamp from a file now. I'm not sure what's the best way to add tests. Finding the file containing the timestamp requires one level of indirection (i.e. need to read the manifest file to find the blob).

Also, because timestamp is a single string, I don't know if there is a way to distinguish it from a label, which makes the semantics of this parameter different from others.

examples/created_timestamp/BUILD
```diff
@@ -0,0 +1,42 @@
...
+        "BUILD_TIMESTAMP",
+    ],
+)
+
+expand_template(
+    name = "build_timestamp",
+    out = "build_timestamp.txt",
+    stamp_substitutions = {"BUILD_TIMESTAMP": "{{BUILD_TIMESTAMP}}"},
+    substitutions = {"BUILD_TIMESTAMP": "946684800"},
+    template = ":build_timestamp_tmpl",
```
alexeagle, line 25: Nit: newer version of this rule accepts a list of strings here which become the lines of the template, avoiding need for the write_file rule

alexeagle (commented):
Feels like the right shape, maybe we would be okay with making the oci_image macro do more of this so there's less BUILD file for users to write

thesayyn:
@liningpan do you mind rebasing this PR against 2.x, and rename the attribute to `created` to follow the spec?

thesayyn:
Closing this as inactive. Feel free to submit a new one

PR #409, opened 2023-11-03
Hermetic Core Utilities

mattyclarkson:
This expands it usage in the shell script for other core utilities in more of the shell scripts.

The usage of `xargs` could be rewritten in pure Bash, maybe.

Hoping to get `rules_oci` to a point where it require zero tooling from the host 🤞

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -107,6 +107,7 @@ def _oci_image_impl(ctx):
     crane = ctx.toolchains["@rules_oci//oci:crane_toolchain_type"]
     registry = ctx.toolchains["@rules_oci//oci:registry_toolchain_type"]
     jq = ctx.toolchains["@aspect_bazel_lib//lib:jq_toolchain_type"]
+    coreutils = ctx.toolchains["@aspect_bazel_lib//lib:coreutils_toolchain_type"]
 
     launcher = ctx.actions.declare_file("image_%s.sh" % ctx.label.name)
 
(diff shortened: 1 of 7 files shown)


mattyclarkson:
Related to #385 in terms of hermetic-ness of the project.

thesayyn:
Hey, thank you for the PR. I would love to land this, but currently there's a problem with the coreutils toolchain on darwin/arm64 where if rosetta is disabled, it fails horribly.

thesayyn:
cross-ref: https://github.com/uutils/coreutils/issues/4104

thesayyn:
This is now unblocked, we will accept the PR once we upgrade bazel-lib here.

alexeagle:
Along with Bazel 7 release we'll drop Bazel 5 support, that will be our chance to make a 2.0 release here, which in turn lets us make the breaking change of requiring bazel-lib 2.0. Sorry this got tied up in our semver ...

mattyclarkson:
No worries. We use `xargs` [here](https://github.com/bazel-contrib/rules_oci/blob/375e5dbf3b2c50e83c9df0d0dc16a027984dab77/oci/private/push.sh.tpl#L56)

```sh
if [[ -e "${TAGS_FILE:-}" ]]; then
  cat "${TAGS_FILE}" | xargs -n1 "${CRANE}" tag $(cat "${REFS}")
fi
```

The docs say:

> a .txt file containing tags, one per line.

```sh
if [[ -e "${TAGS_FILE:-}" ]]; then
  for IFS= read -r TAG; do
    "${CRANE}" tag $(cat "${REFS}")
  done <"${TAGS_FILE}"
fi
```

Which would get rid of another external program. 

I can put up a separate PR or add it to this one, if that is something we'd be happy to take.

thesayyn:
@mattyclarkson do you mind rebasing this PR against 2.x, and we'll land it there?

thesayyn:
Okay closing this as it's not landable as is. Feel free to create a new PR rebased on top 2.x.

Issue #410, opened 2023-11-03
Selecting zot does not work with bzlmod


siddharthab:
When I try to configure a zot toolchain like so:
```
oci.toolchains(
    crane_version = "v0.15.2",
    zot_version = "v1.4.3-rc3",
)
```
I get:
```
build aborted: module extension "oci" from "@rules_oci~1.4.0//oci:extensions.bzl" does not generate repository "oci_crane_registry_toolchains", yet it is imported as "oci_crane_registry_toolchains" in the usage at https://bcr.bazel.build/modules/rules_oci/1.4.0/MODULE.bazel:13:20
```

When I add in a name to not conflict with the default crane registry toolchain:
```
oci.toolchains(
    name = "myrepo",
    crane_version = "v0.15.2",
    zot_version = "v1.4.3-rc3",
)
```
I get:
```
Error in repository_rule: A repo named oci_auth_config is already generated by this module extension at .../external/rules_oci~1.4.0/oci/repositories.bzl:114:28
```

alexeagle:
Sounds correct, do you have any time to make a PR for this?

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/550

PR #411, opened 2023-11-04
Update crane and zot versions in MODULE.bazel

siddharthab:
This makes it consistent with WORKSPACE, and also enables the use of zot
for bzlmod users.

Related issue: #410.


siddharthab:
Actually, I take this back, crane has to be the default option. MODULE.bazel does not have to be consistent with WORKSPACE, because WORKSPACE does not apply to end users.

PR #421, opened 2023-11-20
crane / zot launcher: kill registry serve process on shell exit

### LINKED ISSUE #419: `crane registry serve` daemons are not cleaned up
When executing build actions locally, `crane registry serve` processes are cleaned up after a `bazel build`.
However, when using remote execution (in my case with buildbarn), `crane registry serve` processes survive between actions and pile up over time on the remote executor. This results in hundreds of processes and will starve the host of memory over multiple days.

## How to reproduce

- [setup a buildbarn cluster](https://github.com/buildbarn/bb-deployments) (for example using the docker-compose setup)
- Use the buildbarn cluster for remote execution (`--remote_executor=...`)
- use rules_oci to build container images
- Look at the running processes after the build is complete

### Expected behavior

`crane registry serve` should no longer run

### Actual behavior

`crane registry serve` processes are still running

malt3:
Fixes #419

### DIFF
--- a/oci/private/image.sh.tpl
+++ b/oci/private/image.sh.tpl
@@ -13,6 +13,7 @@ readonly STORAGE_DIR="{{storage_dir}}"
 readonly STDERR=$(mktemp)
 
 silent_on_success() {
+    stop_registry ${STORAGE_DIR}
     CODE=$?
     if [ "${CODE}" -ne 0 ]; then
         cat "${STDERR}" >&1
--- a/oci/private/registry/crane_launcher.sh.tpl
+++ b/oci/private/registry/crane_launcher.sh.tpl
@@ -5,12 +5,14 @@ function start_registry() {
     local storage_dir="$1"
     local output="$2"
     local deadline="${3:-5}"
+    local registry_pid="$1/proc.pid"
 
     mkdir -p "${storage_dir}"
     # --blobs-to-disk uses go's os.TempDir() function which is equal to TMPDIR under *nix.
     # https://pkg.go.dev/os#TempDir
     TMPDIR="${storage_dir}" TMP="${storage_dir}" \
     "${CRANE_REGISTRY_BIN}" registry serve --blobs-to-disk >> $output 2>&1 &
+    echo "$!" > "${registry_pid}"
 
     local timeout=$((SECONDS+${deadline}))
 
@@ -25,5 +27,17 @@ function start_registry() {
         return 1
     fi
     echo "127.0.0.1:${port}"
+    return 0
+}
+
+function stop_registry() {
+    local storage_dir="$1"
+    local registry_pid="$1/proc.pid"
+    if [[ ! -f "${registry_pid}" ]]; then
+        echo "Registry not started" >&2
+        return 0
+    fi
+    echo "Stopping registry process" >&2
+    kill -9 "$(cat "${registry_pid}")" || true
     return 0
 }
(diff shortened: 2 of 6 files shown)


thesayyn (commented):
I believe the cleanup procedure doesn't need to leak into call site rules. the launcher script can easily keep a global state and expose a cleanup function that kills the process.

malt3:
> [...] the launcher script can easily keep a global state and expose a cleanup function that kills the process
This is not correct. The `start_registry` function is invoked in a subshell like this:

```bash
readonly REGISTRY=$(start_registry "${STORAGE_DIR}" "${STDERR}" "${REGISTRY_PIDFILE}")
```

The subshell (and any global state that may be preserved inside) does not survive the subshell invocation.

Or to be more precise:

```bash
function start_registry() {
    # [...]
    "${REGISTRY_BIN}" >> $output 2>&1 &
    readonly REGISTRY_PID=$!
    # [...]
     return 0
}

function stop_registry() {
    if [[ -z "${REGISTRY_PID+x}" ]]; then
        echo "Registry not started"
        return 0
    fi
    echo "Stopping registry process ${REGISTRY_PID}"
    kill -9 "${REGISTRY_PID}" || true
    return 0
}

# if I do this, the variable will survive but we cannot easily capture the registry address without writing to a file:
start_registry # [...] args
stop_registry

# if I do this, the variable is only part of the subshell
readonly REGISTRY=$(start_registry # [...] args)
stop_registry
```

thesayyn:
Ah true, we can write the pid into storage_dir/proc.pid or call `pkill -P $$` kill all subprocesses. I'd rather keep the diff minimal.

malt3:
I think pkill should not be used since it is not part of POSIX and not even in coreutils or util-linux.
I used the fixed path inside the storage dir, as suggested.

oci/private/image.sh.tpl
```diff
@@ -12,6 +12,8 @@ readonly STORAGE_DIR="{{storage_dir}}"
 
 readonly STDERR=$(mktemp)
 
+trap "stop_registry ${STORAGE_DIR}" EXIT
```
thesayyn, line 15: this overrides the existing trap on EXIT called `silent_on_success` can you just call the stop registry function there?

examples/attest/test.bash
```diff
@@ -15,21 +15,22 @@ readonly SBOM_PATH="$7"
 # Launch a registry instance at a random port
 source "${REGISTRY_LAUNCHER}"
 REGISTRY=$(start_registry $TEST_TMPDIR $TEST_TMPDIR/output.log)
+trap "stop_registry ${TEST_TMPDIR}" EXIT
```
thesayyn, line 18: these don't need to change as they are tests. we can keep them as is as we don't have the ghost problem in our CI.

thesayyn (commented):
diff is still big and there seems to be some formatting changes. Can you revert those?

oci/private/registry/crane_launcher.sh.tpl
```diff
@@ -26,4 +28,16 @@ function start_registry() {
...
+}
+
+function stop_registry() {
+    local storage_dir="$1"
+    local registry_pid="$1/proc.pid"
+    if [[ ! -f "${registry_pid}" ]]; then
+        echo "Registry not started"
+        return 0
+    fi
+    echo "Stopping registry process"
```
thesayyn, line 40: unnecessary noise? did you want to print this to `>&2`?

thesayyn:
Sorry @malt3 going to look into this soon.

thesayyn (approved):

PR #426, opened 2023-11-29
feat: bump and assert crane version

thesayyn:
bumps crane version to `0.17.0` and introduces a check to prevent confusing messages


oci/private/image.bzl
```diff
@@ -85,11 +85,6 @@ If `group/gid` is not specified, the default group and supplementary groups of t
     "_empty_tar": attr.label(default = "empty.tar", allow_single_file = True),
 }
 
-def _format_string_to_string_tuple(kv):
```
thesayyn, line 88: this is unused. it's removed in a couple other PRs which are blocked. i want to remove this now.

alexeagle (approved):

PR #429, opened 2023-12-05
Provide an example in the compatibility section

brianduff:
It took me quite a bit of time to figure out how to fix compatibility issues with binaries compiled with newer glibc. Including a specific example of oci_pull with a newer debian cc image should help future wanderers.


alexeagle:
Hi @brianduff !!

Issue #439, opened 2023-12-15
oci_image uses significantly more disk space than container_image from rules_docker


SanjayVas:
`oci_image` appears to copy all base image layer blobs to its Bazel output tree. Additionally it temporarily takes up even more space due to running a separate registry for each target.

I believe `container_image` from `rules_docker` only has the created layer(s) in its output.

Anecdotally, this blew up my CI. My compressed Bazel output cache went from ~3 GiB to ~20 GiB when migrating from rules_docker to rules_oci.

alexeagle:
I think this is fixed by that #440 - @SanjayVas if you have time, could you patch that in and verify?

thesayyn:
@alexeagle partially. we still copy blobs from base image (copy_to_directory) into oci_image's oci-layout (treeartifact) which causes duplication of the blobs. 
we should symlink blobs from base image into the oci_image's output to prevent copies.

thesayyn:
but that's a bigger fix and i don't have time to contribute. happy to review if somebody else is willing to contribute it.

SanjayVas:
> I think this is fixed by that #440 - @SanjayVas if you have time, could you patch that in and verify?

Sorry, the migration to rules_oci was prefactoring for migrating our tree of repos to Bzlmod. I can possibly try after that is done, but it will of course be easier if a version with that change has been published to BCR.

thesayyn:
Okay now that https://github.com/bazel-contrib/rules_oci/pull/560 https://github.com/bazel-contrib/rules_oci/pull/559 landed, this can be closed.

PR #440, opened 2023-12-15
perf: use hardlink

thesayyn:
This avoid having two copies of blobs in the external repository.  

Part of #439

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -330,6 +330,8 @@ load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 copy_to_directory(
     name = "{target_name}",
     out = "layout",
+    # Always use hardlink to avoid having copies of the blobs from external repository in the output-tree.
+    hardlink = "on",
     include_external_repositories = ["*"],
     srcs = [
         "blobs",


alexeagle (approved):

honnix:
Sorry coming back to this very old change. I think we found an issue with this approach.

Also sorry for pasting agent analysis, but it does describe it very clearly than myself. I understand the reason for this change, but still it comes with a big surprise, especially when using bazel-diff/target-determinator to compute hashes as execution bit is part of hash input.

  The `oci_pull` repository rule generates a `copy_to_directory` target with `hardlink = "on"`:

  ```starlark
  copy_to_directory(
      name = "noble_linux_amd64",
      hardlink = "on",
      include_external_repositories = ["*"],
      srcs = glob(["blobs/**"]) + ["oci-layout", "index.json"],
      ...
  )
```

  When this target is built locally (not via remote execution), copy_to_directory creates hardlinks from the source blob files in the external repo directory into the output tree.
  Bazel then chmod's the output files to 0555 after action completion. Since hardlinks share the same inode, this also changes the original source files from their initial mode
  (e.g., 0644 from rctx.download()) to 0555.

  This is a known side effect documented in aspect-build/bazel-lib:

> Creating hardlinks is much faster than making copies of files with the caveat that hardlinks share file permissions with their source. Since Bazel removes write permissions on  ▎ files in the output tree after an action completes, hardlinks to source files are not recommended since write permissions will be inadvertently removed from source files.

  The hardlink = "on" option is documented as "not recommended" for this reason.

  Impact

  Any tooling that hashes source file permissions (e.g., target-determinator for incremental build detection) produces non-deterministic results depending on whether the
  copy_to_directory target has been built on the current machine. Machines that have built it see 0555, machines that haven't see 0644, causing thousands of false-positive impacted  targets.

  Suggestion

  Change the generated BUILD.bazel to use hardlink = "auto" (which only hardlinks tree artifacts already in the output tree) or hardlink = "off" instead of hardlink = "on".

PR #452, opened 2023-12-27
fix: missing error output on shell exit

timonwong:
If oci_image script failes, there is no error output anymore since #421, this simple fix tries to address that.

Since `stop_registry` is fairly simple enough, so I think we can ignore the exit code about it.


thesayyn (approved):

thesayyn:
This is already fixed on main.

PR #441, opened 2023-12-15
support passing label as container cli tool to bazel run //my/tarball

jchorl:
Enables passing a label to `oci_tarball` which will be used to load the image into the local container engine.

Solves https://github.com/bazel-contrib/rules_oci/issues/295

Follows guidance from https://github.com/bazel-contrib/rules_oci/issues/430

I think it would be nice if `oci_register_toolchains` configured this, but I understand not wanting to clutter the configs.

To test:

WORKSPACE:
```
local_repository(
    name = "rules_oci",
    path = "/local/rules_oci",
)

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

load("@rules_oci//oci:dependencies.bzl", "rules_oci_dependencies")

rules_oci_dependencies()

load("@rules_oci//oci:repositories.bzl", "LATEST_CRANE_VERSION", "oci_register_toolchains")

oci_register_toolchains(
    name = "oci",
    crane_version = LATEST_CRANE_VERSION,
)

load("@rules_oci//oci:pull.bzl", "oci_pull")

oci_pull(
    name = "distroless_base",
    digest = "sha256:ccaef5ee2f1850270d453fdf700a5392534f8d1a8ca2acda391fbb6a06b81c86",
    image = "gcr.io/distroless/base",
    platforms = [
        "linux/amd64",
        "linux/arm64",
    ],
)

# from https://download.docker.com/linux/static/stable/x86_64/
http_archive(
    name = "docker_cli",
    url = "https://download.docker.com/linux/static/stable/x86_64/docker-24.0.7.tgz",
    sha256 = "984d59a77fa6acab97ea725a966facd33725e0e09c2fee975397fe1c6379bd3d",
    build_file_content = """
exports_files(["docker/docker"])
    """
)
```

BUILD.bazel:
```
load("@rules_oci//oci:defs.bzl", "oci_image", "oci_tarball")

oci_image(
    name = "image",
    base = "@distroless_base",
    cmd = ["echo", "hi"],
)

oci_tarball(
    name = "tarball",
    image = ":image",
    repo_tags = ["jchorl/test:latest"],
    container_cli_tool = "@docker_cli//:docker/docker",
)
```

```
ubuntu@99ab9feb506a:/src/workspace$ bazel run //:tarball
INFO: Analyzed target //:tarball (1 packages loaded, 4 targets configured).
INFO: Found 1 target...
Target //:tarball up-to-date:
  bazel-bin/tarball/tarball.tar
[...]

PR #453, opened 2024-01-03
refactor: rename new attribute on oci_tarball

alexeagle:
Followup to #441. I think 'command' is too vague, as it might seem to apply to the creation of the tarball under 'bazel build', and also doesn't say anything about what purpose the command has. Also improve the docs around this pair of attributes.

Note: not a breaking change as the 'command' attribute wasn't released yet.


thesayyn (approved):

jchorl:
Thank you for grabbing!

Issue #454, opened 2024-01-03
oci_tarball loading is not incremental


alexeagle:
rules_docker has a fancy program https://github.com/bazelbuild/rules_docker/blob/master/container/incremental_load.sh.tpl#L4

thesayyn:
my 2 cents on this issue; 

This is not really a bazel-shaped problem that i am keen on solving anytime soon. oci_tarball never meant to be a node deep in the dependency graph but a leaf.

a quick look at this issue will show us how messy this is and lead to more issues being reported https://github.com/google/go-containerregistry/issues/895

thesayyn:
one way I solved this in the past was by using a neat trick to start a registry daemon on the host and have docker runtime pull images from there.

juanzolotoochin:
Here's my prototype implementation of a loader that works incrementally: https://github.com/juanique/bazel-examples/blob/load_image_prototype_7ff37/bazel6/rules_oci/2-go-image/app/BUILD.bazel#L15 if anyone wants to use it.

@thesayyn I understand that `oci_tarball` may not be meant for this, but the use case of "I need to load this docker image on multiple remote executors for integration tests" is a valid one that worked very well with `rules_docker` and it's just not well supported on `rules_oci`. So at least there should be some guidance on *what* to do for that use case.

thesayyn:
I understand the need for such a thing. It's a tradeoff between maintenance and performance for us. Using tarballs for this purpose is not the right solution.

Some things can be done to improve DX but maintaining a loader monstrosity is not one of them. 

- Bazel has performance issues around TreeArtifacts: https://github.com/bazelbuild/bazel/issues/17804, https://github.com/bazelbuild/bazel/issues/16913
- Docker tarballs are not meant to be transferred on wire, docker would have used it for distribution to registries if it was.
- Even docker is not willing to address this issue on their end; https://github.com/docker/buildx/issues/107

I am trying to find the right solution rather than a workaround.

FWIW rules_docker's tarball loader shifts the responsibility away from bazel action to runtime where blobs are closest to the executor so it is mostly fast.

alexeagle:
BTW to be clear, if anyone was willing to step up and maintain rules_docker, it could be unarchived. Our goal was never to have feature parity with it.

gzm0:
[...]

PR #455, opened 2024-01-03
chore: example incremental loader

thesayyn:
This is way faster and more incremental than anything out there.


thesayyn:
https://github.com/bazel-contrib/rules_oci/issues/454

oci/private/tarball.bzl
```diff
@@ -51,10 +51,9 @@ attrs = {
```
alexeagle, line 51: update the doc on the `loader` attribute

oci/private/tarball.bzl
```diff
@@ -51,10 +51,9 @@ attrs = {
 
             See the _run_template attribute for the script that calls this loader tool.
             """,
-        allow_single_file = True,
         mandatory = False,
```
alexeagle, line 54: allow_single_file might have been load-bearing if someone just pointed this at an executable .sh file?

alexeagle (approved):

oci/private/tarball.bzl
thesayyn: hmm, i think i am gonna revert this change.

thesayyn: can be done later if needed.

thesayyn: reverted.

PR #459, opened 2024-01-12
Updated `image.sh.tpl` and `tarball.sh.tpl` to use hermetic tools.

rickvanprim:
Depends on https://github.com/aspect-build/bazel-lib/pull/721.

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -103,8 +103,10 @@ def _oci_image_impl(ctx):
 
     crane = ctx.toolchains["@rules_oci//oci:crane_toolchain_type"]
     registry = ctx.toolchains["@rules_oci//oci:registry_toolchain_type"]
+    coreutils = ctx.toolchains["@aspect_bazel_lib//lib:coreutils_toolchain_type"]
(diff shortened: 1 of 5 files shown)


alexeagle (commented):
Thanks, this overlaps with https://github.com/bazel-contrib/rules_oci/pull/385 which I haven't landed because it's a breaking change. It looks like this one isn't though - it works with current versions of dependencies.

alexeagle:
Oh I see, it's certainly breaking as it assumes some bazel-lib changes that haven't landed yet.

rickvanprim:
Yes this is dependent on the other PR.  I updated this PR description to reflect that.  What's the policy around updating the version of `aspect_bazel_lib` that `rules_oci` depends on?

alexeagle:
That's where #385 got blocked - unless we backport things to bazel-lib 1.x, we are treating it as a breaking change for rules_oci to require bazel-lib 2.x.

rickvanprim:
I haven't tried, but presumably I could open another PR in `aspect_bazel_lib` against the `1.x` branch, as the changes to add those toolchains are all pretty self contained.

thesayyn:
Now, both oci_image and oci_tarball is using hermetic tools. Closing this.

PR #460, opened 2024-01-12
feat: look for credential helpers in common paths 

sitaktif:
The docker credential helpers might be present in a few common places besides the default `/bin:/usr/bin:/usr/sbin` one. Look for it in a few directories commonly used by package managers.

Looking for a binary if different places is not ideal in terms of reproductibility but it makes practical sense given the nature of the executable we're looking for (a credential helper specified in the user's configuration).

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -276,7 +276,7 @@ Falling back to using `curl`. See https://github.com/bazelbuild/bazel/issues/178
             fallback_to_curl = True
     else:
         util.warning(rctx, """\
-Could not fetch the manifest. Either there was an authentication issue or trying to pull an image with OCI image media types. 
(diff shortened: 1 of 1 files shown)


thesayyn:
why wouldn't these paths be in PATH by default? I don't know any tool that does this kind of lookup.

sitaktif:
Thanks, this... is a great question.

I stumbled upon a situation where the `PATH` was set to a minimal value such as `/bin:/usr/bin:/usr/sbin` in this very script (echoing the env var and exiting non-zero to check) which was the cause of a failure to pull an image with `oci_pull`, but I can't reproduce it anymore and haven't seen any bazel options or execution context that would explain that behaviour then.

I will close the PR for now and will reopen it with more information if I feel such a change still makes sense.

PR #474, opened 2024-01-25
Always allow specifying the "platforms" attribute in `oci_pull`

sitaktif:
### Changes

This change now allows users to specify a one-element "platforms" attribute in `oci_pull` for single-architecture images.
When doing so, the `oci_pull` repo rule will behave the same as with a multi-architecture image, namely:
- The user-provided platform is checked against the image metadata and errors out in case of mismatch.
- A `mytarget_<os>_<arch>` target will be created instead of `mytarget_single_arch`
- The `mytarget` alias created by the repo rule now uses the same kind of `select()` as multi-arch images, mapping the Bazel target platform's cpu to the platform-specific image. If the Bazel target platform cpu doesn't match the (single) image's architecture, Bazel will error out. In contrast, if the user doesn't specify the "platforms" attribute, the image can be selected regardless of the Bazel target platform's cpu.

### Notes

A few open questions in comments below.

### Background / justification

We found that always specifying the platforms attribute has some nice properties:
- It makes all the oci_pull rules (multi-arch and single-arch) consistent from the user's perspective.
- It ensures that the available image platforms are always explicit just by glancing at the repo rule (no more need for users to pull the image manually to figure out its platform).
- It adds validation by checking the image config's os/architecture against the user-specified platforms attribute.
- Last but not least, because of the way oci_image and oci_pull interact, it allows adding the same platform compatibility check as for multi-arch images, ensuring that an oci_image that is getting built for e.g. an arm64 architecture doesn't blindly take a base image that may be of amd64 architecture. The user gets an explicit error message instead.
    - In terms of implementation, single-arch images with a platforms attribute would have the same kind of oci_alias as multi-arch ones, with a select(...) for a single platform, and if that platform doesn't match, the user gets a useful error message instead of silently building an image with a wrong platform.


oci/private/pull.bzl
```diff
@@ -361,28 +366,86 @@ def _oci_pull_impl(rctx):
...
-    # symlink manifest.json to blobs with it's digest. 
+
+    # symlink manifest.json to blobs with its digest.
     # it is okay to use symlink here as copy_to_directory will dereference it when creating the TreeArtifact.
     rctx.symlink("manifest.json", _digest_into_blob_path(digest))
 
-    # download the image config
-    downloader.download_blob(manifest["config"]["digest"], _digest_into_blob_path(manifest["config"]["digest"]))
+    # if the user provided a platform for the image, validate it matches the config as best effort.
+    if rctx.attr.platform:
```
sitaktif, line 395: Note that this change also affects multi-arch images: we now validate the platform in the images' config metadata (not just in the image index). I don't think this should be considered backward-incompatible because:
- If the os/arch mis-match, there's something very dodgy happening with the user's image index manifest.
- If the os/arch are unset, we don't fail.

oci/private/pull.bzl
```diff
@@ -361,28 +366,86 @@ def _oci_pull_impl(rctx):
...
+                attr_architecture,
+                image_architecture,
+            ))
+
+        if missing_fields:
+            util.warning(rctx, "Could not confirm pulled image {}/{} is for platform {}: {} missing from image metadata".format(
+                rctx.attr.registry,
+                rctx.attr.repository,
+                rctx.attr.platform,
+            ))
```
sitaktif, line 436: Is this warning too intrusive?

The OCI spec marks os/architecture as mandatory but the docker spec doesn't. I'm pretty sure that recent versions of docker do set the os/architecture attributes but some users might be stuck with old images they don't have control over...

On the other hand, it is useful to notify the user that we couldn't confirm the metadata matched so they know to take action to fix their image when they can.

oci/pull.bzl
```diff
@@ -129,8 +133,17 @@ def oci_pull(name, image = None, repository = None, registry = None, platforms =
                 target_name = plat_name,
                 config_path = config_path
             )
+
             if plat in _PLATFORM_TO_BAZEL_CPU:
                 platform_to_image[_PLATFORM_TO_BAZEL_CPU[plat]] = "@" + plat_name
```
sitaktif, line 138: As a side note, I think a nice change here would be to:
- fail if *any* of the "platforms" specified in `oci_pull` is not present in `_PLATFORM_TO_BAZEL_CPU`
- let the user customise the `_PLATFORM_TO_BAZEL_CPU` so they can add their exotic CPUs to fix the failures above

The benefit of doing that is that we won't silently skip one of the provided platforms in the `alias` select() dict anymore.

(but that's out of scope for this PR 😄)

sitaktif:
Just rebased. Could the failing CI test be e.g. a stale `latest` manifest cached by `crane`?

oci/private/pull.bzl
```diff
@@ -199,28 +203,87 @@ def _oci_pull_impl(rctx):
...
+    # if the user provided a platform for the image, validate it matches the config as best effort.
+    if rctx.attr.platform:
+        config_bytes = rctx.read(config_output_path)
+        config = json.decode(config_bytes)
+
+        image_os = config.get("os", None)
+        image_architecture = config.get("architecture", None)
+        image_variant = config.get("variant", None)
+
+        attr_os, _, attr_architecture = rctx.attr.platform.partition("/")
```
thesayyn, line 243: i am sure we are doing this in multiple places. can move this into an util function?

oci/private/pull.bzl
```diff
@@ -199,28 +203,87 @@ def _oci_pull_impl(rctx):
...
+            ))
+        if not image_architecture:
+            missing_fields.append("architecture")
+        elif image_architecture != attr_architecture:
+            fail("Expected image {}/{} to have architecture '{}', got: '{}'".format(
+                rctx.attr.registry,
+                rctx.attr.repository,
+                attr_architecture,
+                image_architecture,
+            ))
```
thesayyn, line 267: I believe this would be a nice utility function imho. i'd like to keep the implementation function as lean as possible.  Goal is https://github.com/bazel-contrib/rules_oci/issues/138

oci/pull.bzl
thesayyn: wanna file an issue for it?

thesayyn: Let's keep the diff here minimal so it's easy to bisect in case something goes wrong.

oci/private/pull.bzl
sitaktif: I looks like the opposite is done in `image.bzl`:

```python
def _platform_str(os, arch, variant = None):
    parts = [os, arch]
    if variant:
        parts.append(variant)
    return "/".join(parts)
```

...but after a cursory check, I didn't see an existing platform -> triplet conversion.

I did extract this block into a function as per your sensible suggestion though 👍

oci/private/pull.bzl
sitaktif: Good point, done 👍

oci/pull.bzl
sitaktif: Sounds good - opened #502

sitaktif:
Thanks @thesayyn - I just addressed your comments.

Should I be worried about this failing test, or is a known flaky one?

thesayyn:
Those are flaky ones i believe.

oci/private/pull.bzl
```diff
@@ -192,35 +196,102 @@ def _find_platform_manifest(image_mf, platform_wanted):
             return mf
     return None
 
+def _platform_triplet(platform_str):
```
thesayyn, line 199: Can we move these into util.bzl?  Otherwise LGTM.

thesayyn: `_platform_triplet ` and `_validate_image_platform `

thesayyn (approved):
Good to after func are moved into util.bzl

thesayyn (approved):

PR #478, opened 2024-01-27
Allow user to provide their own image-refs file

siddharthab:
This allows oci_push to be used as part of bigger pipelines where images
are pushed using Bazel targets (without knowing repository, etc.), and
the image references obtained from the push step can be used in future
pipeline steps.


thesayyn:
i thought you could just do $(bazel run :push) and that would give you the ref.

siddharthab:
Unfortunately, this is not the case when the image already exists. For example, you could try running `bazel run :push` twice and the second run will not output the line to stdout.

alexeagle:
I'm a bit confused, you've edited the template but is that enough to make it usable? Is there some example or e2e here you could update to show how that use case is achieved?

thesayyn:
i think the principled answer is to always cat the refs file.

siddharthab:
Added an example/test.

More concretely, we use it in our Python pipelines as:
```py
def push_image(pusher_path: str) -> str:
    """Push an image created using Bazel and return its URI."""
    refs_file = tempfile.NamedTemporaryFile(mode="r")
    subprocess.run(
        [pusher_path, "--image-refs", refs_file.name],
        check=True,
    )
    # The refs file will have the URI refs of the images pushed, one per line.
    # The ref will look like.
    # <registry_host>/<registry>/<repo>@sha256:[sha256_hex]
    refs = refs_file.readlines()
    if len(refs) != 1:
        raise ValueError(f"multiple image pushes not yet supported: {pusher_path}")
    return refs[0]
```

thesayyn:
Looking into this again, i believe we should fix the issue instead of exposing the flag.  [Here](https://github.com/google/go-containerregistry/blob/ff385a972813c79bbd5fc89357ff2cefe3e5b43c/cmd/crane/cmd/push.go#L80) crane push seems to print ref unconditionally, that should already provide you the information you want.

siddharthab:
I am unsure where the message is getting lost then. It is very clear that the ref is not printed if an image push target is run the second time without any changes to the image.

thesayyn:
Okay, could it be that version of rules_oci that you are using is old?

siddharthab:
I am not sure anymore actually. I am no longer part of the team where I was seeing this issue.

When I try to push now in a personal GCP project, it is like the issue was never there.

cc: @calbach from my previous team, who can comment on whether this issue is present for them without this fix.

siddharthab:
Sorry, my bad. I just checked on my PR branch and did not do a `git bisect`. It seems like #490 fixed this issue. This PR is not needed anymore.

calbach:
Sorry, I missed the mention - confirming that it's working on the latest version without this fix. Thanks!

PR #480, opened 2024-01-30
Added strip to the path read

ankit-agarwal1999:
Simple bug fix to strip the path in the read config file. We had an EOF fixer that was running on our codebase and when it encountered this path, it failed to parse it. This PR adds a simple fix to strip whitespace out of the path to deal with File Not Found exceptions

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -312,7 +312,7 @@ def _get_auth_config_path(rctx):
         path = rctx.read(rctx.path(rctx.attr.config_path))
 
     if path:
-        return json.decode(rctx.read(path))
+        return json.decode(rctx.read(path).strip())
 
     return {}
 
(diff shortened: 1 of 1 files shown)


oci/private/pull.bzl
```diff
@@ -312,7 +312,7 @@ def _get_auth_config_path(rctx):
         path = rctx.read(rctx.path(rctx.attr.config_path))
 
     if path:
-        return json.decode(rctx.read(path))
+        return json.decode(rctx.read(path).strip())
```
jacktigerzhang, line 315:
```suggestion
        return json.decode(rctx.read(path.strip()))
```
Do you mean to strip newlines on the path itself instead?
json decode should already be able to handle trailing newlines

thesayyn:
I don't understand the logic behind this. Why do we need to do this? shouldn't the path not contain any whitespace?

ankit-agarwal1999:
The path itself is read from an another file. Our file had an EOL checker that would add an EOL character. When rctx reads the file path from the other file, it would have that EOL character, so it would fail to find that path. We just strip the whitespace, so that doesn't happen.

thesayyn:
I see. This is not necessary anymore as we deprecated config_path due to various problems.  See: #494
Closing as not needed

PR #483, opened 2024-02-01
chore: upgrade to Bazel 7

alexeagle:
Clear out our WORKSPACE file mostly, as bzlmod is now enabled by default.


thesayyn:
oops same as https://github.com/bazel-contrib/rules_oci/pull/458?

thesayyn:
also https://github.com/bazel-contrib/rules_oci/pull/444

alexeagle:
Evidently I keep trying the same thing and expecting a different result so that's not a good sign...

PR #485, opened 2024-02-01
doc: explain --credential_helper

### LINKED ISSUE #479: Document credential helpers
Bazel knows how to run an arbitrary program to create short-lived credentials for oci_pull.

We don't have any docs explaining how to set it up, or when to use one vs. alternative like do an explicit `login` step prior to running Bazel.

thesayyn:
Closes #479


fetch.bzl
```diff
@@ -31,6 +31,17 @@ def fetch_images():
         ],
     )
 
+    # ECR public registry
```
alexeagle, line 34: does this one require the credential helper? add a comment so we know it's testing that case

examples/credential_helper/auth.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Requirements 
+#   - curl
+#   - jq
```
alexeagle, line 5: technically also awk

examples/credential_helper/auth.sh
```diff
@@ -0,0 +1,34 @@
...
+host="$(echo $uri | awk -F[/:] '{print $4}')"
+log "Host: $host"
+
+if [[ $input == *"/token"* ]]; then
+    log "Auth: None"
+    echo "{}"
+    exit 0
+fi
+
+curl -fsSL https://$host/token | jq '{headers:{"Authorization": [("Bearer " + .token)]}}'
```
alexeagle, line 31: so the result of calling this program is something printed to stdout? more comment plz

docs/pull.md
```diff
@@ -52,6 +52,39 @@ Finally, some less-common use cases are afforded with environment variables `XDG
 See the implementation of `_get_auth_file_path` in `/oci/private/auth_config_locator.bzl` for the complete reference.
 
 
+# Authentication using credential helpers
```
alexeagle, line 55: I think it should be a lower-level header than the one above, so `##` or `###`

fetch.bzl
thesayyn: yes, it works as we know how to authenticate against public ECR. but when --credential_helper goes into affect, the authentication is done by the auth helper instead of oci_pull. I'll this as comment.

examples/credential_helper/auth.sh
thesayyn: added more comments.

alexeagle (approved):

PR #493, opened 2024-02-06
Resolve relative paths for config files

### LINKED ISSUE #492: DOCKER_CONFIG cannot be set to a workspace relative path
Migrating from rules_docker, there is a functionality gap with how docker config files are looked up. Rules_oci uses `util.file_exists` to ensure that the passed docker config file path exists on disk before using it. However, its implementation does not work with relative paths. This proves to be problematic if you have a docker config checked into the repo for pulling images using `oci_pull`

Additionally, the BZL mod implementation of `oci.pull` is missing the attribute to set the `config_path` entirely. This bug is meant to track fixing this use case of having locally checked in authentication files

mrmeku:
Fixes #492

### DIFF
--- a/oci/private/auth_config_locator.bzl
+++ b/oci/private/auth_config_locator.bzl
@@ -16,9 +16,8 @@ def _get_auth_file_path(rctx):
     if "DOCKER_CONFIG" in rctx.os.environ:
         DOCKER_CONFIG = rctx.os.environ["DOCKER_CONFIG"]
 
-    config_path = "{}/config.json".format(DOCKER_CONFIG)
-
-    if util.file_exists(rctx, config_path):
+    config_path = util.get_absolute_path(rctx, "{}/config.json".format(DOCKER_CONFIG))
+    if config_path.exists:
         return config_path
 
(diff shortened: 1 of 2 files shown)


thesayyn:
I need to check if the docker spec says anything about relative paths being allowed.

mrmeku:
The relative paths here are resolved to absolute paths, so I it should play nice with existing tooling. I think allowing relative paths is more of a stance this particular repo. They were formally more or less allowed by the now deprecated `config` attribute of `oci_pull` but with that mechanism deprecated there is no other mechanism to take its place

alexeagle (approved):
It looks right to me, in that we adapt what we accept to be the absolute path we already pass to Docker runtime.

Can you figure out where the doc change goes so that users can discover the `if(starts_with('./'))` logic?

thesayyn:
i agree that there needs to be a way to allow the config file to be relative to workspace. my concern is that DOCKER_CONFIG is not just understood by rules_oci, and i don't want our implementation to be different than the existing tools that also read DOCKER_CONFIG env.  

i am planning to do some work here on this. Can we wait till i am certain that this is the right thing?

thesayyn:
closing in favor of #494

PR #504, opened 2024-02-16
Update python.md

alexeagle:
The TODO here was done


thesayyn (approved):

alexeagle:
@thesayyn something has gone wrong with CI here. I deleted the one cache entry and it was restored, but 8.6 GB out of 10 and the CI machines are running out of disk.
`main` is green so I think something must have changed outside the repo.
I'm just going to merge this one anyway, will need to fix it this week...

PR #506, opened 2024-02-20
perf: base symlinking (bazel 7.1)

### LINKED ISSUE #505: FR: Symlink blobs from `base` image instead of copying over.
Currently when an `oci_image` depends on another one, all the blobs from the base oci_image is copied over. This was done for two reasons; 

- Avoid cascading builds. 
- Bazel did not allow symlinks inside treeartifacts that pointed outside of the treeartifact.

With https://github.com/bazelbuild/bazel/pull/21263 fixed, now we can have symlinks inside treeartifacts pointing outside.

We expect this to significantly improve build time of oci_image.

thesayyn:
fixes #505

### DIFF
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,4 +1,4 @@
-6.4.0
+7.1.0rc1
 
 # The first line of this file is used by Bazelisk and Bazel to be sure
 # the right version of Bazel is used to build and test this repo.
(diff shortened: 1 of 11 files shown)


thesayyn:
See: https://github.com/bazel-contrib/bazel_features/pull/40

Issue #284, opened 2023-07-07
401 Unauthorized during oci_pull


lbjcom:
I am migrating my project from rules_docker to rules_oci. During the migration I found that `oci_pull` raises `401 Unauthorized` in following case:

```bash
$ bazel build //:image
WARNING: Could not fetch the manifest. Either there was an authentication issue or trying to pull an image with OCI image media types.
Falling back to using `curl`. See https://github.com/bazelbuild/bazel/issues/17829 for the context.
INFO: Repository tritonserver_22.08_single instantiated at:
  /home1/irteam/bong/test/WORKSPACE:26:9: in <toplevel>
  /home/bongjinlee/.cache/bazel/_bazel_bongjinlee/9b03f7358921bbbb8549beadd376cf6a/external/rules_oci/oci/pull.bzl:133:18: in oci_pull
Repository rule oci_pull defined at:
  /home/bongjinlee/.cache/bazel/_bazel_bongjinlee/9b03f7358921bbbb8549beadd376cf6a/external/rules_oci/oci/private/pull.bzl:434:27: in <toplevel>
WARNING: Download from https://nvcr.io/v2/nvidia/tritonserver/manifests/sha256:98d08ad85f789fc0ca67c1a0e5db1356d35f096f82446df336e2bcc8fe65f1ea failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
ERROR: An error occurred during the fetch of repository 'tritonserver_22.08_single':
   Traceback (most recent call last):
        File "/home/bongjinlee/.cache/bazel/_bazel_bongjinlee/9b03f7358921bbbb8549beadd376cf6a/external/rules_oci/oci/private/pull.bzl", line 357, column 46, in _oci_pull_impl
                mf, mf_len = downloader.download_manifest(rctx.attr.identifier, "manifest.json")
        File "/home/bongjinlee/.cache/bazel/_bazel_bongjinlee/9b03f7358921bbbb8549beadd376cf6a/external/rules_oci/oci/private/pull.bzl", line 280, column 74, in lambda
                download_manifest = lambda identifier, output: _download_manifest(rctx, state, identifier, output),
        File "/home/bongjinlee/.cache/bazel/_bazel_bongjinlee/9b03f7358921bbbb8549beadd376cf6a/external/rules_oci/oci/private/pull.bzl", line 255, column 18, in _download_manifest
                _download(
[...]

PR #471, opened 2024-01-22
add docker.elastic.co registry support

cristifalcas:
add docker.elastic.co registry support

Based on https://github.com/bazel-contrib/rules_oci/issues/463

PR #511, opened 2024-02-23
feat: add nvcr.io to _WWW_AUTH

nobu-k:
I reported the issue at https://github.com/bazel-contrib/rules_oci/issues/475#issuecomment-1960581407 that `oci_pull` fails to download images from nvcr.io. As suggested in the comment at https://github.com/bazel-contrib/rules_oci/issues/284#issuecomment-1692815824, I added a new entry for nvcr.io to `_WWW_AUTH`. I locally applied the patch and confirmed that `oci_pull` successfully downloaded images which were previously failed.

I took the values from the information provided at https://forums.developer.nvidia.com/t/how-to-get-manifest-from-nvcr-io-docker-registry-using-rest-api/72385/2

### DIFF
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -48,6 +48,11 @@ _WWW_AUTH = {
         "scope": "repository:{repository}:pull",
         "service": "token-service",
     },
+    "nvcr.io": {
+        "realm": "{registry}/proxy_auth",
+        "scope": "repository:{repository}:pull",
+        "service": "{registry}",
+    },
 }
 
 def _strip_host(url):
--- a/fetch.bzl
+++ b/fetch.bzl
@@ -187,6 +187,13 @@ def fetch_images():
         ],
     )
 
+    oci_pull(
+        name = "nvidia_k8s_device_plugin_image",
+        # tag = "v0.14.4",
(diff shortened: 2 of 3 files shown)


oci/private/authn.bzl
```diff
@@ -48,6 +48,11 @@ _WWW_AUTH = {
         "scope": "repository:{repository}:pull",
         "service": "token-service",
     },
+    "nvcr.io": {
+        "realm": "{registry}/proxy_auth",
+        "scope": "repository:{repository}:pull",
+        "service": "{registry}",
```
nobu-k, line 54: Both `"service": "{registry}"` and `"service": "registry"` (as shown in the URL above) worked, so I chose the former to be consistent with other similar ones.

thesayyn:
Could you add a test for this as well?

thesayyn:
i think you need to add the new repo to the tests folder. See: https://github.com/bazel-contrib/rules_oci/pull/471#issuecomment-1904899899

nobu-k:
Thank you, I was about to ask how it should be tested.

nobu-k:
The CI error is due to "no space left". Is this a temporary error, or do I have to find out a smaller image on nvcr.io? (I'm not sure if there's one because a lot of them are >1GB)

thesayyn (approved):

PR #514, opened 2024-02-27
refactor: unmask errors in oci_image

thesayyn:
the following changes has been made to clean up accumulated mess in oci_image.

- replace `oci:layout` with `--from=` (only understood by the wrapper)
- replace `oci:empty_base` with `--scratch`  (only understood by the wrapper)
- remove `get_option` simply by using simpler flag `--scratch=platform_here`
- remove `base_from_layout` function and simply use crane push. 
- introduce secondary stderr file for registry only to separate logs from the wrapper and the registry
- move `zot` invalid manifest type detection to on_exit to simplify error detection logic
- remove input_depsets in image.bzl. it was an overkill as we were not nesting depsets at all. 
- remove deadcode
- some rearrangements of codes for shellcheck.
- introduce shellcheck, we were not using it at all!!!

cc @jasonbedard in case he is interested in reviewing.


oci/private/image.sh
```diff
@@ -1,104 +1,93 @@
...
 
 for ARG in "$@"; do
     case "$ARG" in
-        (oci:registry*) FIXED_ARGS+=("${ARG/oci:registry/$REGISTRY}") ;;
-        (oci:empty_base) FIXED_ARGS+=("$(empty_base $REGISTRY $@)") ;;
-        (oci:layout*) FIXED_ARGS+=("$(base_from_layout ${ARG/oci:layout\/} $REGISTRY)") ;;
-        (--output=*) OUTPUT="${ARG#--output=}" ;;
+        (--scratch=*)
+          # https://www.shellcheck.net/wiki/SC2155
+          REF=
```
gregmagolan, line 78: I suspect you're good without the empty `REF=` line since you're neither using `export` or `local`

gregmagolan: maybe it was the `=+` in `FIXED_ARGS+=` that was masking it before?

thesayyn: hmm, https://github.com/bazel-contrib/rules_oci/blob/65c49297bfe150b6f24db67383119c132f345b8d/oci/private/image.sh.tpl#L89-L90 we did this before.  

this seem to have fixed the issue...  but shellcheck doesn't mention anything about variables declared for the first time...

gregmagolan: tested that in a local script and that doesn't add up either 🤷

thesayyn: hmm, i'll validate this then.

gregmagolan: its unclear how the error code could have been lost in the old script... but the STDERR was clearly not outputted previously

thesayyn: i think we fixed that by https://github.com/bazel-contrib/rules_oci/pull/490

thesayyn: bash: enigma machine

gregmagolan: #490 would have only been a fix if `set -o errexit` wasn't set tho right?

gregmagolan: That was reverted in https://github.com/bazel-contrib/rules_oci/pull/499 as well

thesayyn: You are right. we were sending errors from crane push into void...

oci/private/image.sh
```diff
@@ -1,104 +1,88 @@
...
 ENV_EXPANSIONS=()
 
 for ARG in "$@"; do
     case "$ARG" in
-        (oci:registry*) FIXED_ARGS+=("${ARG/oci:registry/$REGISTRY}") ;;
-        (oci:empty_base) FIXED_ARGS+=("$(empty_base $REGISTRY $@)") ;;
-        (oci:layout*) FIXED_ARGS+=("$(base_from_layout ${ARG/oci:layout\/} $REGISTRY)") ;;
-        (--output=*) OUTPUT="${ARG#--output=}" ;;
+        (--scratch=*)
+          # https://www.shellcheck.net/wiki/SC2155
```
gregmagolan, line 77: nit: SC2155 doesn't apply here since we're not using `export` or `local`

gregmagolan (approved):
🌮

oci/private/image.sh
thesayyn: yep, i found that `+=()` was masking the error.

```
# this fails
REF=$(item)
A+=($REF)

# this doesn't fail
A+=($(item))
```

PR #515, opened 2024-02-28
fix: remove curl cache

thesayyn:
Reduces the surface area for bugs subtle caching bugs. The curl callback is only for manifests and only for a subset of manifests and these are very small files either way. The callback will be removed in 2.0 since a change in Bazel 7 will make it unecessary.


gregmagolan (approved):
🌮

PR #516, opened 2024-02-28
Add a minimal c++ example in doc

rygx:


docs/cpp.md
```diff
@@ -3,8 +3,86 @@
...
+```cpp
+#include <iostream>
 
-An example is needed! Please consider contributing one. :pray:
+int main(){
+    std::cout<<"This is a C++ example!"<<std::endl;
+}
+```
+
+To make image and container for this program, the `BUILD.bazel` file would look like this:
```
thesayyn, line 17:
```suggestion
To make a container image for this program, the `BUILD.bazel` one would have something like this:
```

docs/cpp.md
```diff
@@ -3,8 +3,86 @@
...
+
+# Packaging the binary into tar, which is needed by oci_image rule
+pkg_tar(
+    name = "tar",
+    srcs = [":example_binary"],
+)
+
+# Making image
+# C++ programs usually need some fundamental libraries such as glibc, libstdc++, etc.
+# `distroless_base` image does not provide them and C++ programs will complain about
```
thesayyn, line 41: maybe distroless_cc does?

docs/cpp.md
rygx: Revised as suggested.

docs/cpp.md
rygx: Makes sense. Changed the example to use `gcr.io/distroless/cc-debian12`

rygx:
@thesayyn is this good to move forward?

thesayyn (approved):
Thank you!!

Issue #333, opened 2023-08-17
Support for zstd layer


gdippolito:
Hello! 

Thanks for starting off this project.

I was wondering if `rules_oci` supports zstd as image layers? In the rules_docker repository there were a couple of discussions about it see [here](https://github.com/bazelbuild/rules_docker/issues/2104) and there was also an attempt to add support in this [fork](https://github.com/y0psolo/rules_docker/)

For what concerned running the containers `containerd` supports zstd  layers for quite some time now (1.5 May 2021) so it should not create any problems at runtime. Considering the speed difference of `zstd` compared with `gzip` compression I think it would give a good speed up for both image creation and extraction. 

From my understanding bazel has native support for zstd since 5.1 see [here](https://github.com/bazelbuild/bazel/pull/15087) do you think this project would support that? We are currently using `rules_docker` but we will be very happy to move to `rules_oci` if we can get some new features in. 

Thanks

ReillyBrogan:
Yes, this would be really useful. Both being able to create new image layers as zstd-compressed as well as the ability to convert existing layers to zstd (similar to the buildkit `--force-compression=zstd` argument). Zstd really is a significant improvement, not only are the layers a fair bit smaller but the decompression is much faster which leads to faster start times.

ReillyBrogan:
So I looked into how this project handles layers, and ultimately support for this needs to be added to the crane tool for it to work. `crane append` does not currently support specifying the compressor to use for the new layer (defaults to gzip), though I think my golang skills are good enough that I can probably figure out how to add it since go-containerregistry has zstd code already in the repo. 

A new crane command (`crane convert` ?) would likely need to be added to convert an existing image to use zstd-layers and this is unfortunately likely to be beyond my skills, though such a command would likely be useful to a wide variety of users beyond rules_oci.

thesayyn:
> A new crane command (`crane convert` ?) would likely need to be added to convert an existing image to use zstd-layers and this is unfortunately likely to be beyond my skills, though such a command would likely be useful to a wide variety of users beyond rules_oci.

we had discussions around layer conversion before, generally, it's a bad idea AFAICT due to changing digests, etc. This needs a RFC before an implementation.


[...]

PR #523, opened 2024-03-13
image: allow .tar.zst extension on layer tar

### LINKED ISSUE #519: Allow layers to take in files other than .tar and .tar.gz
We've been using bazel-lib tar to make layers for oci_images. It supports adding compress = "zstd".
It has actually been working fine and speeding up our builds quite a bit.
However, as I've updated bazel-lib, the generated tar file now has a .tar.zst extension instead of a .tar extension.
Only now do oci_images complain since they only accept .tar and .tar.gz files. 
I'm wondering if it should not already be allowed to take more than just those file extensions since it has been able to handle zstd compressed tars already.

alexeagle:
Fixes #519

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -51,7 +51,7 @@ oci_image(
 """
 _attrs = {
     "base": attr.label(allow_single_file = True, doc = "Label to an oci_image target to use as the base."),
-    "tars": attr.label_list(allow_files = [".tar", ".tar.gz"], doc = """\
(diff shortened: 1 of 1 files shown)


thesayyn (approved):

thesayyn:
We should have a test for this as well.

thesayyn:
Coming back to this, i don't believe this works with crane.

alexeagle:
do you think we should revert?

thesayyn:
Yeah, until we can add some tests. https://github.com/bazel-contrib/rules_oci/issues/333

alexeagle:
okay, reverted

PR #525, opened 2024-03-13
Revert "feat: use link instead of copy for blobs in oci_image_index"

thesayyn:
Reverts bazel-contrib/rules_oci#359

Context: users started seeing cryptic errors which i believe started with this commit. Will confirm that it's the case.

I have no idea why i merged this but up until 7.1 bazel did not support symlinks inside treeartifacts pointing out of it.

### DIFF
--- a/oci/private/image_index.sh.tpl
+++ b/oci/private/image_index.sh.tpl
@@ -35,7 +35,7 @@ function copy_blob() {
     local blob_image_relative_path="$3"
     local dest_path="${output_path}/${blob_image_relative_path}"
     mkdirp "$(dirname "${dest_path}")"
-    "${COREUTILS}" ln -f "${image_path}/${blob_image_relative_path}" "${dest_path}"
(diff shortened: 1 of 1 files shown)


oci/private/image_index.sh.tpl
```diff
@@ -35,7 +35,7 @@ function copy_blob() {
     local blob_image_relative_path="$3"
     local dest_path="${output_path}/${blob_image_relative_path}"
     mkdirp "$(dirname "${dest_path}")"
-    "${COREUTILS}" ln -f "${image_path}/${blob_image_relative_path}" "${dest_path}"
+    "${COREUTILS}" cat "${image_path}/${blob_image_relative_path}" > "${dest_path}"
```
alexeagle, line 38: nit: why `cat` instead of `cp` ?

alexeagle (approved):

oci/private/image_index.sh.tpl
thesayyn: They are write protected which prevents cp'ying.

PR #526, opened 2024-03-14
Support inheriting environment variable that's value contains =

sfc-gh-mhazy:
If base image contained a variable  `BASE_VAR=key=value` and one would set `MY_VAR=$BASE_VAR additional value` the `MY_VAR` would be set to `key additional value`

Fixing this by parsing `config.Env` elements as:
1. variable name - string up to first "="
2. variable value - whole string after first "="

### DIFF
--- a/examples/env_inheritance/BUILD.bazel
+++ b/examples/env_inheritance/BUILD.bazel
@@ -9,14 +9,23 @@ pkg_tar(
 )
 
 oci_image(
-    name = "image",
+    name = "middle_image",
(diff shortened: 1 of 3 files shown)


sfc-gh-ptabor (approved):
Good catch. 

@alexeagle, could you, please, take a look. This is a real problem when composite arguments (like JVM flags) are passed through ENV variables.

alexeagle (approved):
nice, thanks for added test

PR #528, opened 2024-03-18
Pass headers to Bazel downloader where available

mortenmj:
As of Bazel 7.1, it is possible to pass arbitrary headers to the Bazel downloader. Combined with proper credentials, this change makes it possible to avoid the cURL fallback on 7.1+.

It feels weird to pass the same headers to every invocation of `_download()`. Should we move `_SUPPORTED_MEDIA_TYPES` and `_DOWNLOAD_HEADERS` somewhere they could be imported in `download.bzl`?

### DIFF
--- a/oci/private/download.bzl
+++ b/oci/private/download.bzl
@@ -118,7 +118,11 @@ def _bazel_download(
         headers = {},
         # passthrough
         **kwargs):
-    return rctx.download(**kwargs)
+    # Passing headers to the downloader is only available as of 7.1
+    if versions.is_at_least("7.1.0", versions.get() or "7.1.0"):
+        return rctx.download(headers = headers, **kwargs)
+    else:
+        return rctx.download(**kwargs)
 
 download = struct(
     curl = _download,
--- a/oci/private/download.bzl
+++ b/oci/private/download.bzl
@@ -109,17 +109,12 @@ def _download(
     )
 
 # A dummy function that uses bazel downloader.
-#  Caveats
-#   - Doesn't support setting custom headers
 def _bazel_download(
         rctx,
-        # custom features
-        # buildifier: disable=unused-variable
         headers = {},
-        # passthrough
         **kwargs):
     # Passing headers to the downloader is only available as of 7.1
-    if versions.is_at_least("7.1.0", versions.get() or "7.1.0"):
+    if versions.is_at_least("7.1.0", versions.get()):
         return rctx.download(headers = headers, **kwargs)
     else:
         return rctx.download(**kwargs)
--- a/oci/BUILD.bazel
+++ b/oci/BUILD.bazel
@@ -69,6 +69,16 @@ bzl_library(
 )
 
 bzl_library(
+    name = "extensions",
+    srcs = ["extensions.bzl"],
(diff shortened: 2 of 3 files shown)


WORKSPACE
```diff
@@ -54,6 +54,11 @@ load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
 
 rules_pkg_dependencies()
 
+# For determining available Bazel features
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
+bazel_features_deps()
```
thesayyn, line 60: This should be called in `oci_register_toolchains` macro otherwise it's a breaking change.

thesayyn (changes requested):

thesayyn:
Thank you for the PR, can you rebase as well?

mortenmj:
> Thank you for the PR, can you rebase as well?

Yes, by all means

WORKSPACE
mortenmj: @thesayyn could you have another look?

thesayyn (approved):

thesayyn (approved):

thesayyn:
You need to run `bazel run :gazelle`

thesayyn:
Oh gazelle breaks some of bzl_libraries, can you revert gazelle commit and manually add the missing bzl_library deps?

kolloch:
```
WARNING: For repository 'bazel_skylib', the root module requires module version bazel_skylib@1.4.1, but got bazel_skylib@1.5.0 in the resolved dependency graph.
Loading: 
WARNING: For repository 'platforms', the root module requires module version platforms@0.0.5, but got platforms@0.0.8 in the resolved dependency graph.
ERROR: Failed to load Starlark extension '@bazel_features_version//:version.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @bazel_features_version
This could either mean you have to add the '@bazel_features_version' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: Error computing the main repository mapping: cycles detected during computation of main repo mapping
```

thesayyn:
OH with i realized we don't need bazel features because we simply use `versions.get()` which just works because we are in the repository rule context. Can we remove bazel_features dep?

kolloch:
@thesayyn you are kind of asking for this: https://github.com/bazel-contrib/rules_oci/pull/549 ;)

But of course, it feels fairer to let @mortenmj handle this.

mortenmj:
@thesayyn could you have another look?

oci/private/download.bzl
```diff
@@ -118,7 +118,11 @@ def _bazel_download(
         headers = {},
         # passthrough
         **kwargs):
-    return rctx.download(**kwargs)
+    # Passing headers to the downloader is only available as of 7.1
+    if versions.is_at_least("7.1.0", versions.get() or "7.1.0"):
```
thesayyn, line 122: can you remove the 'or' here?  i don't its a good idea to treat any unknown version as 7.1

Issue #228, opened 2023-05-06
Feature Request: Promote multi_arch example to oci_multi_arch_image rule


AttilaTheFun:
Hi folks! I feel like building a multi-arch image is a common enough workflow that it merits being a proper rule in the repo instead of just an example:
https://github.com/bazel-contrib/rules_oci/blob/main/examples/multi_arch/BUILD.bazel

This one seems pretty trivial to just pull in as a rule, though I understand it would slightly increase the surface area you have to maintain.

thesayyn:
We have plans to make oci_image_index transition an image for multiple platforms but we avoiding it until the `optional toolchains` feature becomes stable and widespread. We have concerns over introducing transitions to rules_oci because of how many users suffered when rules_docker introduced them immaturely.

thesayyn:
for now, the answer for this FR is no, unfortunately. we'll keep the issue open for tracking purposes, however.

AttilaTheFun:
@thesayyn what would you think about adding something like aspect bazel lib's platform_transition_filegroup a first-party rule? As it stands IMHO it's a bit strange that you need to either write your own rule or use a third-party dependency to cover a fairly common use case (building an image index including an image built for a different platform).

Maybe we could call it "oci_image_transition" and it would have attributes for an image and a platform and it would transition the image to that platform. I actually prefer this approach to having oci_image_index do the transition since it's more explicit what's going on. It would look like:

```
    # Builds an image with the distroless base from the tar layer.
    oci_image(
        name = "_image",
        base = "@distroless_base",
        entrypoint = ["/binary"],
        tars = [":_layer"],
    )

    # Transitions the image to linux_arm64.
    oci_image_transition(
        name = "_linux_arm64_image",
        oci_image_transition = ":_image",
        target_platform = "@io_bazel_rules_go//go/toolchain:linux_arm64",
    )

    # Transitions the image to linux_amd64.
    oci_image_transition(
        name = "_linux_amd64_image",
        image = ":_image",
        target_platform = "@io_bazel_rules_go//go/toolchain:linux_amd64",
    )

    # Builds an image index from the linux_arm64 and linux_amd64 images.
    oci_image_index(
        name = "_image_index",
        images = [
            ":_linux_arm64_image",
            ":_linux_amd64_image",
        ],
    )
```

This does still introduce a transition but in a pretty locked-down manner. And once you have a better option it could be deprecated if you want. WDYT?

thesayyn:
[...]

PR #531, opened 2024-03-19
Use platform transition in the oci_image_index rule

moroten:
The current way of listing images makes Bazel build images for a platform that is not the configured target platform, defined by --platforms. This commit changes the behaviour and applies a transition so that the images are built for the configured target platform.

New way:
```
oci_image_index(
    name = "app",
    image = ":app_linux",
    platforms = [
        "@io_bazel_rules_go//go/toolchain:linux_amd64",
        "@io_bazel_rules_go//go/toolchain:linux_arm64",
    ]
)
```

Old way:
```
oci_image_index(
    name = "app",
    image = [
        ":app_linux_amd64",
        ":app_linux_arm64"
    ],
)
```

For backwards compatibility, the old way is still possible to use.

### DIFF
--- a/docs/image_index.md
+++ b/docs/image_index.md
@@ -7,7 +7,7 @@ Implementation details for oci_image_index rule
 ## oci_image_index
 
 <pre>
-oci_image_index(<a href="#oci_image_index-name">name</a>, <a href="#oci_image_index-images">images</a>)
+oci_image_index(<a href="#oci_image_index-name">name</a>, <a href="#oci_image_index-image">image</a>, <a href="#oci_image_index-images">images</a>, <a href="#oci_image_index-platforms">platforms</a>)
 </pre>
 
 Build a multi-architecture OCI compatible container image.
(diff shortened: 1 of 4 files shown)


alexeagle:
Our original intent here was to minimize the code in rules_oci that needs to understand what the user is doing, we liked that transitions occur explicitly. OTOH I do see how this syntax sugar makes it nicer.

> Bazel build images for a platform that is not the configured target platform

I'm not sure I follow this existing bug you suggest - don't we build for the same platforms before and after this change? The same targets get transitioned in the same way?

moroten:
> > Bazel build images for a platform that is not the configured target platform
> 
> I'm not sure I follow this existing bug you suggest - don't we build for the same platforms before and after this change? The same targets get transitioned in the same way?

Consider the following example:
```starlark
cc_binary(
    name = "binary_linux_arm64",
)
pkg_tar(
    name = "app_tar_linux_arm64",
    srcs = [":binary_linux_arm64"],
)
oci_image(
    name = "app_image_linux_arm64",
    tars = [":app_tar_linux_arm64"],
)
oci_image_index(
    name = "app_index",
    images = [
        ":app_image_linux_arm64"
    ],
)
```

I would normally only have a target called `binary_linux`. Building `binary_linux_arm64` with `--platforms=//:linux_x86_64` is possible, but makes no sense as the result will be a binary `binary_linux_arm64` will not be compatible with arm64. Adding `target_compatible_with=[...arm64]` will make all `_arm64` targets unbuildable on `x86_64`, which is correct, but `:app` will also become unbuildable.

The idea here is that all the `_linux_arm64` suffixes are removed and that `bazel build :app_image --platforms=//:linux_x86_64` and `bazel build :app_image --platforms=//:linux_arm64` produce images for respective OS, even if my host is Windows.

```starlark
cc_binary(
    name = "binary",
)
pkg_tar(
    name = "app_tar",
    srcs = [":binary"],
)
oci_image(
    name = "app_image",
    tars = [":app_tar"],
)
```

The `app_index` target itself is actually compatible with any target platform, or at least multiple platforms, i.e. I expect to get the same output independent on `--platforms` (e.g. building on Windows or MacOS).

thesayyn:
https://github.com/bazel-contrib/rules_oci/issues/228#issue-1698785947

moroten:
Let's keep the discussion about the plan forward in #228.

Are the failing tests due to this change? My feeling is that it is something unrelated. Anyway, some tests should be added for this new functionality.

thesayyn:
Can we close this since we don't have to land a change for this?

thesayyn:
Okay, i am going to close this for house keeping purposes. Thank you for putting time into this.

PR #533, opened 2024-03-23
ci: only test with bazel 6, 7

thesayyn:
Changes

- Add MODULE.bzl to some e2e
- Now crane is used by default for unit testing
- Upgraded bazel-lib for https://github.com/bazelbuild/bazel-central-registry/pull/1700
- Drop bazel 5 from CI


MODULE.bazel
```diff
@@ -7,7 +7,7 @@ module(
 )
 
 # Minimum 1.36.0 to include https://github.com/aspect-build/bazel-lib/pull/594
```
alexeagle, line 9: update this comment too

MODULE.bazel.lock
alexeagle, line 1: I think you meant to `git rm` this file

alexeagle (approved):

Issue #422, opened 2023-11-21
oci_pull from private gitlab container registry


oakad:
Bazel 6.4.0/rules_oci-1.4.2

I need to pull some base images from a privately hosted gitlab registry; I do the usual "docker login"/"crane auth login" and everything works just fine.

However, I cannot use the same registry with `oci_pull`, it always fails. From the look of it, the scenario is not even supported: gitlab requires a temporary token acquisition and while big, public gitlab has an entry in `_WWW_AUTH` array, my private gitlab instance obviously has none.

So, what is the expected approach for this sort of operation?

In general, I wonder why `oci_pull` is implemented in terms of custom and incorrect http download flow which is guaranteed to keep breaking instead of using `crane pull` or some other more reliable method?

I would also like to point out that the days of public docker image registries are counted, with both Docker and Google pushing their users toward authenticated access. For rules_oci to be of any utility, authentication must be properly addressed, otherwise we will be stuck without the ability to build images with Bazel.

alexeagle:
We support credential helpers, which I think is the way to get the temporary token. @thesayyn I don't see docs about credential helpers on the repo?

> why oci_pull is implemented in terms of custom and incorrect http download flow which is guaranteed to keep breaking instead of using crane pull 
The Bazel Downloader has a bunch of features, and our goal is to rely on it exclusively. The primary feature is correct, content-based caching in the `--repository_cache`, and also the ability for Bazel remote execution systems to perform a remote download so the pull never needs to happen on the machine where Bazel runs.

thesayyn:
We are still in the phase of fixing bazel downloader to fit the needs of the rulesets. As Alex pointed out using bazel downloader is the right way to fetch remote content for various reasons.

 As for _WWW auth, the reason why it exists is clearly stated above the variable, and we can't keep expanding it forever. One solution is to allow users to provide WWW auth entry for their private registries. 
 
See: https://github.com/bazel-contrib/rules_oci/issues/126

oakad:
Credential helpers don't help here; the credentials are fine, the problem is with the redirect/reauth flow. What will be of help is support for download helper of some sort which gets docker tag in and returns image tarball out.

You pull go-containerregistry as dependency anyway. The safe approach will be to use a helper which calls into it, not the horrible bazel http support which keeps breaking on mundane proxy use.

thesayyn:
[...]

PR #539, opened 2024-04-04
feat: universal www-authenticate support

hunshcn:
Resolves https://github.com/bazel-contrib/rules_oci/issues/422

This pr implements a universal WWW-Authenticate that supports the regular authentication process (done by `crane auth token`).

Difficulty in using a private registry with authentication was a big hurdle in migrating to rules_oci, and I think I solved it.

Need discussed:
- `crane auth token` will only read `$HOME/.docker/config.json` as config, so attr `config` will be ignored when registry use Bearer challenge. This needs to be discussed. (Of course, we can simply let the original _WWW_AUTH address continue to execute the old code.)
- Similar to the previous, `crane auth token` cannot use outer config, attr `config` will be ignored. However, from another point of view, push supports config, which is a very strange thing in itself.

### DIFF
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -174,7 +174,7 @@ def _get_token(rctx, state, registry, repository, www_authenticate, toolchain):
     if not www_authenticate:
         return pattern
 
-    crane = "@{}".format(_crane_binary(rctx, toolchain)) if toolchain else "@@{}~oci~{}".format(Label(":authn.bzl").workspace_name, _crane_binary(rctx, "oci"))
+    crane = "@{}".format(_crane_binary(rctx, toolchain)) if toolchain else "@@{}~oci~{}".format(Label(":authn.bzl").workspace_name or "_main", _crane_binary(rctx, "oci"))
     image = "{}/{}".format(registry, repository)
     result = rctx.execute([Label(crane), "auth", "token", image])
     if result.return_code != 0:
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -4,61 +4,41 @@ load("@aspect_bazel_lib//lib:base64.bzl", "base64")
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
 load(":util.bzl", "util")
 
+_default_www_auth = [
+    "index.docker.io",
+    "public.ecr.aws",
+    "ghcr.io",
+    "cgr.dev",
+    ".azurecr.io",
+    "registry.gitlab.com",
+    ".app.snowflake.com",
+    "docker.elastic.co",
+    "quay.io",
+    "nvcr.io",
+]
 
-# Unfortunately bazel downloader doesn't let us sniff the WWW-Authenticate header, therefore we need to
-# keep a map of known registries that require us to acquire a temporary token for authentication.
-_WWW_AUTH = {
-    "index.docker.io": {
-        "realm": "auth.docker.io/token",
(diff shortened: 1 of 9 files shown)


oci/private/authn.bzl
```diff
@@ -2,53 +2,42 @@
...
-        "scope": "repository:{repository}:pull",
-        "service": "{registry}",
-    },
-    "docker.elastic.co": {
-        "realm": "docker-auth.elastic.co/auth",
-        "scope": "repository:{repository}:pull",
-        "service": "token-service",
-    },
-}
+_default_www_auth = [
```
hunshcn, line 8: We could actually remove the whitelist and have all requests go through the `crane auth token` logic. Can I do that?

oci/pull.bzl
```diff
@@ -103,7 +103,21 @@ _PLATFORM_TO_BAZEL_CPU = {
...
+        return toolchain
+    for r in native.existing_rules().values():
+        if r["kind"] == "crane_repositories":
+            _toolchain = r["name"][:-len(r["platform"]) - len("_crane_")]
+            if not toolchain:
+                toolchain = _toolchain
+            elif toolchain != _toolchain:
+                fail("Multiple oci toolchains found, please specify the toolchain")
+    if not toolchain:
+        fail("No oci toolchain found, please register first", native.existing_rules().values())
```
hunshcn, line 117: Now the main contents have been completed. But bzlmod's ci failed.

When this macro was called at module_extension context, native.existing_rules() has nothing. And even special the toolchain name, the repository is inaccessible.

`Unable to load package for @@[unknown repo 'oci_crane_darwin_arm64' requested from @@rules_oci~]//:crane: The repository '@@[unknown repo 'oci_crane_darwin_arm64' requested from @@rules_oci~]' could not be resolved: No repository visible as '@oci_crane_darwin_arm64' from repository '@@rules_oci~'`

I read and studied gazelle `go_repository_tools`, but I still don't understand the principle.

@alexeagle @thesayyn could you help me to fix it? Just the last step

hunshcn:
ready now

thesayyn:
i am somehow sympathetic  to this idea, let me give it more thought.

hunshcn:
post mount failed in ci, but why?

hunshcn:
The original method will pollute the MODULE.bazel.lock (platform attribute). Use canonical name to find compatible name.
It looks ready for production.

2 ci fail on post mount.

hunshcn:
@thesayyn WDYT

hunshcn:
ping @thesayyn

hunshcn:
rebased

@thesayyn

peakschris:
This looks like an improvement, thanks! Is there an ETA to get it in?

hunshcn:
I see, before 2.0 released, this pr should be open.

If auth need, this pr can be temp used.

thesayyn:
Hey, sorry after giving this a thorough thought, i don't think we want to do this for the reasons i will do my best to explain. 

- This creates a  hard dependency on crane which we can't remove later on because now it will be leaking into the process oci_pull, meaning that any authorization error will make us ask two questions; is it caused by crane or oci_pull's builtin auth.  

- This is already possible to do without changing oci_pull at all, you can simply use a `--credential_helper` [script which essentially does the same thing.](examples/credential_helper/auth.sh) 

```
common  --credential_helper=index.docker.io=%workspace%/credential_helper.sh
```

- This prevents offline builds, as now any http request made by crane isn't tracked by Bazel, making it impossible to have offline builds. 


Thank you for the effort though!

PR #540, opened 2024-04-05
Allow use rule with multiple files output for `labels` and `attributes` image attributes

bozaro:
I want to add two kind of labels:
 - labels for image lifecycle management - on this labels change image must be rebuild;
 - labels for image stamping (from which Git revision I can build same image) - this lables change everytime and should be excluded from cache key.

This modification allow to me create two labels file and pass it by custom or `filegroup` rule:
 - `labels.txt` - with labels for image lifecycle management;
 - `labels_volatile.txt` - with stamping labels.

After that I can add `labels_volatile.txt` to `omitted_inputs` srubbing rule and cache will ignore every `labels_volatile.txt` file change.


thesayyn:
I am not sure i follow, I don't think this change matches the description. Can you please file an issue first so we can discuss it before agree to the change?

PR #542, opened 2024-04-15
feat(oci_pull): expose the digest as a file

aignas:
This makes the external image repos also expose the digest file, similarly how
the `oci_image` creates a `.digest` target. The actual name of the file can be
something else and I am just putting this out to initiate a discussion.

Let's say a user has `@alpine` image defined via `oci_pull`, they can access the
digest of the upstream image by accessing the `@alpine//:image.digest` file.
We could similarly have the platforms of the image written to a file as well,
but I am not sure if that is useful.

Having the `image.digest` in a file could be useful for various automation that
is done using `write_source_file` rule where one would use a `genrule` to get
the `digest` value from the `@alpine//:image.digest` target, transform it and
output it to a file that is later synced with `write_source_file`. Not needing
to depend on extra CLIs that might not be available to the user of `rules_oci`
as toolchains could be good here.

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -345,7 +345,7 @@ def _oci_alias_impl(rctx):
     available_platforms = []
 
     manifest, _, digest = downloader.download_manifest(rctx.attr.identifier, "mf.json")
-    rctx.file("digest.txt", digest + "\n")
+    rctx.file("digest.txt", digest)
 
     if rctx.attr.platforms:
         if manifest["mediaType"] in _SUPPORTED_MEDIA_TYPES["index"]:
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -303,8 +303,9 @@ oci_pull = repository_rule(
 )
 
 _MULTI_PLATFORM_IMAGE_ALIAS_TMPL = """\
-exports_files(
-    ["digest"],
+filegroup(
+    name = "digest",
+    srcs = ["digest"],
     visibility = ["//visibility:public"],
 )
 
@@ -319,8 +320,9 @@ alias(
 """
 
 _SINGLE_PLATFORM_IMAGE_ALIAS_TMPL = """\
-exports_files(
-    ["digest"],
+filegroup(
+    name = "digest",
+    srcs = ["digest"],
     visibility = ["//visibility:public"],
 )
 
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -303,6 +303,11 @@ oci_pull = repository_rule(
 )
 
 _MULTI_PLATFORM_IMAGE_ALIAS_TMPL = """\
+exports_files(
+    ["image.digest"],
+    visibility = ["//visibility:public"],
+)
+
 alias(
     name = "{target_name}",
     actual = select(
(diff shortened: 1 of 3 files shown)


alexeagle (commented):
Nice, do you mind also adding some usage of it in the e2e or examples, so we have some form of test coverage that we don't drop or break this in the future?

oci/private/pull.bzl
```diff
@@ -272,6 +272,11 @@ oci_pull = repository_rule(
 )
 
 _MULTI_PLATFORM_IMAGE_ALIAS_TMPL = """\
+exports_files(
```
thesayyn, line 275: can we make this a filegroup(name = "digest") instead?

thesayyn:
@aignas do you want to finish this PR?

aignas:
Could you please give some guidance on where the tests should go and if I am expected to bump the lock files?

I would still love the feature, but if someone would like to take over and finish this PR, I am OK with that.

aignas:
The tests are now passing, @thesayyn, let me know what is left to do here:
```console
$ bazel test :test_bzlmod
INFO: Analyzed target //:test_bzlmod (125 packages loaded, 11031 targets configured).
INFO: Found 1 test target...
Target //:test_bzlmod up-to-date:
  bazel-bin/test_bzlmod_bats.sh
INFO: Elapsed time: 210.224s, Critical Path: 151.69s
INFO: 32 processes: 10 internal, 16 darwin-sandbox, 6 local.
INFO: Build completed successfully, 32 total actions
//:test_bzlmod                                                           PASSED in 64.0s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.

$ bazel test :test_nobzlmod
INFO: Analyzed target //:test_nobzlmod (0 packages loaded, 1 target configured).
INFO: Found 1 test target...
Target //:test_nobzlmod up-to-date:
  bazel-bin/test_nobzlmod_bats.sh
INFO: Elapsed time: 69.553s, Critical Path: 66.46s
INFO: 6 processes: 4 internal, 2 local.
INFO: Build completed successfully, 6 total actions
//:test_nobzlmod                                                         PASSED in 66.2s

Executed 1 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

oci/private/pull.bzl
aignas: Making it a `filegroup` may make it harder to be used within repository rule context. Is there any problem with it being a file instead?

thesayyn: I'd imagine this target being only used in the build rule context, may i ask the reason why just reading the index.json file created oci_pull isn't an option?

I may be paranoid, but my concerns 

- This wouldn't (probably) work with RBE
- This is a file laying around in the repository directory, easy to break and easy to accidentally override.
- It's not likely to work for a confused user because of `alias` created by oci_alias target

aignas: Fixed this to use a `filegroup`, it works for my usecase.

thesayyn (approved):

oci/private/pull.bzl
```diff
@@ -333,6 +345,7 @@ def _oci_alias_impl(rctx):
     available_platforms = []
 
     manifest, _, digest = downloader.download_manifest(rctx.attr.identifier, "mf.json")
+    rctx.file("digest.txt", digest + "\n")
```
thesayyn, line 348: what's the reason for `\n` here?  I would expect `$(cat <path_to_digest_file>)` to have no trailing NLs? can we remove this?

aignas: Yeah we can. I though having an EOL character at the end of the file may make it nice.

```suggestion
    rctx.file("digest.txt", digest)
```

PR #543, opened 2024-04-17
fix: Add quay back to authn config 

ankit-agarwal1999:
Looks like the quay config got lost when refactoring `pull.bzl` to `authn.bzl` this PR adds it back


thesayyn:
Could add a test pull to fetch.bzl so that this is caught next time?

thesayyn:
Oh i didn't realize we had it. Unfortunately we don't have a test suite that ensures private images are accessible.

ankit-agarwal1999:
rebase is done, can we merge then?

thesayyn (approved):

PR #547, opened 2024-04-25
Draft: add container_run_and_save rule

alexconrey:
# Overview
Adds `container_run_and_save` rule. This rule is meant to be a mediocre "replacement" for the `container_run_and_commit` rule out of `rules_docker`.

## Why?
* Few/no replacements exist for `container_run_and_*` rules offered by `rules_docker`
* `rules_docker` is not actively maintained
* Updating a base image such as Ubuntu may not be as easy as decompressing debs in the event of widespread dependency requirements (glibc/etc)

## Concessions/Caveats
* Uses similar exec fallback as `oci_tarball`
  * Attempts to find a binary via `loader` parameter, falls back to $PATH if missing
* Must pass a tarfile as a base image to the `container_run_and_save` rule.
  * Ideally this would just be passing in a label to an `oci_image` like other rules
* There are probably other/better ways to do this
  * This works ™️ 
  * This provides an option for people coming from `rules_docker` who need to execute commands in their containers
  * This provides a workaround for a well-known issues in `rules_oci` which is the lack of this and similar rules matching `rules_docker` functionality
* This probably does less than `rules_docker` implementation
  * Something is better than nothing

### Other thoughts
* If this is something that the maintainers would like to avoid the overhead on maintaining and supporting, I would ask to then consider creating these files as a patch which can then be optionally applied in a WORKSPACE file. This would allow for the feature to exist but obviate the need to support it as a GA feature.
* `gazelle` did it's thing and touched a few more corners of this than I was intending. I can revert these changes if desired.

# Usage
```
oci_image(
    name = "example_ubuntu_base",
    base = "@ubuntu",
    entrypoint = ["/bin/bash", "-c"],
    env = {
        "TZDATA": "Etc/Utc",
        "DEBIAN_FRONTEND": "noninteractive",
    },
)

oci_tarball(
    name = "example_ubuntu_base_tarball",
    image = ":example_ubuntu_base",
    repo_tags = ["example-ubuntu-base:latest"]
)

container_run_and_save(
    name = "update_packages",
    base_tarball = ":example_ubuntu_base_tarball",
    cmd = "apt update; apt upgrade -y"
)

oci_image(
    name = "image",
    base = ":example_ubuntu_base",
    tars = [
        ":update_packages",
    ]
)

oci_tarball(
    name = "tarball",
    image = ":image",
    repo_tags = ["updated-ubuntu-base:latest"]
)
```

Produces an image which is up to date and compliant with basic security scans:
```
$ docker run --rm -it  updated-debian-base:latest "sh -c 'apt update; apt upgrade'"
Hit:1 http://archive.ubuntu.com/ubuntu jammy InRelease
Get:2 http://security.ubuntu.com/ubuntu jammy-security InRelease [110 kB]
Get:3 http://archive.ubuntu.com/ubuntu jammy-updates InRelease [119 kB]
Hit:4 http://archive.ubuntu.com/ubuntu jammy-backports InRelease    
Get:5 http://archive.ubuntu.com/ubuntu jammy-updates/universe amd64 Packages [1370 kB]
Get:6 http://archive.ubuntu.com/ubuntu jammy-updates/main amd64 Packages [2035 kB]                                                                                                 
Fetched 3634 kB in 8s (444 kB/s)                                                                                                                                                   
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
All packages are up to date.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -21,3 +21,13 @@ bzl_library(
         "@bazel_tools//tools/build_defs/repo:utils.bzl",
     ],
 )
+
+bzl_library(
+    name = "fetch",
+    srcs = ["fetch.bzl"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "@bazel_tools//tools/build_defs/repo:http.bzl",
+        "@rules_oci//oci:pull",
+    ],
+)
--- a/cosign/private/BUILD.bazel
+++ b/cosign/private/BUILD.bazel
@@ -11,21 +11,15 @@ exports_files([
 ])
 
 bzl_library(
-    name = "sign",
-    srcs = ["sign.bzl"],
-    visibility = [
-        "//cosign:__subpackages__",
-        "//docs:__pkg__",
-    ],
+    name = "attest",
+    srcs = ["attest.bzl"],
+    visibility = ["//cosign:__subpackages__"],
 )
 
 bzl_library(
-    name = "attest",
-    srcs = ["attest.bzl"],
-    visibility = [
-        "//cosign:__subpackages__",
-        "//docs:__pkg__",
-    ],
+    name = "sign",
+    srcs = ["sign.bzl"],
+    visibility = ["//cosign:__subpackages__"],
 )
 
 bzl_library(
--- a/examples/run_commands/BUILD
+++ b/examples/run_commands/BUILD
@@ -1,0 +1,37 @@
+load("//oci:defs.bzl", "oci_image", "oci_tarball", "container_run_and_save")
+
+oci_image(
+    name = "example_ubuntu_base",
+    base = "@ubuntu",
+    entrypoint = ["/bin/bash", "-c"],
+    env = {
+        "TZDATA": "Etc/Utc",
+        "DEBIAN_FRONTEND": "noninteractive",
+    },
+)
+
+oci_tarball(
+    name = "example_ubuntu_base_tarball",
+    image = ":example_ubuntu_base",
+    repo_tags = ["example-ubuntu-base:latest"]
+)
+
+container_run_and_save(
+    name = "update_packages",
+    base_tarball = ":example_ubuntu_base_tarball",
+    cmd = "apt update; apt upgrade -y"
+)
+
+oci_image(
+    name = "image",
+    base = ":example_ubuntu_base",
+    tars = [
+        ":update_packages",
+    ]
+)
+
+oci_tarball(
+    name = "tarball",
+    image = ":image",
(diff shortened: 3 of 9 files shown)


thesayyn:
Thanks for putting this together,  however our position on this rule is clear here: https://github.com/bazel-contrib/rules_oci?tab=readme-ov-file#comparison-with-rules_docker

It's not likely we will commit this to our public api, especially for use cases such as `apt-get install`, as rules such as rules_distroless and rules_apko exist.

That said, to make it easier for people coming from rules_docker, we could accept this as an example here that is not exported as public api.

alexconrey:
@thesayyn is there any way in which this PR could be tagged/labeled "better" such that it'd catch more eyes? A few hundred browser tabs of searching later, something with the right visibility might save some weary engineers.

I would say further to that point, while `rules_distroless` and `rules_apko` are pretty awesome - they don't solve the entirety of the problem. The problem statement is: "I need to provide $engineer with an Ubuntu image that is up-to-date" and I don't know that the industry response of "use Alpine instead" is appropriate.

I'd love to use Alpine/Distroless for all of the things, but some organizations have adopted policies which guide what and how distributions can be used, be it for licensing or other reasons. Other organizations just simply don't have the engineering capacity to say "I know we used to use Ubuntu, but if we want Bazel to keep working - now we have to use Alpine."

Thank you for the quick response on this! Hopefully it prompts useful conversation around real-world usage of these rules and can save some folks some time when searching for a solution.

thesayyn:
rules_distroless naming is a bit unfortunate, but it's more than just distroless, it supports *any* Debian based distribution, it's a *partial* replacement for apt-get install.

As for building base image, using a non-reproducible Dockerfile to build base image is a bad make things go south really quickly. 

> I don't know that the industry response of "use Alpine instead" is appropriate.

I agree with you on that, I am simply pointing out that these two alternatives cover i want to install some system packages, whether it's alpine or debian.

Reason that such rule doesn't exist here is that it's the wrong thing to do under Bazel, instead of encouraging it, we should tell people to do right thing.

Our readme mentions this use case here: https://github.com/bazel-contrib/rules_oci?tab=readme-ov-file#install-system-packages

thesayyn:
It's not likely we will ever implement a rule like container_run_and_save. Now that we have an example of how to this in genrule using buildx, it's less pressing for us to introduce this as an first-class rule here. 

Closing in favor of https://github.com/bazel-contrib/rules_oci/pull/570.

PR #548, opened 2024-04-26
feat: don't produce big tar file as default output of oci_tarball

alexeagle:
This is expected to be a big performance improvement because it doesn't overload the remote cache with these files, that only should be needed for local dev workflow of load'ing the image into a container runtime.

### DIFF
--- a/oci/private/tarball.bzl
+++ b/oci/private/tarball.bzl
@@ -15,7 +15,7 @@ and then run it in a container like so:
 
 ```
 bazel run :tarball
-docker run my-repository:latest
+docker run --rm my-repository:latest
 ```
 """
 
--- a/oci/private/tarball.bzl
+++ b/oci/private/tarball.bzl
@@ -100,10 +100,10 @@ def _tarball_impl(ctx):
     jq = ctx.toolchains["@aspect_bazel_lib//lib:jq_toolchain_type"].jqinfo
 
     image = ctx.file.image
-    tarball = ctx.actions.declare_file("{}/tarball.tar".format(ctx.label.name))
     mtree_spec = ctx.actions.declare_file("{}/tarball.spec".format(ctx.label.name))
     bsdtar = ctx.toolchains["@aspect_bazel_lib//lib:tar_toolchain_type"]
     executable = ctx.actions.declare_file("{}/tarball.sh".format(ctx.label.name))
+
     # Represents either manifest.json or index.json depending on the image format
     image_json = ctx.actions.declare_file("{}/_tarball.json".format(ctx.label.name))
     repo_tags = ctx.file.repo_tags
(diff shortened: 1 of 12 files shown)


oci/private/tarball.sh.tpl
```diff
@@ -101,15 +98,11 @@ for LAYER in $(${JQ} -r ".[]" <<< $LAYERS); do
   add_to_tar "${IMAGE_DIR}/blobs/${LAYER}" "blobs/${LAYER}.tar.gz"
 done
 
-manifest_json=$(mktemp)
```
alexeagle, line 104: as above, `mktemp` doesn't work now that the work is split into two actions.
This time we make it an explicit output of this action, and input to the `tar` action.

oci/private/tarball.bzl
```diff
@@ -130,7 +175,8 @@ def _tarball_impl(ctx):
         runfiles.append(ctx.file.loader)
 
     return [
-        DefaultInfo(files = depset([tarball]), runfiles = ctx.runfiles(files = runfiles), executable = exe),
+        DefaultInfo(files = depset([mtree_spec]), runfiles = ctx.runfiles(files = runfiles), executable = exe),
```
alexeagle, line 178: `runfiles` here does point to the output of the `tar` action. So I'm not sure this PR actually does what we want, unless the user builds with `--nobuild_runfile_links` ?

oci/private/tarball.sh.tpl
```diff
@@ -48,9 +49,7 @@ if [[ "${FORMAT}" == "oci" ]]; then
   # Handle multi-architecture image indexes.
   # Ideally the toolchains we rely on would output these for us, but they don't seem to.
 
-  layout_file=$(mktemp)
-  echo -n '{"imageLayoutVersion": "1.0.0"}' > "$layout_file"
-  add_to_tar "$layout_file" oci-layout
+  add_to_tar "{{oci_layout}}" oci-layout
```
thesayyn, line 52: the oci_image that's being added to the tarball already has a oci-layout file in, we should use that instead.

alexeagle: Can you tell me what change to make then? Do we need to copy it from somewhere? just delete this `add_to_tar`?

thesayyn: i think you just need to change "{{oci_layout}}" to "${IMAGE_DIR}/oci-layout"

oci/private/tarball.bzl
alexeagle: confirmed that even with that flag, the tar is still created because `runfiles` are included in the semantics of "default outputs"

alexeagle:
@thesayyn we have to figure out whether we want to use `tags=manual` and if so, whether there's any point splitting the mtree action from the tar action

oci/private/tarball.bzl
```diff
@@ -98,20 +127,30 @@ def _tarball_impl(ctx):
...
+    tar_args = ctx.actions.args()
+    tar_args.add_all(["--create", "--no-xattr", "--no-mac-metadata"])
+    tar_args.add("--file", tarball)
+    tar_args.add(mtree_spec, format = "@%s")
+    ctx.actions.run(
+        executable = bsdtar.tarinfo.binary,
+        inputs = depset(direct = mtree_outputs, transitive = [tar_inputs]),
+        outputs = [tarball],
+        arguments = [tar_args],
+        mnemonic = "Tar",
```
illicitonion, line 153:
```suggestion
        mnemonic = "OCITarball",
```

Or at least I'd mention OCI in the mnemonic somewhere - we've found we often want to treat "docker packaging (probably huge)" separately from "just tarring some random files" with things like `--modify_execution_info`

oci/private/tarball.bzl
illicitonion: I can't really imagine a world where by default you would want a built docker tar in runfiles where it would be a lot of hassle to grab the output group... Docker tar + runfiles feels quite niche.

illicitonion (approved):
Seems pretty reasonable to me!

oci/private/tarball.bzl
alexeagle: Sahin suggested that we just run `tar` at runtime when you execute the loader script, that's the correct answer because more sophisticated per-layer loaders don't need a tar input at all.

oci/private/tarball_run.sh.tpl
```diff
@@ -14,4 +13,8 @@ else
     exit 1
 fi
 
+# TODO: some loader implementations don't need a tar input, so this might be wasted time
+readonly IMAGE="$(mktemp -u).tar"
```
thesayyn, line 17: this will leak, we probably want to remove this file once it's loaded.

thesayyn (approved):
A small nit, looks great!

oci/private/tarball_run.sh.tpl
thesayyn: we could use the `docker load -i <(tar )` pattern here if we don't want a temporary file at all.

Issue #285, opened 2023-07-07
Registry fails to initialize for highly parallelized builds


hanikesn:
We're building around 168 oci_images in our monorepo build. During one of our CI runs I saw:

```
bazel-out/k8-opt/bin/XXX_image.sh: line 60: $2: unbound variable
```
See: https://github.com/bazel-contrib/rules_oci/blob/main/oci/private/image.sh.tpl#L60

I assume this happened because the registry didn't start in time and timed out. I'm wondering whether there's a safer way to initialize the registry or a way to increase the timeout.

thesayyn:
168 doesn't seem a lot and shouldn't put pressure on the number of TCP ports open.  Ideally, we can make the registry toolchain to allow changing the hardcoded timeout which is 5sec, or have a retry mechanism with an exponential backoff mechanism.

hanikesn:
> 168 doesn't seem a lot
For this part of the build we generate a fairly high load avg of 160 

We're running around 48 builds/runs in parallel on a 48 core machine, which might slow down things enough for the startup to take longer.

The following patch "fixes" the issues for now on our side and can easily be applied via the `patches` field on `http_archive` rule for oci_rules:
```
--- oci/private/image.sh.tpl  2023-07-10 18:24:57.019088204 +0200
+++ oci/private/image.sh.tpl	2023-07-10 18:26:54.093602950 +0200
@@ -85,3 +85,3 @@
 source "${REGISTRY_LAUNCHER}"
-readonly REGISTRY=$(start_registry "${STORAGE_DIR}" "${STDERR}")
+readonly REGISTRY=$(start_registry "${STORAGE_DIR}" "${STDERR}" 20)
```

hanikesn:
Ok, we saw this again, but with a more concrete error message:
```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
HTTP port 0
panic: too many open files

goroutine 1 [running]:
zotregistry.io/zot/pkg/cli.newServeCmd.func1(0xc0002a2a00?, {0xc000bb8dd0, 0x1, 0x1?})
	zotregistry.io/zot/pkg/cli/root.go:54 +0xbe
github.com/spf13/cobra.(*Command).execute(0xc0002a2a00, {0xc000bb8da0, 0x1, 0x1})
	github.com/spf13/cobra@v1.5.0/command.go:876 +0x67b
[...]

Issue #176, opened 2023-04-21
Be able to use crane registry even if zot toolchain is registered


alexeagle:
Instead of https://github.com/bazel-contrib/rules_oci/blob/main/e2e/smoke/WORKSPACE.bazel#L19-L21

thesayyn:
dropping this from the 1.0 milestone as we don't have a concrete design for how we'd like to do this and this can be tackled later.

Issue #170, opened 2023-04-20
zot isn't ready in time (5s), perhaps this should be configurable?


loosebazooka:
On large builds, sometime I see errors like:
```
registry didn't become ready within 5s.
```

I don't know if zot needs more time to start up, but could this be configurable?

thesayyn:
I have been seeing this happening on our GHA CI as well. I agree that this should be configurable.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/550

Issue #148, opened 2023-04-06
registry implementations should print actionable errors 


thesayyn:
especially when INVALID_MANIFEST error is returned to the client.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/550

PR #550, opened 2024-05-02
refactor: registryless oci_image

### LINKED ISSUE #484: FR: Registryless oci_image
Currently, we spin up a registry instance to run the crane, which leads to higher build times.  We are working to change crane a little bit to allow operating on oci-layout directly.

thesayyn:
BREAKING CHANGES

- Uncompressed archives in the `tars` attribute is no longer compressed when creating the image, this might make images bigger compared to `1.x`. Use `pkg_tar#compression` or `tar#compress` attribute to compress the layers beforehand.
- Digests of images created `1.x` rules_oci will not match `2.x` rules_oci, as we are no longer using crane to create the images. However, images should stay the same semantically.

CHANGES

- `OCIImage` action no longer spins an ephemeral registry instance to mutate images locally, making it slightly faster.
- `oci_image` now supports `zstd` compression 
-  fixes #484  and closes https://github.com/bazel-contrib/rules_oci/issues/410, https://github.com/bazel-contrib/rules_oci/issues/285, https://github.com/bazel-contrib/rules_oci/issues/176, https://github.com/bazel-contrib/rules_oci/issues/170, https://github.com/bazel-contrib/rules_oci/issues/148, https://github.com/bazel-contrib/rules_oci/issues/107

CHECKLIST 

- [x] detailed description
- [x] add the layer with correct media type when base image is docker + add a test for it.
- [x] use hermetic zstd
- [x] run against distroless to check breakages, ask some of our customers to see if breaks anything. 
- [x] run a benchmark


oci/private/image.sh
```diff
@@ -1,144 +1,168 @@
...
+    CONFIG=$(jq --arg user "${ARG#--user=}" '.config.User = $user' <<<"$CONFIG")
+    ;;
+  --workdir=*)
+    CONFIG=$(jq --arg workdir "${ARG#--workdir=}" '.config.Workdir = $workdir' <<<"$CONFIG")
+    ;;
+  --labels=*)
+    CONFIG=$(jq --rawfile labels "${ARG#--labels=}" '.config.Labels = ($labels | split("\n") | map(. | split("=")) | map({key: .[0], value: .[1:] | join("=")}) | from_entries)' <<<"$CONFIG")
+    ;;
+  --annotations=*)
+    # while IFS= read -r in || [ -n "$in" ]; do
```
thesayyn, line 157: TODO

mortenmj:
Did a quick test here. Everything builds/tests fine, but running an oci_tarball gives an error like this:

```
open /var/lib/docker/tmp/docker-import-1867629837/blobs/sha256/a74371ebe00e8fad96ff8f7ec8421e280fd0828ecdd38457c8a2f7c1ceb1b870.tar.gz: no such file or directory
```

thesayyn:
yeah, there's an unrelated change in this PR. thanks for the info!

thesayyn:
@mortenmj did you see any significant speed-up/slow-down, do you have an before-after?

mortenmj:
> @mortenmj did you see any significant speed-up/slow-down, do you have an before-after?

It seemed fast, but I didn't do any formal testing

MODULE.bazel
```diff
@@ -12,9 +12,16 @@ bazel_dep(name = "platforms", version = "0.0.8")
 
 oci = use_extension("//oci:extensions.bzl", "oci")
 oci.toolchains()
-use_repo(oci, "oci_crane_registry_toolchains", "oci_crane_toolchains")
+use_repo(oci, "oci_crane_toolchains", "oci_regctl_toolchains")
 
-register_toolchains("@oci_crane_toolchains//:all", "@oci_crane_registry_toolchains//:all")
+register_toolchains("@oci_crane_toolchains//:all", "@oci_regctl_toolchains//:all")
+
+# Workaround https://github.com/aspect-build/bazel-lib/pull/832
```
alexeagle, line 19: can remove workaround now?

MODULE.bazel
```diff
@@ -12,9 +12,16 @@ bazel_dep(name = "platforms", version = "0.0.8")
 
 oci = use_extension("//oci:extensions.bzl", "oci")
 oci.toolchains()
-use_repo(oci, "oci_crane_registry_toolchains", "oci_crane_toolchains")
+use_repo(oci, "oci_crane_toolchains", "oci_regctl_toolchains")
```
alexeagle, line 15: what if we didn't rename `oci_crane_registry_toolchains`?

examples/big_image/BUILD.bazel
```diff
@@ -1,38 +1,60 @@
...
 )
 
 [
-    pkg_tar(
+    tar(
+        name = "blayer_%s" % i,
+        srcs = ["3gib_file"],
+        compress = "gzip",
+        mtree = [
+            "%s.ttxt uid=0 gid=0 mode=0755 type=file contents=$(location :3gib_file)" % i,
```
alexeagle, line 26: nit: `ttxt` intentional?

examples/env_inheritance/BUILD.bazel
```diff
@@ -45,9 +45,6 @@ genrule(
 assert_contains(
     name = "check_digest",
     actual = ":hash",
-    expected = "sha256:d15c83f8f8438e5e05ff40478dc98ad923e9afa272b6630c5b29baf6f7e93585",
-    # TODO: enable once there's a new version of crane
-    # https://github.com/google/go-containerregistry/issues/1542
-    # https://github.com/bazel-contrib/rules_oci/issues/58
+    expected = "sha256:1c5b53cfc93dfa8199a016d305f2c69d43b4653659dc273009690cb157a74683",
     tags = ["manual"],
```
alexeagle, line 49: can remove manual?

e2e/pull/wksp/MODULE.bazel
```diff
@@ -11,7 +11,7 @@ local_path_override(
 oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")
 oci.pull(
     name = "empty_image",
-    digest = "sha256:2d4595bbc0fabeb1489b1071f56c26f44a2f495afaa9386ad7d24e7b3d8dfd3e",
+    digest = "sha256:f7f4df35452e2f7fa0906fd45902f46b35e0c6e95d41ed891df31f4937ba1632",
```
alexeagle, line 14: are you forced to update these?

oci/private/image.sh
```diff
@@ -1,144 +1,174 @@
 #!/usr/bin/env bash
 set -o pipefail -o errexit -o nounset
 
+# Add hermetically built jq, regctl, coreutils into PATH.
+PATH="{{jq_path}}:$PATH"
```
alexeagle, line 5: is there a reason to put them on PATH rather than call them using absolute paths? Is it to ensure that subprocesses see them as well?

oci/private/image.sh
```diff
@@ -1,144 +1,174 @@
...
-        filter+=" | .variant = \"${platform[2]}\""
+function add_layer() {
+  local path=
+  local media_type=
+  local compression=
+  local digest=
+  local diffid=
+  local size=
+
+  path=$(realpath "$1")  # TODO: might not be necessary.
```
alexeagle, line 73: decide before landing?

oci/private/image.sh
```diff
@@ -1,144 +1,174 @@
...
+  digest=$(regctl digest <"$path")
+  diffid="$digest"
+  size=$(wc -c "$path" | awk '{print $1}')
+
+  if [[ $(coreutils od -An -t x1 --read-bytes 2 "$path") == " 1f 8b" ]]; then
+    compression="gzip"
+    diffid=$(zstd --decompress --format=gzip <"$path" | regctl digest)
+  elif zstd -t "$path" 2>/dev/null; then
+    compression="zstd"
+    diffid=$(zstd --decompress --format=zstd <"$path" | regctl digest)
```
alexeagle, line 83: nit: the diffid assignments are the same if you use `$compression` and could be consolidated after the conditional?

oci/private/image.sh
```diff
@@ -1,144 +1,174 @@
...
+  --exposed-ports=*)
+    CONFIG=$(jq --rawfile ep "${ARG#--exposed-ports=}" '.config.ExposedPorts = ($ep | split(",") | map({key: ., value: {}}) | from_entries)' <<<"$CONFIG")
+    ;;
+  --user=*)
+    CONFIG=$(jq --arg user "${ARG#--user=}" '.config.User = $user' <<<"$CONFIG")
+    ;;
+  --workdir=*)
+    CONFIG=$(jq --arg workdir "${ARG#--workdir=}" '.config.WorkingDir = $workdir' <<<"$CONFIG")
+    ;;
+  --labels=*)
```
alexeagle, line 156: if it's no longer a file, does that mean you can't `--stamp` values into them?

examples/env_inheritance/BUILD.bazel
thesayyn: Yes

e2e/pull/wksp/MODULE.bazel
thesayyn: Yes, because digests created in 1.x is different than 2.x, these should be stable from now, if not it'll catch some accidental changes we might make that changes the digest  and not realize.

oci/private/image.sh
thesayyn: I like them being explicitly called, makes it easier to understand.

oci/private/image.sh
thesayyn: All that said, i can take that back if you have a strong reason.

MODULE.bazel
thesayyn: Will do it a follow up, when we cut a new release of bazel-lib.

thesayyn: Not necessary to cut a release just for this.

thesayyn:
Benchmark: distroless building all images `5m 17s` -> `3m 59s`

thesayyn:
Breaking changes are observed here: https://github.com/GoogleContainerTools/distroless/pull/1586#issuecomment-2097035728 🌮

thesayyn:
PTAL again, this is now ready.

oci/private/image.sh
alexeagle: nope that answers it, if they're still files then they can be stamped 👍🏻

MODULE.bazel
alexeagle: ah my mistake

alexeagle (approved):

hanikesn:
Great work!

hunshcn:
There seems to be an error in the description, which should be `pkg_tar#extension`.

PR #551, opened 2024-05-02
refactor: remove registry references

thesayyn:
Prefactor for #550 

Changes: 

- Removes LATEST_ZOT_VERSION, LATEST_CRANE_VERSION from public api, no longer configurable. 
- Removes ZOT as a registry implementation, leaving only CRANE to keep 2.x branch green
- Removes LAUNCHER_WRAPPER from tests to make it easier to drop the remaining registry toolchains
- Removes some e2e tests for registry toolchains
- Changes BCR presubmit to be run against `e2e/smoke` rather than `e2e/crane_as_registry`
- Replaces non-hermetic `xz` call in genrule to be bsdtar instead.


alexeagle (approved):

mering:
Will v2 drop `WORKSPACE` compatibility? If not how should users define the crane version for `oci_register_toolchains()` without the `LATEST_CRANE_VERSION` variable?

PR #557, opened 2024-05-07
Reduce GHA caching

alexeagle:
The --disk_cache flag produces large outputs and causes flaky CI when we run out of disk.


thesayyn:
Do we want to make this against 2.x?

alexeagle:
Yeah I will pick it up when I do the cherrypicks

thesayyn (approved):

Issue #510, opened 2024-02-22
FR: Calculate digest and diffid outside of oci_image 


thesayyn:
Currently, when oci_image assembles an image, it spends most of the time calculating DiffID and Digest of a layer. Since everything is being done in a single action, calculations are repeated everytime an input changes. 

This is not ergonomic in terms of cacheability of these calculations and parallelization of machine resources.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/558

PR #558, opened 2024-05-07
feat: implement out-of-band descriptor calculation

thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/510


oci/private/descriptor.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -o pipefail -o errexit -o nounset
+
+# shellcheck disable=SC2153
+PATH="$HPATH:$PATH"
```
alexeagle, line 5: What happens today if you replace PATH instead of prepend to it? Maybe add a TODO to be fully hermetic path, if you can't change it yet

oci/private/descriptor.sh
```diff
@@ -0,0 +1,23 @@
...
+path="$1"
```
alexeagle, line 6: nit: it's potentially confusing to have both `path` and `PATH`

oci/private/descriptor.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
alexeagle, line 1: include a file-level comment about what json document you're producing here? bonus points for link to a spec

oci/private/image.bzl
```diff
@@ -96,6 +97,35 @@ def _platform_str(os, arch, variant = None):
         parts["variant"] = variant
     return json.encode(parts)
 
+def _calculate_descriptor(ctx, idx, layer, zstd, jq, coreutils, regctl):
+    descriptor = ctx.actions.declare_file("%s.%s.descriptor.json" % (ctx.label.name, idx))
+    args = ctx.actions.args()
+    args.add(layer)
+    args.add(descriptor)
+    ctx.actions.run(
```
alexeagle, line 105: in theory we should be adding the windows batch wrapper around new shell script executables. I guess we can come back to it later in 2.0 release

alexeagle (approved):

oci/private/descriptor.sh
thesayyn: we need bash toolchain to be fully hermetic, which i will do at once in a follow-up.

oci/private/image.bzl
thesayyn: i can do that here, good catch.

Issue #508, opened 2024-02-21
FR: Symlink `tars` instead of copying them.  


thesayyn:
Currently, we copy them which takes long time for big files. Technically it's possible to symlink them into the output which should be instantaneous.

This wasn't possible before due to a limitation is Bazel.  with https://github.com/bazelbuild/bazel/pull/21263 is fixed, it possible to do this now.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/559

PR #559, opened 2024-05-07
feat: implement tars symlinks

thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/508, needs https://github.com/bazel-contrib/rules_oci/pull/558 first.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -9,6 +9,7 @@ module(
 bazel_dep(name = "aspect_bazel_lib", version = "2.7.2")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "0.0.8")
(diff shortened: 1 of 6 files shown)


thesayyn:
@gregmagolan is case you are interested, this is the one that will make RBE go fast.

mortenmj:
Seeing a lot of very similar errors when testing this one. At the surface, it looks like this broke all of our images that we pass tars to (built with pkg_tar in our case).

thesayyn:
That is interesting, do you have more information on this? eg which service are you using?

thesayyn:
This could be a Bazel bug!

alexeagle (approved):
Long journey to get here!

mortenmj:
> That is interesting, do you have more information on this? eg which service are you using?

We use Buildbarn. As far as I can tell, from looking at what got scheduled remotely, this never hit the build cluster at all.

Issue #436, opened 2023-12-13
`oci_image` based on `pytorch` fails with "could not parse reference"


faximan:
I am trying to build a simple "empty" image based on pytorch:

```
# WORKSPACE
oci_pull(
    name = "pytorch_base",
    image = "docker.io/pytorch/pytorch",
    tag = "1.12.1-cuda11.3-cudnn8-devel",
)

# BUILD
oci_image(
    name = "pytorch_image",
    base = "@pytorch_base",
)
```

Building `pytorch_image` fails with an error I cannot decipher:

```
$ bazel-6.3.2 build //:pytorch_image --verbose_failures
ERROR: /usr/local/home/faximan/repo//BUILD:3:10: OCI Image //:pytorch_image failed: (Exit 1): image_pytorch_image.sh failed: error executing command (from target //:pytorch_image)
  (cd /usr/local/home/faximan/.cache/bazel/_bazel_faximan/aa7e559759091793a65d16afaac928ae/execroot/repo && \
  exec env - \
  bazel-out/k8-fastbuild/bin/image_pytorch_image.sh mutate oci:layout/bazel-out/k8-fastbuild/bin/external/pytorch_base_single/layout '--output=bazel-out/k8-fastbuild/bin/pytorch_image')
# Configuration: 080d533b09bde3a5731cc76453084f9861e149e4d6a9fc0dc59f983aed67055b
# Execution platform: //bazel/rbe_config/config:platform
2023/12/13 14:28:11 storing blobs in bazel-out/k8-fastbuild/bin//storage_pytorch_image
2023/12/13 14:28:11 serving on port 34765
2023/12/13 14:28:11 GET /v2/
2023/12/13 14:28:11 HEAD /v2/oci/layout/manifests/latest 404 NAME_UNKNOWN Unknown name
2023/12/13 14:28:11 HEAD /v2/oci/layout/blobs/sha256:8889d023c18ea53dc42fff5ad1d81de2858ca7990ccffb179850543b0877c5e9 404 BLOB_UNKNOWN Unknown blob
2023/12/13 14:28:11 HEAD /v2/oci/layout/blobs/sha256:40dd5be53814ae70b2898558673b7ea18d58bf7ab3433560b9ce3cb76d9ff0b1 404 BLOB_UNKNOWN Unknown blob
[...]

Issue #365, opened 2023-09-19
Runtime of oci_image is proportional to the total size of of all tars plus the base image


wskinner-dd:
The time required to build `oci_image` seems to be proportional to the total size of the tars plus the size of the base image. This was surprising to me. I expected `oci_image` to take time proportional to the size of only the modified or added layers. After reading through the source code and running some tests, it seems like the reason for this may have something to do with [crane mutate](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_mutate.md). 

Are these runtime characteristics an intended or necessary property of `rules_oci` or the OCI spec? Or are they due to implementation details of `rules_oci` and its dependencies like `crane`?

# How to reproduce the issue
I tested this behavior by comparing the build time of four scenarios:
1. An image with a 21GB base image, one 3.6GB layer and one tiny layer.
2. An image with a 21GB base and one tiny layer.
3. An image with a 1GB base, one 3.6GB layer and one tiny layer.
4. An image with a 1GB base and one tiny layer.

For both scenarios, I built the target, then modified `app.py` and measured the time required to build again.

```
$ bazel build //:two_layer_big_image
INFO: Elapsed time: 58.190s, Critical Path: 58.13s

$ bazel build //:one_layer_big_image
INFO: Elapsed time: 21.697s, Critical Path: 21.65s

$ bazel build //:two_layer_small_image
INFO: Elapsed time: 48.928s, Critical Path: 48.88s

$ bazel build //:one_layer_small_image
INFO: Elapsed time: 1.089s, Critical Path: 1.04s 
```

# BUILD
```
pkg_tar(
    name = "app_tar",
    srcs = ["//:app.py"],
)

pkg_tar(
    name = "big_file",
    srcs = ["//:ubuntu-22.04.3-desktop-amd64.iso"],
)

oci_image(
    name = "two_layer_small_image",
    base = "@python_311",
    tars = [
        ":big_file",
        ":app_tar",
    ],
)

oci_image(
    name = "one_layer_small_image",
    base = "@python_311",
    tars = [
        ":app_tar",
    ],
)

oci_image(
    name = "two_layer_big_image",
    base = "@nct_image",
    tars = [
        ":big_file",
        ":app_tar",
    ],
)

oci_image(
    name = "one_layer_big_image",
    base = "@nct_image",
    tars = [
        ":app_tar",
    ],
)
```

wskinner-dd:
[...]

PR #560, opened 2024-05-08
feat: support symlinking blobs from base

### LINKED ISSUE #505: FR: Symlink blobs from `base` image instead of copying over.

thesayyn:
fixes #505  #436 #439 #365

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -166,6 +166,8 @@ def _oci_image_impl(ctx):
         # reuse given base image
         args.add(ctx.file.base.path, format = "--from=%s")
         inputs.append(ctx.file.base)
+        if use_symlinks:
+            transitive_inputs.append(ctx.file.base)
     else:
         # create a scratch base image with given os/arch[/variant]
         args.add(_platform_str(ctx.attr.os, ctx.attr.architecture, ctx.attr.variant), format = "--scratch=%s")
--- a/oci/private/image.sh
+++ b/oci/private/image.sh
@@ -33,11 +33,17 @@ function base_from() {
   local path="$1"
   # shellcheck disable=SC2045
   for blob in $(ls -1 -d "$path/blobs/"*/*); do
-    local relative=${blob#"$path/"}
-    mkdir -p "$OUTPUT/$(dirname "$relative")"
-    cat "$blob" >"$OUTPUT/$relative"
+    local relative_to_blobs="${blob#"$path/blobs"}"
+    mkdir -p "$OUTPUT/blobs/$(dirname "$relative_to_blobs")"
+    if [[ "$USE_TREEARTIFACT_SYMLINKS" == "1" ]]; then
+      # Relative path from `output/blobs/sha256/` to `$blob`
+      relative="$(coreutils realpath --relative-to="$OUTPUT/blobs/sha256" "$blob" --no-symlinks)"
+      coreutils ln -s "$relative" "$OUTPUT/blobs/$relative_to_blobs"
+    else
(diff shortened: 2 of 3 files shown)


oci/private/image.bzl
```diff
@@ -166,6 +166,8 @@ def _oci_image_impl(ctx):
         # reuse given base image
         args.add(ctx.file.base.path, format = "--from=%s")
         inputs.append(ctx.file.base)
```
gregmagolan, line 168: did you mean to only call `inputs.append(ctx.file.base)` if `not use_symlinks`?

oci/private/image.sh
```diff
@@ -33,11 +33,17 @@ function base_from() {
   local path="$1"
   # shellcheck disable=SC2045
   for blob in $(ls -1 -d "$path/blobs/"*/*); do
-    local relative=${blob#"$path/"}
-    mkdir -p "$OUTPUT/$(dirname "$relative")"
-    cat "$blob" >"$OUTPUT/$relative"
+    local relative_to_blobs=${blob#"$path/blobs"}
```
gregmagolan, line 36: quote?
```
local relative_to_blobs="${blob#"$path/blobs"}"
```

oci/private/image.sh
```diff
@@ -33,11 +33,17 @@ function base_from() {
...
+    mkdir -p "$OUTPUT/blobs/$(dirname "$relative_to_blobs")"
+    if [[ "$USE_TREEARTIFACT_SYMLINKS" == "1" ]]; then
+      # Relative path from `output/blobs/sha256/` to `$blob`
+      relative=$(coreutils realpath --relative-to="$OUTPUT/blobs/sha256" "$blob" --no-symlinks)
```
gregmagolan, line 40: quote?
```
relative="$(coreutils realpath --relative-to="$OUTPUT/blobs/sha256" "$blob" --no-symlinks)"
```

oci/private/image.sh
```diff
@@ -33,11 +33,17 @@ function base_from() {
...
+      coreutils ln -s "$relative" "$OUTPUT/blobs/$relative_to_blobs"
+    else
+      coreutils cat "$blob" > "$OUTPUT/blobs/$relative_to_blobs"
+    fi
   done
-  cat "$path/oci-layout" >"$OUTPUT/oci-layout"
+  coreutils cat "$path/oci-layout" >"$OUTPUT/oci-layout"
   jq '.manifests[0].annotations["org.opencontainers.image.ref.name"] = "intermediate"' "$path/index.json" >"$OUTPUT/index.json"
```
gregmagolan, line 47: jq from the PATH here? not from the `jq.jqinfo.bin,` toolchain?

oci/private/image.bzl
thesayyn: No, base has to still stay as input because we read some files from there, but when symlinks are enabled we have to also add base into transitive inputs for downstream targets, if not symlinks will be dangling.

oci/private/image.sh
thesayyn: hermetic jq is already in the PATH, added at the top of the script.

gregmagolan (approved):

PR #565, opened 2024-05-09
oci_push to support Windows builds

### LINKED ISSUE #564: oci_push fails on Windows with "%1 is not a valid Win32 application"
This rule is executable (`bazel run`) and executes in `push_app-image-push.sh.runfiles`, so we can't use `util.maybe_wrap_launcher_for_windows` as in #555, because `bash_launcher.path` used in this function is relative to the workspace directory, not the runtime directory (that would be `wrap_app-image-push.bat.runfiles`).

bartoszpop:
This pull request updates `util#_maybe_wrap_launcher_for_windows` to use relative paths to fix `bazel run` failing on Windows with `/usr/bin/bash: line 1: bazel-out/x64_windows-fastbuild/bin/some-project/push_app-image-push.sh: No such file or directory`
and update push.bzl to use `util#maybe_wrap_launcher_for_windows` (fixes #564)

### DIFF
--- a/oci/private/BUILD.bazel
+++ b/oci/private/BUILD.bazel
@@ -103,5 +103,8 @@ bzl_library(
     name = "util",
     srcs = ["util.bzl"],
     visibility = ["//oci:__subpackages__"],
-    deps = ["@bazel_skylib//lib:versions"],
+    deps = [
+        "@bazel_skylib//lib:paths",
+        "@bazel_skylib//lib:versions",
+    ],
 )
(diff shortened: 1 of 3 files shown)


thesayyn:
@alexeagle is better suited to review this change.

alexeagle (approved):
I'm not going to boot up a windows machine to verify, so I trust @bartoszpop tested this and found the windows sitation is improved.

@bartoszpop if you're willing to help a bit more with Windows - I tried really hard and got 90% of the way to turning on CI check for it: https://github.com/bazel-contrib/rules_oci/issues/97#issuecomment-1662909629
That one got wedged a bit on Zot I think, but on the 2.x branch we've removed that.

PR #567, opened 2024-05-10
fix: canonicalize missing components

### LINKED ISSUE #566: Tar symlinks cannot find layer from external repository
I was trying out the `2.x` changes for the performance improvements and ran into an issue where I cannot use a tar file layer from an external repository. The generated `image.sh` will fail to find them:
```
realpath: bazel-out/darwin_arm64-fastbuild/bin/external/_main~_repo_rules~git/data.tar: No such file or directory
```

Sample reproduction: https://github.com/brett-patterson-ent/rules_oci_repro. I have a commented out patch in there that fixes the immediate issue, but I'm not sure if it's the proper fix.

thesayyn:
In 2.x, tars stopped being direct inputs `OCIImage` actions, this meant that none of the tars passed by the user actually went into the sandbox.  fixes #566 

Imagine an oci_image in `examples/env/image` 

`bazel-out/darwin_arm64-fastbuild/bin/external/bash_amd64/data.tar.zst`  -> did fail because `external` segment does not exist, therefore can't be canonicalized.

`bazel-out/darwin_arm64-fastbuild/bin/examples/env/image/data.tar.zst` -> did NOT fail because all components in the path except the file itself existed, therefore can be canonicalized.

In the scenario above `realpath` failed if any of the directory segments in the path was missing leading to errors only when external repository labels passed to tars.  This fixes the issue by telling `realpath` to ignore missing segments, and we don't care if the file actually exists on disk. 

I changed `examples/deb` a bit so that genrule is in the external repo, providing us a test coverage for this.

### DIFF
--- a/examples/deb/BUILD.bazel
+++ b/examples/deb/BUILD.bazel
@@ -9,26 +9,13 @@ _ARCH = [
 # any other compression algorithm is simply is runtime error.
 # we decompress the tar archive with bsdtar and recompress with a
 # known compression algorithm.
-#
-# TODO: this is slow as it parses the tar archive as well
-# use zstd rule once we have one.
-[
-    genrule(
-        name = "decompress_" + architecture,
-        srcs = ["@bash_{}//:layer".format(architecture)],
-        outs = ["_{}.tar.gz".format(architecture)],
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
nice test case

Issue #562, opened 2024-05-08
Drop curl fallback for oci_pull 


thesayyn:
When we implemented oci_pull we created a fallback to curl behavior due to a missing feature in Bazel https://github.com/bazelbuild/bazel/pull/19501, now that Bazel supports `headers` feature, we no longer need this fallback and we should remove it as a breaking change in 2.x and encourage people to upgrade Bazel.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/568

PR #568, opened 2024-05-10
refactor: remove curl fallback

thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/562

BREAKING CHANGE

Now that curl fallback is gone, users hitting that code path has to do following. 

People hitting the curl fallback due to Schema 1 manifest will have to, either upgrade to `Bazel 7.1` or use a `digest` instead of a `tag` for oci_pull.

People hitting the curl fallback due to oci-media-typed images being fetched without an `Accept` header will have to, either upgrade to `Bazel 7.1`, or use a [`Credential Helper`](https://bazel.build/reference/command-line-reference#flag--credential_helper) to set headers expected by the registry, including the authorization headers.


oci/private/pull.bzl
```diff
@@ -108,17 +103,19 @@ def _download(rctx, authn, identifier, output, resource, download_fn = download.
...
-        rctx,
+    kwargs = dict(
         output = output,
         sha256 = sha256,
         url = registry_url,
         auth = {registry_url: auth},
-        headers = headers,
         allow_fail = allow_fail,
     )
+    if versions.is_at_least("7.1.0", versions.get()):
```
alexeagle, line 115: could this use `bazel_features` instead? Maybe there's no advantage

alexeagle (commented):
In the PR description you say "users in this situation need to do X, and other users in other situation need to do Y"

Can we somehow print these as error messages to make it more discoverable?

thesayyn:
> In the PR description you say "users in this situation need to do X, and other users in other situation need to do Y"

Good suggestion, some of these are not exactly detectable but we can hint those.  I'll see what can be done.

thesayyn:
> Can we somehow print these as error messages to make it more discoverable?

Okay, it's done.

thesayyn:
@alexeagle PTAL again.

.github/workflows/ci.yaml
```diff
@@ -103,6 +103,19 @@ jobs:
...
         run: echo "bzlmod_flag=--enable_bzlmod" >> $GITHUB_OUTPUT
 
+      - name: Set credential helpers flag
+        # Add --config needs_credential_helpers to add additional credential helpers
+        # to fetch from registries with HTTP headers set by credential helpers.
+        id: set_credential_helper_flag
+        if: matrix.bazelversion == '6.4.0' && matrix.folder == '.'
+        run: echo "credential_helper_flag=--config=needs_credential_helpers" >> $GITHUB_OUTPUT
+
+      - name: Setup credential helpers
```
alexeagle, line 113: is this name correct? how is setup-crane doing credential helpers?

alexeagle (approved):

.github/workflows/ci.yaml
thesayyn: Right, it could name it better, it just downloading crane to CI machine so that the credential helper can invoke it.

thesayyn: okay fixed it,

Issue #473, opened 2024-01-24
`image.sh.tpl` is too unhermetically hermetic!


GregBowyer:
I suspect this is rendered moot by https://github.com/bazel-contrib/rules_oci/pull/459, but.

We run an aggressive hermetic setup on top of NixOS, I suspect this might happen for others who are being highly hermetic.

Presently, `image.bzl` is as follows:

```python
    ctx.actions.run(
        inputs = depset(transitive = inputs_depsets),
        arguments = [args],
        outputs = [output],
        env = action_env,
        executable = util.maybe_wrap_launcher_for_windows(ctx, launcher),
        tools = [crane.crane_info.binary, registry.registry_info.launcher, registry.registry_info.registry, jq.jqinfo.bin],
        mnemonic = "OCIImage",
        progress_message = "OCI Image %{label}",
    )
```

This fails as Bazel wipes out `$PATH` and as a result, `mktemp` (and well everything) is basically none functional.

It appears that adding in the dreaded `use_default_shell_env = True` fixes this.

rickvanprim:
That should be fixed by #459.  Apologies for the slow progress.

thesayyn:
will be fixed by https://github.com/bazel-contrib/rules_oci/pull/569

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/569

PR #569, opened 2024-05-10
fix: make oci_image fully hermetic

thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/473

now oci_image has no system dependency other than bash.

### DIFF
--- a/oci/private/image.sh
+++ b/oci/private/image.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 set -o pipefail -o errexit -o nounset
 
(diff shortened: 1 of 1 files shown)


oci/private/image.sh
```diff
@@ -2,7 +2,8 @@
 set -o pipefail -o errexit -o nounset
 
 # Add hermetically built jq, regctl, coreutils into PATH.
```
alexeagle, line 4: Nit: add -> replace

alexeagle (approved):

Issue #35, opened 2022-09-26
FR: Build Dockerfiles with incremental cache support


stabai:
Feature request:
Allow Bazel to build Dockerfile images in a way that takes advantage of incremental build caching. The `dockerfile_image` rule in rules_docker contrib allows building such images, but it's incredibly slow because of the lack of build cache.

cgrindel:
Thanks for your proposal. @thesayyn has been thinking about this, as well. He will comment on this issue with his thoughts. Hopefully, we can come up with a solution that satisfies your goals.

thesayyn:
There are some problems that'll prevent us from doing this the way you want it;

- bazel has its own cache so anything that falls out of it is not tracked by bazel. therefore it would not be trivial to get the docker cache included in the bazel cache. 
- an alternative is to keep the docker build cache local to the host and let it leak into the sandbox. but this would increase the likelihood of non-hermetic builds.
- biggest drawback is that `Dockerfile` is not hermetic by design, meaning instructions such as `RUN apt-get install curl` or `make build` is not repeatable, which would lead to non-reproducible builds. 
- bazel already knows how to [cross-]compile various languages, so it doesn't make sense to delegate this task to docker.
- requires rules_oci to declare a dependency on a docker_toolchain which we are not willing to take.

I have been thinking about this issue for a while, how to tackle the problem without sacrificing a whole lot. combination of `gazelle + rules_oci` is the way to do this. to explain the idea roughly;

given a Dockerfile; 

```Dockerfile
FROM node:12

RUN apt-install curl

WORKDIR /app

COPY index.js  .

RUN yarn install --production

CMD ["node", "index.js"]

EXPOSE 3000
```
a gazelle plugin for rules_oci would translate it to;

WORKSPACE
```starlark

debian_archive(
    name = "debian_amd64_curl"
    urls = ["debian_archive_url"]
)

oci_pull(
    name = "node_12",
    repository = "index.docker.io/library/node",
    ref = "12"
)

translate_pnpm_lock(
   name = "npm",
   yarn_lock = "yarn.lock"
)
```

BUILD
```starlark

js_binary(
   name = "app",
   entry_point = "index.js"
)
oci_image(
   name = "image",
   base = "@node_12//:image", 
   tars = [
      ":app"
   ]
   workdir = "/app",
   cmd = ["index.js"],
   ports = [3000],
   
)
```

mvgijssel:
[...]

Issue #132, opened 2023-03-29
FR: Run Bazel actions inside a container


ding-ma:
Similar feature to `container_run_and_commit`  in rules_docker

[See](https://github.com/bazelbuild/rules_docker/blob/master/docker/util/README.md#container_run_and_commit)

thesayyn:
fundamentally this is a bad idea. see why: https://github.com/bazel-contrib/rules_oci/issues/35#issuecomment-1285954483

alexeagle:
I think it's possible that there's a better way to containerize actions, we should give it some thought. But as we wrote in the main README, the goal here is to prevent unmaintainable design mistakes that we learned from rules_docker.

Also Bazel has its own way of providing isolation for build actions, so it's not clear to me why we would implement another one in rules_oci. I'd like to see some experiments with https://bazel.build/reference/command-line-reference#flag--experimental_enable_docker_sandbox (and https://bazel.build/reference/command-line-reference#flag--experimental_docker_image).

thesayyn:
I dug a bit into what can be done; this looks possible with the docker strategy. with the https://github.com/GoogleContainerTools/kaniko combination, it should work

ding-ma:
I'll experiment with `--experimental_docker_image` and report back findings.

alexeagle:
I think we should slip this from 1.0 already, since we'd like to get an RC out this week, and it's still a science project.

aw185176:
>Also Bazel has its own way of providing isolation for build actions, so it's not clear to me why we would implement another one in rules_oci

One use case is pre-generating a .bazelrc for remote caching that has a cache key prefix based on the system dependencies that aren't tracked by Bazel. Such a task needs to be ran in the container to be accurate. I do agree that generally speaking such a thing should not really be necessary.

alexeagle:
@aw185176 I don't think that sounds like a containerization task. If you want Bazel flags to be dynamic based on the system, then a `/tools/bazel` wrapper is the way to set those flags.

aw185176:
> @aw185176 I don't think that sounds like a containerization task. If you want Bazel flags to be dynamic based on the system, then a `/tools/bazel` wrapper is the way to set those flags.

[...]

PR #570, opened 2024-05-11
example: add dockerfile build

thesayyn:
This is nice example demonstrating a `Dockerfile` built by Bazel used as a base for oci_image.

closes https://github.com/bazel-contrib/rules_oci/issues/35

See https://github.com/bazel-contrib/rules_oci/issues/132


examples/dockerfile/BUILD.bazel
```diff
@@ -0,0 +1,65 @@
...
+
+# Convert tarball to an oci_image base
+genrule(
+    name = "regctl_gen",
+    outs = ["regctl.sh"],
+    cmd = "echo '$(REGCTL_BIN) $$@' >> $@",
+    toolchains = ["@oci_regctl_toolchains//:current_toolchain"],
+)
+
+run_binary(
```
thesayyn, line 37: i wish run_binary supported toolchains attribute like genrule did. this is an unfortunate workaround which makes this example a  bit confusing.

.github/workflows/ci.yaml
```diff
@@ -141,6 +141,10 @@ jobs:
               ],
           )
           EOF
+
+      - name: Setup buildx
```
alexeagle, line 145: I don't understand why this is needed. The genrule runs the same thing as part of the build, right?

thesayyn: No, this creates a new driver in local daemon to allow `--builder=container` argument in genrule.  And we need a driver that supports exporting an oci-layout, instead of requiring us to call docker save to get the built image off of the daemon.

alexeagle: okay so needs a comment then. Also the `||true` is suprpsing, it's okay for this to fail?

alexeagle (approved):

.github/workflows/ci.yaml
thesayyn: || true might not be necessary. removing.

thesayyn: added the comment.

Issue #572, opened 2024-05-11
Adding bazel tag support to oci_pull (import_tags) because an oci_pull for a large image does not work when run on remote execution


steve-261370:
I'm migrating from rules_docker's `container_pull` to `oci_pull` and the [current migration instructions](https://docs.aspect.build/guides/rules_oci_migration/#container_pull) say that the old `import_tags` attribute:

> `import_tags` is unsupported. no plans to add support for it.

We were using the `import_tags` to set the tag `no-remote` as we noticed one of the large images we pulled did not work on remote execution.  The image is over 70GB.  The build failure we get for that large image is:

> ERROR: /home/steve-261370/.cache/bazel/_bazel_steve-261370/2914a5da309afa1aa83708b7e78fb71f/external/large_image_single/BUILD.bazel:16:18: Copying files to directory large_image_single/blobs/sha256 failed: (Exit 34): Error while uploading artifact with digest '0ddf758e26952be880786328ac7091616623a2577ed9f5d74dc4a412becde8df/75284488761'

Is there a reason for there being "no plans to add support for it"?  I ended up patching in support for it (which allows the build for my large image to work remotely).  I'll send in the patch I applied shortly to add support for it.

thesayyn:
This is a good use case, I'll think about how we could support this. the current workaround would be setting `--modify_execution_requirements=CopyToDirectory=...` https://bazel.build/reference/command-line-reference#flag--modify_execution_info

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/573

PR #573, opened 2024-05-11
Add import_tags attribute to oci_pull

steve-261370:
Adds an import_tags attribute to the oci_pull macro to allow passing Bazel tags to the underlying rules. This implementation follows the naming and general implementation of the import_tags attribute on the old legacy container_pull rule.

#572

### DIFF
--- a/oci/pull.bzl
+++ b/oci/pull.bzl
@@ -103,7 +103,7 @@ _PLATFORM_TO_BAZEL_CPU = {
     "linux/mips64le": "@platforms//cpu:mips64",
 }
 
-def oci_pull(name, image = None, repository = None, registry = None, platforms = None, digest = None, tag = None, reproducible = True, is_bzlmod = False, config = None, config_path = None, import_tags = []):
+def oci_pull(name, image = None, repository = None, registry = None, platforms = None, digest = None, tag = None, reproducible = True, is_bzlmod = False, config = None, config_path = None, bazel_tags = []):
     """Repository macro to fetch image manifest data from a remote docker registry.
 
     To use the resulting image, you can use the `@wkspc` shorthand label, for example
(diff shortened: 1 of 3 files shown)


thesayyn:
the name import_tags is really confusing here.  it's gonna be confused for Docker tags.

thesayyn:
And i belive the the `import` is used here because rules_docker used a rule called `container_import` under the hood.

steve-261370:
ahh - good point.  I made a change to rename `import_tags` to `bazel_tags`.  If you had another suggestion on the name, let me know.  Or if you come up with a better way to handle this.

oci/pull.bzl
```diff
@@ -225,4 +228,5 @@ def oci_pull(name, image = None, repository = None, registry = None, platforms =
         config = config,
         # TODO(2.0): remove
         config_path = config_path,
+        tags = bazel_tags,
```
thesayyn, line 231: could you also rename the attribute to be `bazel_tags` instead of just tags ?

steve-261370: I'm not following.  The `oci_alias` here is a repository rule, and all rules have the common attribute `tags`.  I wasn't aware that you could rename or alias existing common attributes.  Feel free to modify/make changes to my pull request if there was a way to do that.

thesayyn: by default, repository rules only have a single `name` attribute, `tags` is common for build rules. `tags` here in the repository rule context is confusing, it's https://bazel.build/reference/be/common-definitions#typical-attributes 

you can just add a new attribute called `bazel_tags`.

steve-261370: ahh - thank you for the clarification.  In the next commit, I've renamed it to `bazel_tags` and added the attribute.

However, I did remove adding the attribute to `oci_alias` because it wasn't being used.  Let me know if you'd still like me to add it to `oci_alias`

thesayyn (approved):

oci/pull.bzl
steve-261370: Resolving this conversation - I see a thumbs up on the last comment and the changes have been approved.

PR #574, opened 2024-05-13
Exclude the use of xattrs

FlynnB17:
Error was seen in multiple environments all Mac OS 14.3.1
lsetxattr com.apple.provenance /blobs/sha256/something.tar.gz: operation not supported

After making this change the issue is not observed.


thesayyn:
Thanks for the PR. https://github.com/bazel-contrib/rules_oci/pull/385 is landed already on 2.x which fixes this already.

FlynnB17:
> Thanks for the PR. #385 is landed already on 2.x which fixes this already.

Sweet thanks will follow that and close this

PR #579, opened 2024-05-16
fix: don't include mac metadata

alexeagle:
Partial back-port of #385 from the 2.x branch


thesayyn (approved):
LGTM, hopefully bsdtar and gnutar both have these options.

alexeagle:
GNU `tar` doesn't have `--no-mac-metadata` so I removed from this PR. waiting on Sean to find out if that's enough to solve the problem.

alexeagle:
This is strictly improvement so I'm going to merge even if it doesn't resolve an issue.

Issue #529, opened 2024-03-19
FR: Return module_ctx.extension_metadata for bazel mod tidy


nobu-k:
`bazel mod tidy` can automatically update `use_repo(oci)` if the extension returns pulled image repository names using `root_module_direct_deps` of `module_ctx.extension_metadata`.

The same approach as Gazelle should work with older versions of Bazel.

https://github.com/bazelbuild/bazel-gazelle/blob/master/internal/bzlmod/go_deps.bzl

alexeagle:
Ooh that sounds like a great feature, I guess it's similar to how renovate updates MODULE.bazel?

thesayyn:
This is currently broken due to root module having use_repo calls such as `oci_crane_toolchains`

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/611

PR #580, opened 2024-05-19
Return extension_metadata from oci.pull()

mortenmj:
Return extension_metadata from oci.pull(), making `bazel mod tidy` automagically populate `use_repo(oci, ...)` in MODULE.bazel

Fixes https://github.com/bazel-contrib/rules_oci/issues/529


thesayyn (commented):
Could you add some comments?

alexeagle (approved):

Issue #129, opened 2023-03-29
FR: oci_pull support identity tokens


thesayyn:
credHelper response with `{"Username": "<token>"}` should be handled differently. 

See https://github.com/moby/moby/issues/36926

wingsofovnia:
Any updates? Is there any priority assigned to this issue or perhaps there are blockers?

thesayyn:
No, PRs are welcome.

BradHolmes:
This small change in `authn.bzl` got things working for me.  Anything else needed?  

```
     if response["Username"] == "<token>":
-        fail("Identity tokens are not supported at the moment. See: https://github.com/bazel-contrib/rules_oci/issues/129")
+        return {
+            "type": "pattern",
+            "pattern": "Bearer <password>",
+            "password": response["Secret"],
+        }
```

I don't really have a good idea about adding an example or test though.

wingsofovnia:
Should be fixed by https://github.com/bazel-contrib/rules_oci/pull/626

thesayyn:
Closing as fixed.

PR #581, opened 2024-05-20
add OAuth2 support

xbog300:
As per https://github.com/bazel-contrib/rules_oci/issues/129, I implemented `OAuth2` support for `rules_oci`.

A little bit of a background story of the proposed solution:

`OAuth2` requires a `POST` request to be sent. This, unfortunately, cannot be done by the `repository_ctx.download` func.
**Bazel does not provide a way to send a POST request.**

In order to enable `OAuth2` support, a tiny `python` script was added. The entire purpose of the script is to do the `POST` request. In order to do that, a `requests` package was used. 

The support is optional, meaning it's only enabled if `SUPPORT_OAUTH2` env var is being set (appropriate information about the lack of var set is being displayed).

Also, the change I made verifies that the environment in which the rule is being executed, has `python3` binary available, and if the python installed in the environment has `requests` package available, too.

**What are the options?**

It's clear the `POST` request can't be handled by the Bazel itself, meaning we must look for other ways of sending it.
Python is one option, other one is `curl / wget` on `Linux` and `PowerShell` on `Windows`.

Please let me know what are your thoughts on the solution I proposed, to engage `python` in order to enable `OAuth2` support.

### DIFF
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -124,14 +124,69 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
 
     response = json.decode(result.stdout)
 
-    if response["Username"] == "<token>":
-        fail("Identity tokens are not supported at the moment. See: https://github.com/bazel-contrib/rules_oci/issues/129")
-
     return {
         "type": "basic",
         "login": response["Username"],
         "password": response["Secret"],
     }
+
+def _oauth2(rctx, realm, scope, service, secret):
+    if not rctx.os.environ.get("SUPPORT_OAUTH2", False):
+        fail("oauth2 support is disabled (set SUPPORT_OAUTH2 to True)")
+
+    python_bin = "python3"
+
+    # oauth2 requires python to be available in the env
+    if not rctx.which(python_bin):
+        fail("oauth2 failed: to use oauth2, make sure %s is available" % python_bin)
+
+    # oauth2 requires "requests" package installed alongside python
+    result = rctx.execute([python_bin, "-c", "import requests"])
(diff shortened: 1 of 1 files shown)


thesayyn:
Thanks for the PR,  there's a lot to be discussed here, I'll do my best reply again this week.

xbog300:
ping @thesayyn

thesayyn:
> In order to enable `OAuth2` support, a tiny `python` script was added. The entire purpose of the script is to do the `POST` request. In order to do that, a `requests` package was used.

Okay, this sounds bad, python dependency is a "no-go" for us. 

> other one is `curl / wget` on `Linux` and `PowerShell` on `Windows`.

This sounds like a better alternative than using a python script.

xbog300:
> Okay, this sounds bad, python dependency is a "no-go" for us.

Affirm, I used python just to create a concept of a potential solution and to prove it works.

Will do the changes.

For `linux`, I think it'd be great to try use `curl` first, and if that's not available in the path (which might be the case), then treat `wget` as a last resort.

For `Windows`, I'll just go with `PowerShell`, which is there by default AFAICS.

lukaszwawrzyk:
@thesayyn @xbog300 I tried to address the comments in https://github.com/bazel-contrib/rules_oci/pull/626

thesayyn:
@xbog300 looks like [lukaszwawrzyk](https://github.com/lukaszwawrzyk) picked up your PR, do you want to close yours as i already left bunch of comments at the other PR?

thesayyn:
closing in favor of https://github.com/bazel-contrib/rules_oci/pull/626

Issue #432, opened 2023-12-12
FR: download layers in parallel


thesayyn:
Bazel 7 introduced a feature where you can download in parallel in the same repository rule context. We should feature detect and use that.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/601

PR #582, opened 2024-05-20
Use parallel download on Bazel >7.1.0

mortenmj:
On Bazel 7.1.0 and later, `rctx.download()` takes a `block` argument. When `block` is False, download() the return value has a `wait()` method which returns the regular `download()` return value.

This PR changes the `_download()` function in pull.bzl to operate on lists of identifiers/outputs, so that we can handle this complexity internally in the function, and downloads asynchronously where supported.

Fixes https://github.com/bazel-contrib/rules_oci/issues/432

### DIFF
--- a/oci/private/pull.bzl
+++ b/oci/private/pull.bzl
@@ -80,7 +80,7 @@ def _digest_into_blob_path(digest):
     digest_path = digest.replace(":", "/", 1)
     return "blobs/{}".format(digest_path)
 
-def _download(rctx, authn, identifier, output, resource, headers = {}, allow_fail = False):
+def _download(rctx, authn, identifier, output, resource, headers = {}, allow_fail = False, block = True):
     "Use the Bazel Downloader to fetch from the remote registry"
 
     if resource != "blobs" and resource != "manifests":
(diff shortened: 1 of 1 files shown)


mortenmj:
cc: @thesayyn

oci/private/pull.bzl
```diff
@@ -80,42 +80,63 @@ def _digest_into_blob_path(digest):
     digest_path = digest.replace(":", "/", 1)
     return "blobs/{}".format(digest_path)
 
-def _download(rctx, authn, identifier, output, resource, headers = {}, allow_fail = False):
+def _download(rctx, authn, identifiers, outputs, resource, headers = {}, allow_fail = False):
```
thesayyn, line 83: i don't think the place to implement this, and i don't think we care about downloading manifests and small blobs like config concurrently. 

Can we just make download accept additional **kwargs and download only layers in parallel in the for loop below?

mortenmj: Which kwargs would you have it accept?

thesayyn: accepting an additional `block` argument and applying it conditionally the same way as the headers argument would be fine here. 

then down below where we download layers sequentially can pass `block` argument unconditionally and collect results and do `[r.wait() for r in results if "wait" in r]`. that would be cleanest implementation imho.

thesayyn (approved):
Awesome. Thank you!

thesayyn:
@mortenmj do you care if we land this to main or 2.x?

mortenmj:
> @mortenmj do you care if we land this to main or 2.x?

No

thesayyn:
Okay, i'll get this landed after CURL one.

thesayyn:
i didn't close this???? @mortenmj could you reopen a new one?

PR #590, opened 2024-05-24
Use a transition to resolve the correct versions of crane and yq

EdSchouten:
Bazel doesn't provide a very elegant way to specify that a toolchain should be resolved in such a way that it can be executed on a *target* platform:

https://github.com/bazelbuild/bazel/issues/19645

To work around this, we can pass in the toolchains through ordinary labels and apply an outgoing edge transition to set --extra_execution_platforms equal to --platforms. This causes the toolchains to be resolved the way we want.

Though I think using transitions for this is pretty slick, it's a shame that this has to be done by patching up command line options.

### DIFF
--- a/oci/private/push.bzl
+++ b/oci/private/push.bzl
@@ -98,6 +98,23 @@ oci_push(
 ```
 """
 
+# Helper rule for ensuring that the crane and yq toolchains are actually
+# resolved for the architecture we are targeting.
+def _transition_to_target_impl(settings, attr):
+    return {
+        # String conversion is needed to prevent a crash with Bazel 6.x.
+        "//command_line_option:extra_execution_platforms": [
+            str(platform)
+            for platform in settings["//command_line_option:platforms"]
+        ],
+    }
+
+_transition_to_target = transition(
+    implementation = _transition_to_target_impl,
+    inputs = ["//command_line_option:platforms"],
+    outputs = ["//command_line_option:extra_execution_platforms"],
+)
+
 _attrs = {
     "image": attr.label(
         allow_single_file = True,
@@ -124,19 +141,30 @@ _attrs = {
         """,
         allow_single_file = [".txt"],
     ),
+    "_allowlist_function_transition": attr.label(
+        default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+    ),
+    "_crane": attr.label(
+        cfg = _transition_to_target,
+        default = "@oci_crane_toolchains//:current_toolchain",
+    ),
     "_push_sh_tpl": attr.label(
         default = "push.sh.tpl",
         allow_single_file = True,
     ),
     "_windows_constraint": attr.label(default = "@platforms//os:windows"),
+    "_yq": attr.label(
+        cfg = _transition_to_target,
+        default = "@yq_toolchains//:resolved_toolchain",
+    ),
 }
 
 def _quote_args(args):
     return ["\"{}\"".format(arg) for arg in args]
 
 def _impl(ctx):
-    crane = ctx.toolchains["@rules_oci//oci:crane_toolchain_type"]
-    yq = ctx.toolchains["@aspect_bazel_lib//lib:yq_toolchain_type"]
+    crane = ctx.attr._crane[0][platform_common.ToolchainInfo]
+    yq = ctx.attr._yq[0][platform_common.ToolchainInfo]
 
     if ctx.attr.repository and ctx.attr.repository_file:
         fail("must specify exactly one of 'repository_file' or 'repository'")
@@ -184,11 +212,7 @@ def _impl(ctx):
 oci_push_lib = struct(
     implementation = _impl,
     attrs = _attrs,
-    toolchains = [
-        "@rules_oci//oci:crane_toolchain_type",
-        "@aspect_bazel_lib//lib:yq_toolchain_type",
-        "@bazel_tools//tools/sh:toolchain_type",
-    ],
+    toolchains = ["@bazel_tools//tools/sh:toolchain_type"],
 )
 
 oci_push = rule(
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,7 +17,7 @@ use_repo(oci, "oci_crane_registry_toolchains", "oci_crane_toolchains")
 register_toolchains("@oci_crane_toolchains//:all", "@oci_crane_registry_toolchains//:all")
 
 # Dev dependencies
-bazel_lib = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "toolchains", dev_dependency = True)
+bazel_lib = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "toolchains")
 bazel_lib.jq()
 bazel_lib.yq()
 use_repo(bazel_lib, "jq_toolchains", "yq_toolchains")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -19,7 +19,8 @@ register_toolchains("@oci_crane_toolchains//:all", "@oci_crane_registry_toolchai
 # Dev dependencies
 bazel_lib = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "toolchains", dev_dependency = True)
 bazel_lib.jq()
-use_repo(bazel_lib, "jq_toolchains")
+bazel_lib.yq()
+use_repo(bazel_lib, "jq_toolchains", "yq_toolchains")
 
 bazel_dep(name = "rules_go", version = "0.46.0", dev_dependency = True, repo_name = "io_bazel_rules_go")
 bazel_dep(name = "rules_pkg", version = "0.7.0", dev_dependency = True)


anguslees:
> I have been told that the exec_groups solve this problem, any way we could use that here?

Unfortunately, exec_groups don't support transitions afaics.  `exec_group(exec_compatible_with=HOST_CONSTRAINTS)` is my current local-hack work-around for bazel-run commands like oci_push (which we can assume are going to run on host).  Ed's suggestion is more correct in general - and more correct for host constraints beyond simple os/cpu - and I will be moving my own patches to that.

oci/private/push.bzl
```diff
@@ -128,15 +141,19 @@ _attrs = {
         default = "push.sh.tpl",
         allow_single_file = True,
     ),
+    "_tools": attr.label(
+        cfg = _transition_to_target,
+        default = "oci_push_tools",
+    ),
```
anguslees, line 147: Minor: Fwiw, I think we can use the existing `:current_toolchain` targets here, and avoid the new `oci_push_tools` target.  I think the result is a bit more understandable, at the cost of having to declare crane and yq separately:

ie:
```starlark
  "_crane": attr.label(
      cfg = _transition_to_target,
      default = "@oci_crane_toolchains//:current_toolchain",
  ),
```

... and then refer to it in `_impl()` below using: `crane = ctx.attr._crane[0][platform_common.ToolchainInfo]`

And similar for yq.

oci/private/push.bzl
```diff
@@ -98,6 +98,19 @@ oci_push(
 ```
 """
 
+# Helper rule for ensuring that the crane and yq toolchains are actually
+# resolved for the architecture we are targeting.
+def _transition_to_target_impl(settings, attr):
+    return {
+        "//command_line_option:extra_execution_platforms": settings["//command_line_option:platforms"],
```
anguslees, line 105: This didn't work for me on bazel 6.4.0, throwing an error like:

```
FATAL: bazel crashed due to an internal error. Printing stack trace:
java.lang.RuntimeException: Unrecoverable error while evaluating node 'com.google.devtools.build.lib.skyframe.RegisteredExecutionPlatformsValue$Key@98ab941b' (requested by nodes 'ToolchainContextKey{configurationKey=BuildConfigurationKey[6e2757b80890975f9114c79f8e56d8ad17b03e4d5656996857adff85802c3252], toolchainTypes=[ToolchainTypeRequirement{toolchainType=@rules_oci//oci:crane_toolchain_type, mandatory=true}], execConstraintLabels=[], forceExecutionPlatform=Optional.empty, debugTarget=false}')
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:633)
	at com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:365)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool.scan(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(Unknown Source)
Caused by: java.lang.ClassCastException: class com.google.devtools.build.lib.cmdline.Label cannot be cast to class java.lang.String (com.google.devtools.build.lib.cmdline.Label is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
	at com.google.devtools.build.lib.skyframe.TargetPatternUtil.parseAllSigned(TargetPatternUtil.java:83)
	at com.google.devtools.build.lib.skyframe.RegisteredExecutionPlatformsFunction.compute(RegisteredExecutionPlatformsFunction.java:92)
	at com.google.devtools.build.skyframe.AbstractParallelEvaluator$Evaluate.run(AbstractParallelEvaluator.java:562)
	... 7 more
```

I had to force it to a string explicitly:
```suggestion
        "//command_line_option:extra_execution_platforms": str(settings["//command_line_option:platforms"][0]),
```

oci/private/push.bzl
EdSchouten: Great! I wasn't aware of these targets.

oci/private/push.bzl
EdSchouten: Done, though I did add a loop to iterate over the platforms.

EdSchouten:
PTAL!

oci/private/push.bzl
```diff
@@ -124,19 +141,27 @@ _attrs = {
         """,
         allow_single_file = [".txt"],
     ),
+    "_crane": attr.label(
+        cfg = _transition_to_target,
```
thesayyn, line 145: Shouldn't this simply be `cfg = "exec"`?  What's the reason to transition oci_push target platform?

EdSchouten: Using cfg = “exec” would give you binaries suitable for use in ctx.actions.run(). But that doesn’t necessarily give you binaries that can be launched on the target, which is the platform for which oci_push() builds its launcher script.

thesayyn: Hmm, perhaps i'm wrong, doesn't bazel use `exec` (aka host) configuration when running `bazel run :target`?

EdSchouten: No. That only holds if the first platform you register happens to be equal to your host platform.

If you run Bazel on a Mac, and you register Linux followed by macOS, then Bazel will try to execute as many actions as possible on Linux.

EdSchouten:
Whoops! Forgot to add `_allowlist_function_transition` for Bazel 6 support. PTAL!

EdSchouten:
Hmmm... In the bzlmod case we're unable to access `@yq_toolchains`. What would be the most logical way to address this? Request that `@aspect_bazel_lib` provides an `alias()` for the resolved toolchain?

thesayyn:
Something along the lines of https://github.com/bazel-contrib/rules_oci/blob/6620d469c84c64a21a2d42213204b3ad4c8c5549/MODULE.bazel#L27-L30

oci/private/push.bzl
anguslees: ^that.  Eg, when using bazel on mac with a linux RBE cluster (ie: `bazel run :push --platforms=mac --extra_execution_platforms=linux`), the `expand_template` action will have target=mac and exec=linux.  We need to find the crane/yq binaries _for mac_, because we're using their paths in our _output_ and not actually invoking these tools as part of the expand_template action itself.

alexeagle (approved):
fancy!

alexeagle:
This doesn't cherry-pick onto 2.x so I think we'll need to take that upstream fix for our 2.0.0 release anyway. thanks for reminding us @BoleynSu

EdSchouten:
Again, I don’t see a reason why we should declare all toolchains twice, purely to distinguish in what context they end up getting used.

EdSchouten:
Are you sure that's correct...? Given your example, we're only forcing that a toolchain is resolved with both target & exec set to Windows. That doesn't necessarily imply that in order to build dependencies of that toolchain we also apply such logic.

EdSchouten:
> A repro it pretty easy to make.

Feel free to post one.

EdSchouten:
In my case it failed because it first failed as follows:

```
ERROR: /Users/ed/foo/BUILD.bazel:4:8: Executing genrule //:main_go [for tool] failed: (Exit 1): bash.exe failed: error executing Genrule command (from target //:main_go) c:/tools/msys64/usr/bin/bash.exe -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; touch "bazel-out/darwin_arm64-opt-exec-ST-6b9de8e24f96/bin/main.go"'

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
src/main/tools/process-wrapper-legacy.cc:80: "execvp(c:/tools/msys64/usr/bin/bash.exe, ...)": No such file or directory
Target //:image failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 6.134s, Critical Path: 0.11s
INFO: 4 processes: 4 internal.
ERROR: Build did NOT complete successfully
```

That's because it tried to run the `genrule()` on Windows. Adding an `exec_compatible_with = ["@platforms//os:macos"]` to that sorted that out.  Then it failed like this:

```
ERROR: /private/var/tmp/_bazel_ed/0ba728bada58c95ae54c22d2bde4e2e8/external/rules_go~~go_sdk~rules_go__download_0_windows_amd64/BUILD.bazel:62:15: GoToolchainBinaryBuild external/rules_go~~go_sdk~rules_go__download_0_windows_amd64/builder.exe [for tool] failed: (Exit 2): builder.exe.bat failed: error executing GoToolchainBinaryBuild command (from target @@rules_go~~go_sdk~rules_go__download_0_windows_amd64//:builder) bazel-out/darwin_arm64-opt-exec-ST-6b9de8e24f96/bin/external/rules_go~~go_sdk~rules_go__download_0_windows_amd64/builder.exe.bat

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
bazel-out/darwin_arm64-opt-exec-ST-6b9de8e24f96/bin/external/rules_go~~go_sdk~rules_go__download_0_windows_amd64/builder.exe.bat: line 1: @echo: command not found
bazel-out/darwin_arm64-opt-exec-ST-6b9de8e24f96/bin/external/rules_go~~go_sdk~rules_go__download_0_windows_amd64/builder.exe.bat: line 8: unexpected EOF while looking for matching `"'
bazel-out/darwin_arm64-opt-exec-ST-6b9de8e24f96/bin/external/rules_go~~go_sdk~rules_go__download_0_windows_amd64/builder.exe.bat: line 10: syntax error: unexpected end of file
Target //:image failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.591s, Critical Path: 0.38s
INFO: 2 processes: 2 internal.
ERROR: Build did NOT complete successfully
```

That I could solve by constraining the registration of Go toolchains, by not registering ones where `host_goos == "windows"`. So in that sense you can get it to work, even with a transition like that in place.

That said, I agree that this is suboptimal. But it's still not as awful as 'runtime toolchains'. We should simply work with the Bazel developers to get the right set of features added to Bazel to do resolution of toolchains the way we want.

anguslees:
How do we do this with one toolchain type?  How will resolution know if we need it for the exec or target platform (when those are not the same)?

I agree the exec platform transition propagates through the toolchain transition, and this breaks later actions.  Sad face.

I think we're back to "no good solution" here.  Duplicate toolchains works but isn't "good" (duplicates work for everyone and confuses target vs exec in definition).  Exec platform transition breaks deeper builds, and would require further work I'm bazel to prevent that (not clear exactly what, since we've lost information in the transition).

manuelnaranjo:
@EdSchouten @thesayyn @alexeagle folks be mindful of #706 this is breaking cross cpu compilation, not sure if anyone else is looking into a fix, looks like when crosscompiling from linux/x64 to linux/arm64 it also happens, so it's trivial to add a unit test to validate that things are broken, and then fix from there,  I may have some capacity this week, but I can't promise much, but if someone else has capacity feel free to tackle it

thesayyn:
Okay, i think this workaround has created problems as much as it solved, i am not sure what to do here.

anguslees:
> Okay, i think this workaround has created problems as much as it solved, i am not sure what to do here.

The only current solution without problems (that I'm aware of) are:

1. Don't use toolchains here, and instead use a regular cfg=target label that (for example) points at a select() block to choose amongst pre-built binaries.  _or_
2. Introduce new crane_target and yq_target toolchains, that are based around target_compatible_with rather than exec_compatible_with constraints.

PR #595, opened 2024-05-29
Copy files using coreutils cp

plobsing:
Unlike `cat` redirected to a file, `cp` sees both its input and output and can more readily know the details of its arguments. These details can sometimes be exploited to achieve zero-copy transfers when the OS and filesystem know how to do copy-on-write (e.g. `clonefile` on Mac, `copy_file_range` on Linux). uutils' `cp` will attempt to take advantage of these opportunities where possible (see: https://github.com/uutils/coreutils/blob/35232689366fb102854693b6bd6b6161901d192e/src/uu/cp/src/cp.rs#L2190 ).

We need to pass `--no-preserve=mode` in order for `cp` to create the files as writable; which is necessary as some of the copies we perform need to be mutated afterwards. Depending on the nature of the Bazel setup, action input files might be readable but not writable, so the default behaviour of `cp` to preserve the permissions from its input to its output in this case would cause our later mutations to fail.

Also convert one of the remaining references to the system's `cat` (rather than uutils') in order to also convert it to a call to `cp`.

### DIFF
--- a/oci/private/image.sh
+++ b/oci/private/image.sh
@@ -41,10 +41,10 @@ function base_from() {
       relative="$(coreutils realpath --relative-to="$OUTPUT/blobs/sha256" "$blob" --no-symlinks)"
       coreutils ln -s "$relative" "$OUTPUT/blobs/$relative_to_blobs"
     else
-      coreutils cat "$blob" > "$OUTPUT/blobs/$relative_to_blobs"
+      coreutils cp "$blob" "$OUTPUT/blobs/$relative_to_blobs"
     fi
   done
-  coreutils cat "$path/oci-layout" >"$OUTPUT/oci-layout"
+  coreutils cp "$path/oci-layout" "$OUTPUT/oci-layout"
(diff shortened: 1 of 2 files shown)


thesayyn:
Thanks for the detailed write up. I never got cp working with Bazel permissions, mode.

plobsing:
> Thanks for the detailed write up. I never got cp working with Bazel permissions, mode.

Looks like I failed to get it working too. It seems I overlooked the fact that `oci-layout` needs to be writable in order to accept later updates. Got it working locally (MacOS) now.

thesayyn:
My nightmares came alive again, bonus points if you could get it working. :)

plobsing:
OK, despite my best efforts, I couldn't get this latest instance to repro locally, but I did manage to run the workflow in my fork to get a repro. And based on that, it seems that if we pass `--no-preserve=mode` consistently everywhere, it works. PTAL.

thesayyn (approved):

PR #598, opened 2024-05-29
Support passing headers to download if Bazel supports it

ddeville:
Rather than falling back to curl let's actually pass the headers to the native Bazel downloader if we're running a version of Bazel that supports them (i.e. >= 7.1.0). I had to bring the `bazel_features` repo to detect this.

Note that I've kept support for curl rather than introducing the breaking change (as #562 discusses) so that we can start taking advantage of it without breaking for people still on older versions of Bazel.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,6 +7,7 @@ module(
 )
 
 bazel_dep(name = "aspect_bazel_lib", version = "1.42.1")
+bazel_dep(name = "bazel_features", version = "1.11.0")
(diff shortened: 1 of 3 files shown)


thesayyn:
This is already landed on 2.x via https://github.com/bazel-contrib/rules_oci/pull/528 

See https://github.com/bazel-contrib/rules_oci/blob/2.x/oci/private/pull.bzl

ddeville:
ah oops I didn't realize there was a 2.x branch...

PR #599, opened 2024-05-29
   fix: add tars into action inputs

thesayyn:
In 2.0, we promised that `tars` were no longer directly inputs to oci_image, but it was never possible to do it because of subtle ways of how bazel works. Even though it worked with most execution strategies, it did not work with RBE and BwoB because of how bazel calculates hashes for treeartifacts. 

Discussion here: https://github.com/bazelbuild/bazel/issues/22504

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -157,7 +157,10 @@ def _oci_image_impl(ctx):
         },
     )
 
-    inputs = [builder]
+    # Unfortunately even though tars are not needed directly in the sandbox we have
+    # to add it to the inputs so bazel can calculate the hash for the output treeartifact.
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
bummer but thanks for clear communication on that upstream issue.

alexeagle:
I guess we don't have a way to add the missing test that would have been red? Should we add RBE to rules_oci nowish?

thesayyn:
> I guess we don't have a way to add the missing test that would have been red? Should we add RBE to rules_oci nowish?

If we have some free tier RBE access yes that would give us more coverage. It would prevent a lot of accidental features/bugs like these.

PR #600, opened 2024-05-29
Support Azure identity token in docker config

ddeville:
When using `az acr login` to authenticate with the ACR (Azure Container Registry) the Docker `config.json` file gets populated with an `auth` field that contains the base64 encoding of `username:password` where `username` is `00000000-0000-0000-0000-000000000000` (that actually gets ignored I believe) and `password` is empty, and an `identitytoken` entry. The latter is a short-lived token and for the use case here can be used as the password.

Tested with my own private ACR and with this change I was able to `oci_pull` an image from it after having `az acr login` populate my `~/.docker/config.json`.

(Note: buildifier also ran on the file, hope that's ok otherwise I can remove these changes from the commit)

### DIFF
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -4,7 +4,6 @@ load("@aspect_bazel_lib//lib:base64.bzl", "base64")
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
 load(":util.bzl", "util")
 
-
 # Unfortunately bazel downloader doesn't let us sniff the WWW-Authenticate header, therefore we need to
 # keep a map of known registries that require us to acquire a temporary token for authentication.
 _WWW_AUTH = {
@@ -167,6 +166,8 @@ def _get_auth(rctx, state, registry):
                     login, sep, password = base64.decode(raw_auth).partition(":")
                     if not sep:
                         fail("auth string must be in form username:password")
+                    if not password and "identitytoken" in auth_val:
+                        password = auth_val["identitytoken"]
                     pattern = {
                         "type": "basic",
                         "login": login,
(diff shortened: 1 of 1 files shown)


thesayyn:
Hmm, this looks like really specific to Azure, is there documentation somewhere that explains this?

ddeville:
[These docs](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token) do mention using the `00000000-0000-0000-0000-000000000000` as username and the token as password when doing `docker login`.

thesayyn:
i believe https://github.com/bazel-contrib/rules_oci/pull/626 PR fixes this correctly. We do use standard docker auth mechanism which doesn't need a change that looks like this. 

Going to close it in favor of #626

ddeville:
That's great thanks!

However I believe #626 fixes a different issue: dealing with creds helpers that return an identity token.

The issue that I'm trying to address is `oci_pull` trying to use an existing docker config file that includes an identity token, because of a previous `az acr login` on the host.

To give a more concrete example, in order to `docker login` with an ACR, one runs `az acr login` which pulls the identity token from Azure and does `docker login` with it. It results in a `config.json` file that looks like:

```
{
        "auths": {
                "my_acr_name.azurecr.io": {
                        "auth": "MDAwMDAwMDAtMDAwMC0wMDAwLTAwMDAtMDAwMDAwMDAwMDAwOg==",
                        "identitytoken": "SOME_VERY_LOGING_IDENTITY_TOKEN_AS_A_JWT"
                }
        }
}
```

In this context the value of auth is no more that base64-encoded `00000000-0000-0000-0000-000000000000:` and the real "password" is the token.

Today in `rules_oci` there's no way to do an `oci_pull` that uses a docker config that looks like that. It will just use the `auth` entry, split it into the fake username and the empty password and get a 401 when it tries to get an auth token...

PR #602, opened 2024-05-31
fix: Support global credsStore in docker config

### LINKED ISSUE #388: Unable to authenticate against ECR using amazon-ecr-credential-helper
I am required to use aws sso to login (`aws sso login` via CLI), and I also use the [amazon-ecr-credential-helper](https://github.com/awslabs/amazon-ecr-credential-helper). My `~/.docker/config.json` looks like this

```
{
        "credsStore": "ecr-login"
}
```
After doing the sso login, `docker pull image` works just fine when I'm pulling from the private ECR registry, but I am still getting the 401 when I use the oci_pull rules. 

If I remove the `credsStore` line from the `config.json`, then login following the [instructions here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html), the oci_pull works fine. 

It appears the credential helper isn't supported even though this [comment](https://github.com/bazel-contrib/rules_oci/issues/353#issuecomment-1719955082) says 
>  if docker pull works, then oci_pull should too

Also this is somewhat related to the auth issue in #353 but since that is closed and this issue appears to be a credential-helper issue, I'm creating a new issue.

msiebuhr:
It is perfectly normal to have config.json with just the global config store:

    { "credsStore": "osxkeychain" }

See https://github.com/docker/docker-credential-helpers?tab=readme-ov-file#usage and https://github.com/google/go-containerregistry/tree/main/pkg/authn#helpers for examples.

Previously, the authn-code would only look at the `credsStore` if there was a specific host maching in `auths`, but lacking other authn-fields.

Now it also looks at the global `credsStore` as a fallback.

Fixes #388.

### DIFF
--- a/e2e/pull/test.bats
+++ b/e2e/pull/test.bats
@@ -91,6 +91,17 @@ EOF
     assert_success
 }
 
+@test "global credstore" {
+    cat > "$DOCKER_CONFIG/config.json" <<EOF
+{
+  "credsStore": "oci"
+}
+EOF
+    update_assert '{"Authorization": ["Basic dGVzdGluZzpvY2k="]}'
+    run bazel build @empty_image//... $BAZEL_FLAGS
+    assert_success
+}
+
 @test "credstore misbehaves" {
     cat > "$DOCKER_CONFIG/config.json" <<EOF
 {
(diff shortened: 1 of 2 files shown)


thesayyn (approved):
GREAT! thanks for adding tests.

Issue #603, opened 2024-05-31
Reenable bcr presubmit for macos


thesayyn:
https://github.com/bazelbuild/bazel-central-registry/pull/2120 was failing on macos, and @alexeagle made this commit to make it green, https://github.com/bazelbuild/bazel-central-registry/pull/2120/commits/4b28f9eec6f524e8c7c6acbb28a06e5ba38a2d03

we should enable macos back once we figure out how to exclude docker tests from macos.

PR #604, opened 2024-05-31
chore: disable presubmit for macos

thesayyn:
https://github.com/bazel-contrib/rules_oci/issues/603


.bcr/presubmit.yml
```diff
@@ -6,7 +6,8 @@ bcr_test_module:
     # File "C:/b/eznkjcld/external/rules_oci~override/oci/repositories.bzl", line 100, column 64, in _stucture_test_repo_impl
     # integrity = ST_VERSIONS[repository_ctx.attr.st_version][platform],
     # Error: key "windows-amd64" not found in dictionary
-    platform: ["debian10", "macos", "ubuntu2004"]
+    # TODO(#603): add macos
+    platform: ["debian10", "ubuntu2004"]
```
alexeagle, line 10: I think we could just make a correct fix now instead - the `test_targets` below should have a different result for one platform. I can take it if you like.

thesayyn: that'd be better. let me close this PR and cut a release so you can iterate on BCR until you get it green.

PR #605, opened 2024-05-31
chore: disable dockerfile example

thesayyn:
This example required a bit of preconfiguration in CI settings, but the release pipeline uses a composite ruleset from bazel-contrib so we can't change it.  Disabling the examples/dockerfile for now.


alexeagle (approved):
unblocking to get a release out - I think this could just use a different job that's outside the bazel-contrib matrix though

Issue #610, opened 2024-06-03
Document how to replace per-platform repos


aran:
For some time rules_oci generated per-platform repos, e.g.

```
oci.pull(
    name = "sealed_secrets_controller",
    digest = "sha256:5477ce0b63030b70fe27ffe76664bfee572059035bbaa30927394fadc0517cd3",
    image = "index.docker.io/bitnami/sealed-secrets-controller",
    # tag = "0.26.1",
    platforms = [
        "linux/amd64",
        "linux/arm64",
    ],
)
use_repo(oci, "sealed_secrets_controller")
use_repo(oci, "sealed_secrets_controller_linux_amd64")
```
...
```
oci_push(
    name = "push_sealed_secrets_controller_image",
    image = @sealed_secrets_controller_linux_amd64,
    repository = REGISTRY + "/" + IMAGE_NAME,
    visibility = ["//visibility:public"],
)
```

I'm not sure exactly what version, but at some point `rules_oci` has started warning about this in both the 1.x and 2.x branches:

```
WARNING: .../MODULE.bazel:378:20: The module extension oci defined in @rules_oci//oci:extensions.bzl reported incorrect imports of repositories via use_repo():

Imported, but reported as indirect dependencies by the extension:
    ..., sealed_secrets_controller_linux_amd64

Fix the use_repo calls by running 'bazel mod tidy'.
```

I didn't see anything in the release notes about a change though.

I'm assuming I'm supposed to wrap my `oci_push` in a platform transition, or always use `--platforms` with it, but not 100% if I'm out of line with the latest recommended usage. `bazel mod sync` just removes those dependencies, and the `oci_push` then breaks. Since something has changed in the API, it would be great to have a release note or some other guidance on this. I took a brief glance through the project log but didn't immediately see what change would have done this.

aran:
As an aside that would bypass this entirely, it would be nice if 

```
oci.pull(name = "<X>"...)
...
oci_push(...image = @<X>)
```
"just worked", but from what I can tell that's actually tricky.

thesayyn:
This was broken in the alpha release. now we fixed it via #611 which should generate correct `use_repo` calls.

aran:
Thank you so much! Works* for me in the 2.x branch. Not sure if you intended to merge to 1.x also, I don't see it there.

[...]

PR #611, opened 2024-06-04
fix: return correct extension metadata

thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/529 https://github.com/bazel-contrib/rules_oci/issues/610

### DIFF
--- a/oci/extensions.bzl
+++ b/oci/extensions.bzl
@@ -42,16 +42,21 @@ def _oci_extension(module_ctx):
             )
 
             if mod.is_root:
-                if module_ctx.is_dev_dependency(pull):
-                    root_direct_dev_deps.append(pull.name)
-                else:
-                    root_direct_deps.append(pull.name)
+                deps = root_direct_dev_deps if module_ctx.is_dev_dependency(pull) else root_direct_deps
+                deps.append(pull.name)
+                for platform in pull.platforms:
+                    deps.append("_".join([pull.name] + platform.split("/")))
+
         for toolchains in mod.tags.toolchains:
             if toolchains.name != "oci" and not mod.is_root:
                 fail("""\
                 Only the root module may override the default name for the oci toolchains.
                 This prevents conflicting registrations in the global namespace of external repos.
                 """)
+            if mod.is_root:
+                deps = root_direct_dev_deps if module_ctx.is_dev_dependency(toolchains) else root_direct_deps
+                deps.append("%s_crane_toolchains" % toolchains.name)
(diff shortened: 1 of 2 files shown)


alexeagle (approved):

oci/extensions.bzl
```diff
@@ -42,16 +42,21 @@ def _oci_extension(module_ctx):
...
 
             if mod.is_root:
-                if module_ctx.is_dev_dependency(pull):
-                    root_direct_dev_deps.append(pull.name)
-                else:
-                    root_direct_deps.append(pull.name)
+                deps = root_direct_dev_deps if module_ctx.is_dev_dependency(pull) else root_direct_deps
+                deps.append(pull.name)
+                for platform in pull.platforms:
+                    deps.append("_".join([pull.name] + platform.split("/")))
```
aignas, line 48: Out of curiosity, what would be the reason to include these deps here? `rules_oci` v1.7 works without me including them via `use_repo` and I am curious, why I am suddenly asked to include them via the `bazel mod tidy` command.

alexeagle: My understanding is it resolves an error like

```
WARNING: /builds/app/MODULE.bazel:38:20: The module extension oci defined in @rules_oci//oci:extensions.bzl reported incorrect imports of repositories via use_repo():
Not imported, but reported as direct dependencies by the extension (may cause the build to fail):
    java_base_debug_nonroot_debian12, java_base_nonroot_debian12
    java_base_debug_nonroot_debian12_linux_amd64, java_base_debug_nonroot_debian12_linux_arm64_v8, java_base_nonroot_debian12_linux_amd64, java_base_nonroot_debian12_linux_arm64_v8
Fix the use_repo calls by running 'bazel mod tidy'.
```

thesayyn: `bazel mod tidy` could support something like `absolutely neccesary repos` vs `could have if build requires it` kind of `use_repo` calls.

thesayyn: but this what we have to do for now, if we want to avoid that annoying warning message. @fmeum probably has better ideas for this.

fmeum: I like @thesayyn's idea of something similar to a `root_module_optional_deps` attribute that silences the warning without also adding `use_repo` arguments. Could you file an issue for this if you think it would be useful?

thesayyn: @fmeum  Yes, something like that is what we need, i'll file an issue.

Issue #607, opened 2024-06-03
[v2.0.0-alpha2] Fail to use the platform transition group and the filegroup to get the tarball


aignas:
With the following MODULE.bazel:
```starlark
module(name = "repro")

bazel_dep(name = "rules_oci", version = "2.0.0-alpha2")
bazel_dep(name = "aspect_bazel_lib", version = "2.7.7")

oci = use_extension("@rules_oci//oci:extensions.bzl", "oci")

# Declare external images you need to pull, for example:
oci.pull(
    name = "alpine",
    digest = "sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b",
    image = "alpine",
    platforms = [
        "linux/386",
        "linux/amd64",
        "linux/arm/v6",
        "linux/arm/v7",
        "linux/arm64/v8",
        "linux/ppc64le",
        "linux/s390x",
    ],
)
use_repo(oci, "alpine")
```

And a `BUILD.bazel`:
```starlark
load("@aspect_bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
load("@rules_oci//oci:defs.bzl", "oci_tarball")

platform(
    name = "linux_x86_64",
    constraint_values = [
        "@platforms//os:linux",
        "@platforms//cpu:x86_64",
    ],
)

platform_transition_filegroup(
    name = "transitioned",
    srcs = ["@alpine"],
    target_platform = "//:linux_x86_64",
)

oci_tarball(
    name = "tarball",
    image = "transitioned",
    repo_tags = ["rules_oci_repro_alpine"],
)

filegroup(
    name = "tarball_filegroup",
    srcs = [":tarball"],
    output_group = "tarball",
)
```

The building of `tarball_filegroup` will fail with:
```console
$ bazel build :tarball_filegroup
INFO: Analyzed target //:tarball_filegroup (82 packages loaded, 788 targets configured).
ERROR: /Users/ignas.anikevicius/src/github/aignas/rules_oci_2_alpha_tarball_repro/BUILD.bazel:18:12: OCITarball tarball/tarball.tar failed: (Exit 1): tar failed: error executing OCITarball command (from target //:tarball) external/aspect_bazel_lib~~toolchains~bsd_tar_darwin_amd64/tar --create --no-xattr --no-mac-metadata --cd bazel-out/darwin_x86_64-fastbuild/bin --file ... (remaining 2 arguments skipped)

[...]

PR #614, opened 2024-06-06
fix: make oci_tarball work with disjoint output root

thesayyn:
A while back we changed oci_tarball for performance reasons, and we introduced two different ways to load tarball into the daemon, one where one run `bazel run :tarball_target` or `bazel build :tarball_target --output_groups=tarball && docker load -i tarball.path` or directly get the output in DAG. 

So in order to support all these cases we introduced some hacks like using `--cd` option to cd back into the execroot and strip the paths using `<ctx.bin_dir.path>`. 

Unfortunately this did not work for cases where bindir of oci_tarball and oci_image is disjoint, eg oci_image is transitioned but oci_tarball isn't simply because the assumption that everything is in the same `bazel-bin/cfg/bin` did not hold true in cases above. 

This PR fixes it by making eveything relative to execroot, which the case for action version by default, by introducing some execroot detection code copied from rules_js.

### DIFF
--- a/examples/assertion/BUILD.bazel
+++ b/examples/assertion/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@aspect_bazel_lib//lib:tar.bzl", "tar")
+load("@aspect_bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
-load("//oci:defs.bzl", "oci_image")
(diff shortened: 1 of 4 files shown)


thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/607

Strum355:
Confirmed this fixes #607 for us. We also transition oci_image

alexeagle (approved):

PR #616, opened 2024-06-07
    fix: env vars containing =

thesayyn:
This fixes a bug where oci_image removed the content of env variables after first `=` in the value.  This fixes that issue and adds more tests to test against other cases. 

I added a nice assertion rule similar to `assert_tar_contains` which doesn't depend on a docker daemon and doesn't require a yaml file as a configuration language.

(it was already somewhere in a branch, i just added it here so i can use it in the future.)

### DIFF
--- a/fetch.bzl
+++ b/fetch.bzl
@@ -4,7 +4,7 @@ This file is similar to how bazel_gazelle can manage go_repository calls
 by writing them to a generated macro in a .bzl file.
 """
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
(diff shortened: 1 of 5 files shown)


fetch.bzl
```diff
@@ -222,3 +222,30 @@ genrule(
         ],
         sha256 = "d7c7af5d86f43a885069408a89788f67f248e8124c682bb73936f33874e0611b",
     )
+
+    http_file(
```
alexeagle, line 226: I should put rules_multitool in here as devdep

alexeagle (approved):

fetch.bzl
thesayyn: that'd be great a follow-up

PR #619, opened 2024-06-08
feat: bump and assert crane, zot and cosign versions

farcop:


thesayyn (changes requested):
We don't use most of these anymore. Can you just bump the crane version so it's easier to cherry pick?

farcop:
May be its better just close this PR then?

Issue #588, opened 2024-05-23
Bug: "invalid path" with rules_oci 2.0.0-alpha0


mering:
We get the following error when upgrading to the v2 pre-release:
```
invalid reference, invalid path for scheme "ocidir": bazel-out/k8-fastbuild/bin/external/repo~/package/test-image:intermediate
```

When building via `bazel build --config=remote @repo//package:test-image`. My spontaneous guess would be the Bazel version separator for external dependencies `~` or the target separator `:`.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/622

PR #622, opened 2024-06-10
    fix: support external oci_image targets

thesayyn:
fixes https://github.com/bazel-contrib/rules_oci/issues/588, blocked on: https://github.com/regclient/regclient/pull/763


mering:
@thesayyn FYI seems like the blocker has been resolved upstream.

thesayyn:
We need a new version of regctl before we can solve it here.

alexeagle (approved):

alexeagle:
@thesayyn this PR broke our BCR publishing https://github.com/bazelbuild/bazel-central-registry/pull/2563 because it assumes e2e/ can reference examples/.
Easy fix is to start publishing examples/ in our release archive - is that desirable?

PR #623, opened 2024-06-11
nixos action_env

lromor:
In some systems (nixos), action_envs are used to tell the sandbox environment where to find basic system libraries. With this change, the user replace for instance:

```
test          --test_env=NIX_LD=/run/current-system/sw/share/nix-ld/lib/ld.so        --test_env=NIX_LD_LIBRARY_PATH=/run/current-system/sw/share/nix-ld/lib
common      --action_env=NIX_LD=/run/current-system/sw/share/nix-ld/lib/ld.so      --action_env=NIX_LD_LIBRARY_PATH=/run/current-system/sw/share/nix-ld/lib
common --host_action_env=NIX_LD=/run/current-system/sw/share/nix-ld/lib/ld.so --host_action_env=NIX_LD_LIBRARY_PATH=/run/current-system/sw/share/nix-ld/lib
```

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -188,6 +188,7 @@ def _oci_image_impl(ctx):
         executable = util.maybe_wrap_launcher_for_windows(ctx, builder),
         tools = [crane.crane_info.binary, registry.registry_info.launcher, registry.registry_info.registry, jq.jqinfo.bin],
         mnemonic = "OCIImage",
+        use_default_shell_env = True,
         progress_message = "OCI Image %{label}",
     )
 
--- a/oci/private/image_index.bzl
+++ b/oci/private/image_index.bzl
@@ -65,6 +65,7 @@ def _oci_image_index_impl(ctx):
         arguments = [args],
         outputs = [output],
         executable = launcher,
+        use_default_shell_env = True,
         tools = [yq.yqinfo.bin, coreutils.coreutils_info.bin],
         mnemonic = "OCIIndex",
         progress_message = "OCI Index %{label}",
--- a/oci/private/tarball.bzl
+++ b/oci/private/tarball.bzl
@@ -101,6 +101,7 @@ def _tarball_impl(ctx):
         outputs = [tarball],
         tools = [yq_bin],
         mnemonic = "OCITarball",
+        use_default_shell_env = True,
         progress_message = "OCI Tarball %{label}",
     )
 


thesayyn:
There are some weird behavior around use_default_shell_env which i don't want rules_oci to get involved in and this would cause more action cache for everyone even if they don't use Nix. 

Note that if these variables are already exported from your shell env, you can simply do 

```shell
common --test_env=NIX_LD --action_env=NIX_LD --host_action_env=NIX_LD
common --test_env=NIX_LD_LIBRARY_PATH --action_env=NIX_LD_LIBRARY_PATH --host_action_env=NIX_LD_LIBRARY_PATH
```

lromor:
I figured out the issue is connected with the current packaging of `bazel_7` on nixos. Specifically, https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/development/tools/build-managers/bazel/bazel_6/actions_path.patch this patch doesn't seem to be applied.

lromor:
Hi @fionera , yes, latest bazel packages (bazel 7) doesn't use that patch. 
What we ended up using a slightly modified hack:
```
diff --git a/src/main/java/com/google/devtools/build/lib/shell/JavaSubprocessFactory.java b/src/main/java/com/google/devtools/build/lib/shell/JavaSubprocessFactory.java
index 649647c5f2..0bd28eae23 100644
--- a/src/main/java/com/google/devtools/build/lib/shell/JavaSubprocessFactory.java
+++ b/src/main/java/com/google/devtools/build/lib/shell/JavaSubprocessFactory.java
@@ -165,7 +165,7 @@ public class JavaSubprocessFactory implements SubprocessFactory {
     }
     builder.command(argv);
     if (params.getEnv() != null) {
-      builder.environment().clear();
+      builder.environment().entrySet().removeIf(entry -> !(entry.getKey().equals("NIX_LD_LIBRARY_PATH") || entry.getKey().equals("PATH")));
       builder.environment().putAll(params.getEnv());
     }
```

To let only PATH and NIX_LD_LIBRARY_PATH pass through.

Via
```
        bazelOverride = pkgs.bazel_7.overrideAttrs (
          previousAttrs: {
            # Provide our own hack, latest enableNixHacks introduces repo issues.
            patches = previousAttrs.patches ++ [
              ./bazel.patch
            ];
          });
```

Depending on your needs, you can also scope the PATH and packages by wrapping the final bazel binary.
I've opened an issue to allow optionally such behavior at: https://github.com/NixOS/nixpkgs/issues/359895#issuecomment-2531256898

PR #626, opened 2024-06-12
Oauth2 support without Python

lukaszwawrzyk:
Due to inactivity, I picked up this PR: https://github.com/bazel-contrib/rules_oci/pull/581

I implememented the mentioned workarounds, i.e. in case of windows, I try to run script using powershell, otherwise, I fallback to curl and then wget, both wrapped in a bash script.

I tested these with an azure repo which requires oauth2. I also tested it on windows. I found that the config.json is not found due to "stat" binary not existing on windows, so I fixed this too.

### DIFF
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -130,13 +130,7 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
         "password": response["Secret"],
     }
 
-
-def _oauth2(rctx, realm, scope, service, secret):
-    if rctx.os.name.startswith("windows") and rctx.which("powershell"):
-        executable = "oauth2.ps1"
-        rctx.file(
-            executable,
-            content = """\
+OAUTH_2_SCRIPT_POWERSHELL = """\
 param (
     [string]$url,
     [string]$service,
@@ -160,13 +154,8 @@ try {
     exit 1
 }
 """
-        )
-        result = rctx.execute(["powershell", "-File", rctx.path(executable), realm, service, scope, secret])
-    elif rctx.which("curl"):
-        executable = "oauth2.sh"
-        rctx.file(
-            executable,
-            content = """\
+
+OAUTH_2_SCRIPT_CURL = """\
 url=$1
 service=$2
 scope=$3
@@ -181,13 +170,8 @@ fi
 
 echo "$response"
 """
-        )
-        result = rctx.execute(["bash", rctx.path(executable), realm, service, scope, secret])
-    elif rctx.which("wget"):
-        executable = "oauth2.sh"
-        rctx.file(
-            executable,
-            content = """\
+
+OAUTH_2_SCRIPT_WGET = """\
 url=$1
 service=$2
 scope=$3
@@ -202,7 +186,19 @@ fi
 
 echo "$response"
 """
-        )
+
+def _oauth2(rctx, realm, scope, service, secret):
+    if rctx.os.name.startswith("windows") and rctx.which("powershell"):
+        executable = "oauth2.ps1"
+        rctx.file(executable, content = OAUTH_2_SCRIPT_POWERSHELL)
+        result = rctx.execute(["powershell", "-File", rctx.path(executable), realm, service, scope, secret])
+    elif rctx.which("curl"):
+        executable = "oauth2.sh"
+        rctx.file(executable, content = OAUTH_2_SCRIPT_CURL)
+        result = rctx.execute(["bash", rctx.path(executable), realm, service, scope, secret])
+    elif rctx.which("wget"):
+        executable = "oauth2.sh"
+        rctx.file(executable, content = OAUTH_2_SCRIPT_WGET)
         result = rctx.execute(["bash", rctx.path(executable), realm, service, scope, secret])
     else:
         fail("oauth2 failed, could not find either of: curl, wget, powershell")
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -130,63 +130,86 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
         "password": response["Secret"],
     }
 
+
 def _oauth2(rctx, realm, scope, service, secret):
-    if not rctx.os.environ.get("SUPPORT_OAUTH2", False):
-        fail("oauth2 support is disabled (set SUPPORT_OAUTH2 to True)")
+    if rctx.os.name.startswith("windows") and rctx.which("powershell"):
+        executable = "oauth2.ps1"
+        rctx.file(
+            executable,
+            content = """\
+param (
+    [string]$url,
+    [string]$service,
+    [string]$scope,
+    [string]$refresh_token
+)
 
-    python_bin = "python3"
+try {
+    $response = Invoke-RestMethod -Uri $url -Method Post -Body @{
+        grant_type = "refresh_token"
+        service = $service
+        scope = $scope
+        refresh_token = $refresh_token
+    } -ErrorAction Stop
 
-    # oauth2 requires python to be available in the env
-    if not rctx.which(python_bin):
-        fail("oauth2 failed: to use oauth2, make sure %s is available" % python_bin)
-
-    # oauth2 requires "requests" package installed alongside python
-    result = rctx.execute([python_bin, "-c", "import requests"])
-    if result.return_code:
-        fail("oauth2 failed: to use oauth2, make sure \"requests\" package is installed")
-
-    executable = "oauth2.py"
-
-    rctx.file(
-        executable,
-        content = """\
-import json
-import requests
-import sys
-
-if __name__ == "__main__":
-    url = sys.argv[1]
-    
-    params = {}
-    params["grant_type"] = "refresh_token"
-    params["service"] = sys.argv[2]
-    params["scope"] = sys.argv[3]
-    params["refresh_token"] = sys.argv[4]
-    
-    response = requests.post(url, data=params)
-
-    ret = {
-        "code": response.status_code,
-    }
-
-    if response.status_code == 200:
-        ret["response"] = json.loads(response.text.encode('utf8'))
-    else:
-        ret["response"] = {}
-
-    print(json.dumps(ret))
+    $jsonResponse = $response | ConvertTo-Json
+    echo $jsonResponse
+} catch {
+    $ErrorMessage = $_.Exception.Message
+    Write-Error "oauth2 failed: PowerShell request failed with error: $ErrorMessage"
+    exit 1
+}
 """
-    )
-    
-    result = rctx.execute([python_bin, rctx.path(executable), realm, service, scope, secret])
+        )
+        result = rctx.execute(["powershell", "-File", rctx.path(executable), realm, service, scope, secret])
+    elif rctx.which("curl"):
+        executable = "oauth2.sh"
+        rctx.file(
+            executable,
+            content = """\
+url=$1
+service=$2
+scope=$3
+refresh_token=$4
+
+response=$(curl --silent --show-error --fail --request POST --data "grant_type=refresh_token&service=$service&scope=$scope&refresh_token=$refresh_token" $url)
+
+if [ $? -ne 0 ]; then
+    echo "oauth2 failed: curl request failed" >&2
+    exit 1
+fi
+
+echo "$response"
+"""
+        )
+        result = rctx.execute(["bash", rctx.path(executable), realm, service, scope, secret])
+    elif rctx.which("wget"):
+        executable = "oauth2.sh"
+        rctx.file(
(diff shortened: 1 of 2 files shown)


oci/private/util.bzl
```diff
@@ -171,7 +171,10 @@ if defined args (
     return win_launcher
 
 def _file_exists(rctx, path):
-    result = rctx.execute(["stat", path])
+    if rctx.os.name.startswith("windows"):
```
thesayyn, line 174: This could probably could into a separate PR. note that we don't have windows CI ATM.

oci/private/authn.bzl
```diff
@@ -124,15 +124,93 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
...
+
+echo "$response"
+"""
+        )
+        result = rctx.execute(["bash", rctx.path(executable), realm, service, scope, secret])
+    elif rctx.which("wget"):
+        executable = "oauth2.sh"
+        rctx.file(
+            executable,
+            content = """\
```
thesayyn, line 190: i prefer these long templated scripts to be above the script into a constant variable.

oci/private/authn.bzl
```diff
@@ -124,15 +124,93 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
...
+    exit 1
+}
+"""
+        )
+        result = rctx.execute(["powershell", "-File", rctx.path(executable), realm, service, scope, secret])
+    elif rctx.which("curl"):
+        executable = "oauth2.sh"
+        rctx.file(
+            executable,
+            content = """\
```
thesayyn, line 169: i prefer these long templated scripts to above the script into a constant variable. (same as below)

oci/private/authn.bzl
```diff
@@ -124,15 +124,93 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
...
         "password": response["Secret"],
     }
 
+
+def _oauth2(rctx, realm, scope, service, secret):
+    if rctx.os.name.startswith("windows") and rctx.which("powershell"):
+        executable = "oauth2.ps1"
+        rctx.file(
+            executable,
+            content = """\
```
thesayyn, line 139: i prefer these long templated scripts to above the script into a constant variable.

oci/private/authn.bzl
```diff
@@ -201,16 +279,34 @@ def _get_token(rctx, state, registry, repository):
             # if a token for this repository and registry is acquired, use that instead.
             if url in state["token"]:
                 return state["token"][url]
+            
+            auth = None
+            if pattern["login"] == "<token>":
```
thesayyn, line 284: Okay, this feels like the best we could do here. However, i don't want to commit this to the public api yet. Can we hide this feature behind a flag and print a fail statement that say something along the lines of;

```
OAuth2 support for oci_pull is highly experimental and is not enabled by default. 

We may change or abandon it without a notice. Use it at your own peril!

To enable this feature, add `common --repo_env=OCI_ENABLE_OAUTH2_SUPPORT` to the `.bazelrc` file.
```

oci/private/authn.bzl
```diff
@@ -124,15 +124,93 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
...
+            content = """\
+url=$1
+service=$2
+scope=$3
+refresh_token=$4
+
+response=$(wget --quiet --output-document=- --post-data "grant_type=refresh_token&service=$service&scope=$scope&refresh_token=$refresh_token" $url)
+
+if [ $? -ne 0 ]; then
+    echo "oauth2 failed: wget request failed" >&2
```
thesayyn, line 199: can we print more information about why it failed?

oci/private/authn.bzl
```diff
@@ -124,15 +124,93 @@ exec "docker-credential-{}" get <<< "$1" """.format(helper_name),
...
+            content = """\
+url=$1
+service=$2
+scope=$3
+refresh_token=$4
+
+response=$(curl --silent --show-error --fail --request POST --data "grant_type=refresh_token&service=$service&scope=$scope&refresh_token=$refresh_token" $url)
+
+if [ $? -ne 0 ]; then
+    echo "oauth2 failed: curl request failed" >&2
```
thesayyn, line 178: maybe more information about why it failed?

thesayyn (changes requested):

thesayyn:
Okay, this looks a bit better, this could use some tests similar to https://github.com/bazel-contrib/rules_oci/blob/2.x/e2e/assertion/oci_pull_auth_tests.bats

We will also need better error messages.

Sowmya-Iyer:
While trying to oci.pull an azure container registry, I still get this error: 
```
Error in download: java.io.IOException: Error downloading [<>] to /home/siyer/.cache/bazel/_bazel_siyer/c0c06e0bd45258079fb49c5a3cba7fd4/external/rules_oci~~oci~ard_x86_64_base/www-authenticate.json: GET returned 401 Unauthorized 
```
The PR #600 had actually fixed that issue for me though I didn't like the way it was done either

thesayyn:
we should somehow incorporate that change here. identity tokens are not something we support at the moment, so we might as well it right at once and improve on it.

lukaszwawrzyk:
> While trying to oci.pull an azure container registry, I still get this error:
> 
> ```
> Error in download: java.io.IOException: Error downloading [<>] to /home/siyer/.cache/bazel/_bazel_siyer/c0c06e0bd45258079fb49c5a3cba7fd4/external/rules_oci~~oci~ard_x86_64_base/www-authenticate.json: GET returned 401 Unauthorized 
> ```
> 
> The PR #600 had actually fixed that issue for me though I didn't like the way it was done either

If I `az acr login --name <my repo>` from terminal and then run bazel with changes in this PR it works. While it doesn't work without these changes at all.

oci/private/authn.bzl
lukaszwawrzyk: `rctx.getenv("OCI_ENABLE_OAUTH2_SUPPORT")` returns `None` for suggested `--repo_env=OCI_ENABLE_OAUTH2_SUPPORT`, so instead I suggest to add `common --repo_env=OCI_ENABLE_OAUTH2_SUPPORT=1`

lukaszwawrzyk: `rctx.getenv("OCI_ENABLE_OAUTH2_SUPPORT")` returns `None` for suggested `--repo_env=OCI_ENABLE_OAUTH2_SUPPORT`, so instead I suggest to add `common --repo_env=OCI_ENABLE_OAUTH2_SUPPORT=1`

oci/private/authn.bzl
lukaszwawrzyk: I removed this line altogether, the failure reason is already in the stderr. For the code that you refer to, it would look like:
```
Error in fail: oauth2 failed: 
STDOUT:

STDERR:
curl: (22) The requested URL returned error: 403
oauth2 failed: curl request failed
```

after removing this line it is:

```
Error in fail: oauth2 failed: 
STDOUT:

STDERR:
curl: (22) The requested URL returned error: 403
```

oci/private/authn.bzl
lukaszwawrzyk: see above

lukaszwawrzyk:
> Okay, this looks a bit better, this could use some tests similar to https://github.com/bazel-contrib/rules_oci/blob/2.x/e2e/assertion/oci_pull_auth_tests.bats
> 
> We will also need better error messages.

I think I fixed everything requested except for tests.

Current tests do not pass for me so it is hard to develop new, any suggestions?
I run them like this:
```
% pwd                                                                                                                                                             
/Users/lukasz/dev/rules/rules_oci/e2e/pull
% bazel test //:test_nobzlmod  
```
And this is the test log that I get:
[test.log](https://github.com/user-attachments/files/15820537/test.log)

Sowmya-Iyer:
> > While trying to oci.pull an azure container registry, I still get this error:
> > ```
> > Error in download: java.io.IOException: Error downloading [<>] to /home/siyer/.cache/bazel/_bazel_siyer/c0c06e0bd45258079fb49c5a3cba7fd4/external/rules_oci~~oci~ard_x86_64_base/www-authenticate.json: GET returned 401 Unauthorized 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The PR #600 had actually fixed that issue for me though I didn't like the way it was done either
> 
> If I `az acr login --name <my repo>` from terminal and then run bazel with changes in this PR it works. While it doesn't work without these changes at all.

1) I added ```common --repo_env=OCI_ENABLE_OAUTH2_SUPPORT=1``` to my bazelrc
2) uodated Module.bzl with this commit as a git_override (4f78d52446b074a377d7e12a1baf7ac8072a7343)
3) az acr login --name <container_registry>

I ran the bazel with the above three changes 

ERROR:
```
File "/home/siyer/.cache/bazel/_bazel_siyer/c0c06e0bd45258079fb49c5a3cba7fd4/external/rules_oci~/oci/private/authn.bzl", line 304, column 30, in _get_token
		rctx.download(
Error in download: java.io.IOException: Error downloading [<redacted>] to /home/siyer/.cache/bazel/_bazel_siyer/c0c06e0bd45258079fb49c5a3cba7fd4/external/rules_oci~~oci~<image_name>/www-authenticate.json: GET returned 401 Unauthorized
```

Sowmya-Iyer:
So, I think the problem is in parsing the ~/.docker/config.json:

```
siyer@siyer-ThinkPad-T15g-Gen-2i:~/workspace/container/BuildFarm$ cat ~/.docker/config.json
{
	"auths": {
		"<registry>.azurecr.io": {
			"auth": "MDAwMDAwMDAtMDAwMC0wMDAwLTAwMDAtMDAwMDAwMDAwMDAwOg==",
			"identitytoken": "<redacted>"
		}
	}
}
```

Line 240 looks for what follows  ":" as password but as per this config it is ["identity_token"] when sep (":") does not exist

oci/private/authn.bzl
thesayyn: we probably don't want to use `rctx.getenv` as it a newer api and might not work on some 6.x versions of bazel. 

`rctx.os.environ` should do the trick, i am okay with the naming suggestion.

oci/private/authn.bzl
```diff
@@ -202,15 +283,40 @@ def _get_token(rctx, state, registry, repository):
...
-            rctx.download(
-                url = [url],
-                output = "www-authenticate.json",
-                # optionally, sending the credentials to authenticate using the credentials.
-                # this is for fetching from private repositories that require WWW-Authenticate
-                auth = {url: pattern},
-            )
+            auth = None
+            if pattern["login"] == "<token>":
+                if versions.is_at_least("7.0.0", versions.get() or "0.0.0"):
```
thesayyn, line 288: even though i like the feature detection here; we probably should just use `rctx.os.environ` for now to match rest of the repo, and add `OCI_ENABLE_OAUTH2_SUPPORT` to list of env that invalidates this repo.

lukaszwawrzyk:
> So, I think the problem is in parsing the ~/.docker/config.json:
> 
> ```
> siyer@siyer-ThinkPad-T15g-Gen-2i:~/workspace/container/BuildFarm$ cat ~/.docker/config.json
> {
> 	"auths": {
> 		"<registry>.azurecr.io": {
> 			"auth": "MDAwMDAwMDAtMDAwMC0wMDAwLTAwMDAtMDAwMDAwMDAwMDAwOg==",
> 			"identitytoken": "<redacted>"
> 		}
> 	}
> }
> ```
> 
> Line 240 looks for what follows ":" as password but as per this config it is ["identity_token"] . Since here auth is weirdly ending with ":" `00000000-0000-0000-0000-000000000000:` , the "if not sep" condition fails and there is no error message and password is empty as well

My best bet is that you are using linux, I managed to reproduce the issue on ubuntu. So the fix in the current PR works for Mac OS, but not for linux.

lukaszwawrzyk:
I integrated fix from @ddeville to this PR, I believe we now have a complete support of azure across operating systems. And for oauth2 hopefully in all cases.

thesayyn (approved):
This looks fantastic now. Only we knew how to add some tests for this.

thesayyn:
I am going to hold this for a little bit so we can figure out some testing infra for this. i'll circle back to this next week.

thesayyn:
Hmm, there are some failures. Can you handle those?

lukaszwawrzyk:
@thesayyn it was a simple check missing, now it's all good

wingsofovnia:
Great work! Looking forward to having it in the next releases. 
Any chance @thesayyn we can get this merged for 1.8.1?

wingsofovnia:
Thank you. @thesayyn will this also get pulled to 2.x?

PR #632, opened 2024-06-20
feat: add support for settings volumes

### LINKED ISSUE #406: Allow setting volumes
Analogous to #220 for ports.

PR to crane to add support: https://github.com/google/go-containerregistry/pull/1824

gzm0:
Fixes #406.

### DIFF
--- a/oci/defs.bzl
+++ b/oci/defs.bzl
@@ -22,7 +22,7 @@ oci_image_rule = _oci_image
 oci_image_index = _oci_image_index
 oci_push_rule = _oci_push
 
-def oci_image(name, labels = None, annotations = None, env = None, cmd = None, entrypoint = None, exposed_ports = None, **kwargs):
+def oci_image(name, labels = None, annotations = None, env = None, cmd = None, entrypoint = None, exposed_ports = None, volumes = None, **kwargs):
     """Macro wrapper around [oci_image_rule](#oci_image_rule).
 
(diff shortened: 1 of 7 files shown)


oci/defs.bzl
```diff
@@ -22,7 +22,7 @@ oci_image_rule = _oci_image
 oci_image_index = _oci_image_index
 oci_push_rule = _oci_push
 
-def oci_image(name, labels = None, annotations = None, env = None, cmd = None, entrypoint = None, exposed_ports = None, **kwargs):
+def oci_image(name, labels = None, annotations = None, env = None, cmd = None, entrypoint = None, exposed_ports = None, volumes = None, **kwargs):
```
gzm0, line 25: Regarding the naming of the argument (referring to the discussion here: https://github.com/bazel-contrib/rules_oci/pull/503#discussion_r1582505282).

The reference in that comment is about mounts in running containers, not the volume directive in images.

Volumes in images are specified here (and called volumes):
https://github.com/opencontainers/image-spec/blob/main/config.md#properties

Mounts are concrete attachments to paths inside the container (so they need a counterpart path on the host or an actual volume).

Volumes are merely metadata about an image, stating that it stores data there. (docker uses this to automatically create anonymous volumes, I do not know whether this behavior is specified).

thesayyn (approved):

PR #386, opened 2023-10-04
refactor: remove yq dependency

thesayyn:
Fixes #384. blocked on bazel-lib 2.0

PR #633, opened 2024-06-20
fix: accumulate image manifests in oci_image_index

### LINKED ISSUE #638: multi arch images broken on 2.x
In the 2.x branch and respective releases the functionality to create multi arch images is broken due to jq overwriting the first manifest while creating the index manifest.

https://github.com/bazel-contrib/rules_oci/blob/0d90696187ab86b0ab73a7b03a163e661799a6e3/oci/private/image_index.sh.tpl#L27

Here the jq command should merge the existing manifest with the newly added one from the image. But instead it overrides the existing one. JQ has an explicit append to array operator _+=_ which is better suited in this context.

seh:
In the recent bazel-contrib/rules_oci#386, when we replaced use of the _yq_ tool with the _jq_ tool, we [translated the `oci_image_index` rule's accumulation of image manifests incorrectly](https://github.com/bazel-contrib/rules_oci/pull/386/files#r1646778569), such that we lose all but the last index that we add to the index.

In `jq`, [the _update-assignment_ operator](https://jqlang.github.io/jq/manual/#update-assignment) (`|=`) does not append to an array supplied as its left-hand context value automatically. Instead, one must use [a _complex assignment_](https://jqlang.github.io/jq/manual/#complex-assignments), mentioning the context value explicitly together with [the addition operator](https://jqlang.github.io/jq/manual/#addition).

See [the preceding discussion in the "docker" channel](https://bazelbuild.slack.com/archives/CA3NW13MH/p1718836203461049) of the "Bazel" Slack workspace a chronicle of discovering this defect.

Fixes #638.

### DIFF
--- a/oci/private/image_index.sh.tpl
+++ b/oci/private/image_index.sh.tpl
@@ -24,7 +24,7 @@ function add_image() {
         local platform=$("${JQ}" -c '{"os": .os, "architecture": .architecture, "variant": .variant, "os.version": .["os.version"], "os.features": .["os.features"]} | with_entries(select( .value != null ))' "${image_path}/blobs/${config_blob_path}")
         "${JQ}" --argjson platform "${platform}" \
                 --argjson manifest "${manifest}" \
-                '.manifests |= [$manifest + {"platform": $platform}]'\
+                '.manifests += [$manifest + {"platform": $platform}]'\
                 "${output_path}/manifest_list.json" > "${output_path}/manifest_list.new.json"
         "${COREUTILS}" cp --no-preserve=mode "${output_path}/manifest_list.new.json" "${output_path}/manifest_list.json"
     done
--- a/examples/multi_architecture_image/BUILD.bazel
+++ b/examples/multi_architecture_image/BUILD.bazel
@@ -43,5 +43,5 @@ genrule(
 assert_contains(
     name = "check_digest",
     actual = ":hash",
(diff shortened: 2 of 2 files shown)


thesayyn (approved):

thesayyn:
would be nice to have a test for this.

seh:
It appears that the lone CI workflow failure was due to a flaky server. I lack permission to run the workflow again. Can one of the maintainers please trigger that workflow to run again?

oci/private/image_index.sh.tpl
```diff
@@ -24,7 +24,7 @@ function add_image() {
         local platform=$("${JQ}" -c '{"os": .os, "architecture": .architecture, "variant": .variant, "os.version": .["os.version"], "os.features": .["os.features"]} | with_entries(select( .value != null ))' "${image_path}/blobs/${config_blob_path}")
         "${JQ}" --argjson platform "${platform}" \
                 --argjson manifest "${manifest}" \
-                '.manifests |= [$manifest + {"platform": $platform}]'\
+                '.manifests |= . + [$manifest + {"platform": $platform}]'\
```
seh, line 27: I do like @tobiaskohlbau's use of `+=` in #637. Which do you prefer here?

thesayyn: += probably better, let's do that

PR #635, opened 2024-06-24
tarball uses hermetic coreutils

rickvanprim:
Updates `tarball.sh.tpl` to use the hermetic coreutils.  I opted to match the existing script form of supplying a variable for the executable instead of the newer `image.sh.tpl` form of modifying `PATH` and assuming the binary name (particularly as updating `tar` doesn't seem like it would work as the binary name is different between Windows and Unix).  If you'd prefer I modify `PATH` to find `coreutils`, I could do that instead.

### DIFF
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -5,9 +5,10 @@ set -o pipefail -o errexit -o nounset
 
 readonly FORMAT="{{format}}"
 readonly JQ="{{jq_path}}"
+readonly COREUTILS="{{coreutils_path}}"
 readonly TAR="{{tar}}"
 readonly IMAGE_DIR="{{image_dir}}"
-readonly REPOTAGS=($(cat "{{tags}}"))
+readonly REPOTAGS=($("${COREUTILS}" cat "{{tags}}"))
 readonly INDEX_FILE="${IMAGE_DIR}/index.json"
 
 readonly OUTPUT="{{output}}"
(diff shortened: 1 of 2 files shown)


oci/private/tarball.sh.tpl
```diff
@@ -17,14 +18,14 @@ readonly OUTPUT="{{output}}"
 function add_to_tar() {
     content=$1
     tar_path=$2
-    echo >>"${OUTPUT}" "${tar_path} uid=0 gid=0 mode=0755 time=1672560000 type=file content=${content}"
+    "${COREUTILS}" echo >>"${OUTPUT}" "${tar_path} uid=0 gid=0 mode=0755 time=1672560000 type=file content=${content}"
```
thesayyn, line 21: isn't echo shell built-in? i think we don't need coreutils for this.

rickvanprim: Fair, I will remove that.  I was thinking it wasn't builtin to `sh` but seems I'm foolishly wrong and it's builtin everywhere: https://unix.stackexchange.com/a/1359

rickvanprim: Updated PR.  Busy day 😅

thesayyn (approved):

thesayyn:
@rickvanprim there seems to be some conflicts? can you resolve those?

rickvanprim:
> @rickvanprim there seems to be some conflicts? can you resolve those?

Done.  That was weird, the merge conflict was in `load.bzl` which I hadn't touched 🤷‍♂️

rickvanprim:
Ah the file got renamed, that makes sense 😅

PR #636, opened 2024-06-24
refactor: rename oci_tarball to oci_load

alexeagle:
The primary use is with 'bazel run' to load the image into the container daemon. Producing a tarball is an optional side-effect and not recommended for most use cases. The new name makes this more clear.


thesayyn (approved):
LGTM, maybe name could be `oci_daemon_load` to be explicit?

alexeagle:
Decided over VC that this shorter name is good.

PR #637, opened 2024-06-24
fix oci image index script

### LINKED ISSUE #638: multi arch images broken on 2.x

tobiaskohlbau:
The script does not work for multiarch images as it does not append to the manifests. Instead it replaces with the latest one which results in the last architecture to be the only one supported. JQ has an option to append an element to an array.

Fixes #638


tobiaskohlbau:
We at @myopenfactory deployed these changes into our monorepo and confirmed that it works like the 1.x versions.

seh:
This is a duplicate of #633.

tobiaskohlbau:
> This is a duplicate of #633.

Oh my bad, did look over the PRs but did not associate this. Let's try to get the other one in. Maybe you could reference the ticket #638 to solve by your ticket.

PR #641, opened 2024-07-01
Reproduction for oci_load bug

alexeagle:
Run with:

```
bazel run //examples/assertion:case12_run
```


gergelyfabian:
This should be closed. As we discussed on Slack, I needed to add the runfiles incantation at the beginning of my sh_binary target, to ensure that the rules_oci-generated load script will find its runfiles. Then everything works.

Surely, this is a yet undocumented change, but could be expected in the Bazel world.

thesayyn:
Closing this as there's nothing to fix.

Issue #643, opened 2024-07-02
rules_oci@v2beta1 bazel run with remote execution does not work correctly


juanzolotoochin:
When I do:
```
$ bazel run //path/to:oci_push
````

And I have remote execution enabled via `--bes_backend` and `--remote_download_outputs=toplevel` it fails with:

```2024/07/02 10:33:57 existing blob: sha256:e768a5010a8fdfb45f31fe540aae5d3c41466353afcf6550568fde10a5bde1cb
2024/07/02 10:33:57 existing blob: sha256:893fbf7914f5d3b1fd9cb21e026aff8272338b581ea4eea19748091b0d7ed80c
2024/07/02 10:33:57 existing blob: sha256:b30e7a7784779eeb8b5f08aba3cfe4b5222760a62a3e41293d2234d59be8a182
2024/07/02 10:33:57 existing blob: sha256:878dd088de17af19b02f566314fd514bff227297f01b074dc436701059185699
2024/07/02 10:33:57 existing blob: sha256:9305790b8fb52a3ba1e96257850b796294084cac27c403e5334b0bda0a7050d7
Error: open path/to/image/blobs/sha256/70f0a1617dc909b6d0da8c1cd1b18ca098e6c84c679d274bab4e07f52e6d9c60: no such file or directory
```

it works fine if you set `--remote_download_outputs=all` but AFAIK it should not be necessary if my understanding of the flag is correct. This works fine in v1.

thesayyn:
this is probably fixed by  #646, can you try using the 2.x branch?

mkmik:
I encountered the issue in 2.0.0-beta0. I tried 2.0.0-beta1 and it fixed the issue for me.

I assume the title of this issue has a typo and it should be beta0 since beta1 contains the fix referenced by @thesayyn and the issue predates the beta1 release

thesayyn:
Closing this as fixed by https://github.com/bazel-contrib/rules_oci/pull/646

Issue #642, opened 2024-07-01
[OCI Load] OCI Load fails if dependents are not pre-built


ankit-agarwal1999:
Hey all! We are on the beta release of rules_oci and noticed that oci_load sometimes fails if its dependents are not built. 

As an example: 

BUILD FILE: 

```

tar(
    name = "my_layer",
    testonly = True,
    srcs = [
        ":my_srcs",
    ],
)


oci_image(
    name = "my_image",
    testonly = True,
    base = "@alpine//:alpine",
    tars = [
        ":my_tar",
    ],
    workdir = "/",
)


oci_load(
    name = "my_tarball",
    testonly = True,
    format = "docker",
    image = ":my_image",
    repo_tags = [
        "bazel/tar:latest",
    ],
)
```

And command 

```
bazel run //tarball:my_tarball 
```
has the following error when run: 

```
tar: Error reading archive -: Can't open tarball/myimage/blobs/sha256/284f204b5b52127454c9404d7c426126e84c965603635598f0d76278bc767be1
tar: Error exit delayed from previous errors.                                                                                                                                                                                                                                  
open /var/lib/docker/tmp/docker-import-3272200725/blobs/json: no such file or directory
```

But if I run 

```
bazel build //tarball:my_layer
bazel run //tarball:my_tarball
```
Then this works correctly. 

Looking at the code, my guess is that the oci_load rule doesn't actually have the sha256 blobs in the runfiles of the oci_load rule, so I think those output files would likely need to be propagated up so the load rule can access it.

alexeagle:
I'm not able to reproduce.

Changes: https://github.com/bazel-contrib/rules_oci/pull/644

```
alex@a:~/Projects/rules_oci/e2e/smoke$ bazel clean --expunge
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
alex@a:~/Projects/rules_oci/e2e/smoke$ bazel run load
Starting local Bazel server and connecting to it...
WARNING: For repository 'bazel_skylib', the root module requires module version bazel_skylib@1.5.0, but got bazel_skylib@1.6.1 in the resolved dependency graph.
WARNING: For repository 'platforms', the root module requires module version platforms@0.0.8, but got platforms@0.0.9 in the resolved dependency graph.
INFO: Analyzed target //:load (71 packages loaded, 321 targets configured).
INFO: Found 1 target...
Target //:load up-to-date:
  bazel-bin/load/tarball.spec
[...]

Issue #279, opened 2023-06-22
Any plans to add providers for writing rules that build on-top of rules_oci?


ensonic:
We're trying to port from rules_docker to rules_oci (see https://github.com/googlecloudrobotics/core/issues/130), but have trouble finding a good replacement for:

load("@io_bazel_rules_docker//container:providers.bzl", "ImageInfo", "ImportInfo")
see https://github.com/googlecloudrobotics/core/blob/main/bazel/app_chart.bzl#L1

thesayyn:
Usually, our response to "adding something new" to the ruleset is why do you think you need it to accomplish what you are trying to do?

ensonic:
@thesayyn In order to port our bazel rule, we need access to the digest. From https://github.com/bazel-contrib/rules_oci/blob/main/oci/private/image.bzl#L172 it looks like there are no additonal outputs to obtain it.

thesayyn:
Is this what would you like to obtain? https://github.com/bazel-contrib/rules_oci/issues/60

alexeagle:
I think we should add an OutputGroupInfo so you can request the "digest" output file, but we wouldn't have to create any new Provider types to satisfy this FR. Sound right?

ensonic:
> Is this what would you like to obtain? #60

Thanks! That seems to work for now.

thesayyn:
Does #346 fix this for you?

gzm0:
Maybe I can chime in my experience here: The additional label with the digest file is going to be very hard to compose for downstream rules (need to recalculate a label, so forced to use a macro, not configurable, etc.).

So far, I've just resorted to read the oci-layout index.json directly. This is through a well specified standard. Downsides are:
- Needs some json processing
- Needs validation to check the assumption that there is only one manifest in the output (oci-layout allows for more).

gzm0:
That is correct, but only under the assumption that users always pass the correct rules forward: Without a provider, all there is to validate at analysis time is that the output is a single tree artifact.

For any more serious ruleset, I would not be comfortable just assuming the build user has wired everything correctly (for comparison: with a provider, I only need to trust the rule implementor).

gzm0:
Another datapoint: Turns out I needed the config digest, not the manifest digest. IIUC with the OIC layout it is infeasible to predict where it is going to end up (since it is content addressable). So to use it, the call-site needs to handle OIC layout anyways.

thesayyn:
[...]

PR #646, opened 2024-07-03
fix: propagate transitive inputs

thesayyn:
This has surfaced as bug in most of the places like https://github.com/bazel-contrib/rules_oci/issues/643, https://github.com/bazel-contrib/rules_oci/issues/642 (most likely). 

The problem was that when symlinks were enabled on Bazel 7, the logic did not propagate the file dependencies correctly downstream which lead to some obscure errors i have mentioned above. 

Now that we have some provider hierarchy, it makes me think if we should have a provider here, as requested in:  https://github.com/bazel-contrib/rules_oci/issues/279

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -171,7 +171,7 @@ def _oci_image_impl(ctx):
         args.add(ctx.file.base.path, format = "--from=%s")
         inputs.append(ctx.file.base)
         if use_symlinks:
-            transitive_inputs.append(ctx.file.base)
+            transitive_inputs += [ctx.file.base] + ctx.attr.base[DefaultInfo].default_runfiles.files.to_list()
     else:
         # create a scratch base image with given os/arch[/variant]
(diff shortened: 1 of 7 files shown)


oci/private/image.bzl
```diff
@@ -171,7 +171,7 @@ def _oci_image_impl(ctx):
         args.add(ctx.file.base.path, format = "--from=%s")
         inputs.append(ctx.file.base)
         if use_symlinks:
-            transitive_inputs.append(ctx.file.base)
+            transitive_inputs += [ctx.file.base] + ctx.attr.base[DefaultInfo].default_runfiles.files.to_list()
```
alexeagle, line 174: I think you can avoid a `to_list` here since `ctx.actions.run#inputs` accepts a depset

alexeagle (approved):

oci/private/image.bzl
thesayyn: right, fixed it now.

PR #652, opened 2024-07-17
Add oci_push label to support running push script on host

Lev1ty:


thesayyn:
Could you please provide the rationale behind this change? I thought we fixed oci_push for good.  

I'd like to get @EdSchouten thoughts on this.

EdSchouten:
I don't think a change along these lines is necessary. Are you by any chance trying to do something like this?

```
bazel run --platforms=//platform/of/the/resulting/image //the/push/target
```

If so, that's invalid. `--platforms` should in this case correspond to the platform on which the push target is being run. Namely, the host platform. If you want to push an image intended for another platform, please use `platform_transition_filegroup()`:

https://github.com/buildbarn/bb-storage/blob/master/tools/container.bzl#L37-L41

Lev1ty:
My workflow builds the image for Linux (target) and pushes the image to a docker repo, all the while running on my macOS laptop (host). Before the latest release, things worked fine. Afterwards, the push script was built for target instead of host. So I patched it back to transition to host platform in my repo, and thought others could benefit from the flexibility.

I'm using toolchains llvm as extra toolchains, and setting platforms to Linux. If possible I would not want to build a multi-arch image and keep the existing workflow.

Would this make a case for the flag addition? Or it would fall under your proposed file transition group?

EdSchouten:
> If possible I would not want to build a multi-arch image and keep the existing workflow.

There's no need to build a multi-arch image. The code I linked is just an example of how to do such a transition. Whether you end up using `oci_image_index()` is completely your choice.

Not that I'm in any way an owner of rules_oci, I would vote against an addition like this. Sure, Bazel's overall handling of platforms could sure use a bit more love. But as it stands, this change ends up using platforms incorrectly.

If you do a `bazel build //oci/push/target`, you are not building a container image. You are building a tool that is capable of pushing a container image. Even though you typically want to run this tool on the same system as where you run Bazel, this is by no means a requirement.

If all you're looking for is a way where you can have a single push target, but still want to have the freedom to specify the platform of the image on the command line, you'd better use a [`string_flag()`](https://bazel.build/extending/config#predefined-settings) or something.

Lev1ty:
Let me show the relevant code involved. I believe I am using platforms flag correctly, setting it as the cross-compilation target platform. Upon re-examining the error, it may be an issue with the binary compatibility of `yq`.

With the patch, the push succeeds, and the web app runs fine on the linux-amd64 server, so I know the image is correctly targeting linux-amd64.

If I comment out the `single_version_override` applying the patch, this is the new error in 1.8.0:

```bash
~/Projects/l3> bazel run //prod/versor:push --config=prod; fly deploy --config=prod/versor/fly.toml                                                                                              2024-07-20 12:06:38 PM
INFO: Invocation ID: fe78459f-d3e8-4561-b2f6-36cc760a52f3
WARNING: Build options --compilation_mode, --platforms, and --strip have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
INFO: Analyzed target //prod/versor:push (632 packages loaded, 43258 targets configured).
INFO: Found 1 target...
Target //prod/versor:push up-to-date:
  bazel-bin/prod/versor/push_push.sh
INFO: Elapsed time: 37.914s, Critical Path: 32.05s
INFO: 838 processes: 23 disk cache hit, 737 internal, 76 darwin-sandbox, 2 local.
INFO: Build completed successfully, 838 total actions
INFO: Running command line: bazel-bin/prod/versor/push_push.sh
/private/var/tmp/_bazel_lev1ty/de89f262b469edfa61610312bc506b3e/execroot/_main/bazel-out/darwin_arm64-opt/bin/prod/versor/push_push.sh: line 47: ../aspect_bazel_lib~~toolchains~yq_linux_amd64/yq: cannot execute binary file
```

## Code

```bazel
# MODULE.bazel
bazel_dep(name = "rules_oci", version = "1.8.0")
single_version_override(module_name = "rules_oci", version = "1.8.0", patches = ["//:patches/rules_oci_1_8_0.patch"])
```

```patch
# patches/rules_oci_1_8_0.patch
--- oci/private/push.bzl
+++ oci/private/push.bzl
@@ -114,1 +114,1 @@
-    inputs = ["//command_line_option:platforms"],
+    inputs = ["//command_line_option:host_platform"],

--- oci/private/push.bzl
+++ oci/private/push.bzl
@@ -108,1 +108,1 @@
-            for platform in settings["//command_line_option:platforms"]
+            for platform in [settings["//command_line_option:host_platform"]]

```

```bash
# .bazelrc
build --extra_toolchains=@llvm_toolchain_with_sysroot//:cc-toolchain-aarch64-linux
build --extra_toolchains=@llvm_toolchain_with_sysroot//:cc-toolchain-x86_64-linux

build:release --compilation_mode=opt
build:release --strip=always

build:staging --config=release
build:staging --platforms=@toolchains_llvm//platforms:linux-aarch64

build:prod --config=release
build:prod --platforms=@toolchains_llvm//platforms:linux-x86_64
```

```bazel
# prod/versor/BUILD.bazel
load("@rules_oci//oci:defs.bzl", "oci_image", "oci_push", "oci_tarball")
load("@rules_pkg//pkg:tar.bzl", "pkg_tar")

pkg_tar(
    name = "tar_versor",
    srcs = ["//go/cmd/versor"],
)

pkg_tar(
    name = "tar_env",
    srcs = [".env"],
)

oci_image(
    name = "image",
    base = "@distroless_static",
    entrypoint = ["/versor"],
    env = {"ENV": "/.env"},
    tars = [
        ":tar_versor",
        ":tar_env",
    ],
)

oci_tarball(
    name = "tarball",
    image = ":image",
    repo_tags = ["versor:latest"],
)

oci_push(
    name = "push",
    image = ":image",
    remote_tags = ["latest"],
    repository = "[REDACTED]",
)
```

EdSchouten:
Your bazelrc is missing something along the lines of:
```
run --platforms=@bazel_tools//tools:host_platform
```

Once you add that, it will end up using a copy of yq and crane for the correct platform.

Then the remaining issue is that your container image will be built for the host platform. This can be addressed by adding that `platform_transition_filegroup()`.

Lev1ty:
Thank you I'll try that and post the code here for posterity.

Typically platform flag expresses the cross-compilation target, so setting it to host platform does seem off somewhat. If the run action defaults to host machine, the platform flag can be set to the semantically correct cross-compilation target instead of the pragmatic host platform. Can consider this a sharp edge of bazel

EdSchouten:
> Typically platform flag expresses the cross-compilation target, so setting it to host platform does seem off somewhat.

If you think about it, it’s the only **valid** choice. You’re using Bazel to build something, so that you can run it on your local system. That in this case it happens to be a tool that pushes a container image for Linux is irrelevant. The Linux container image is just an asset that is used by the tool that does the pushing.

Lev1ty:
I see, in that perspective it makes more sense. The issue here is I want part of the dependency tree to target Linux and the rest to target local. The platform flag uniformly sets the platform for the whole dep tree, so in cases like this we would have to use transitions

EdSchouten:
Using a transition is unavoidable here. If you look at the runfiles directory for your push target, you can see it contains assets for multiple platforms. So some part of the build graph needs to make that switch.

Lev1ty:
Works perfectly with `platform_transition_filegroup`. Will close this PR now. The code example is below.

The command from before works without patching with `single_version_override`.

## Code

```bash
# .bazelrc
build --extra_toolchains=@llvm_toolchain_with_sysroot//:cc-toolchain-aarch64-linux
build --extra_toolchains=@llvm_toolchain_with_sysroot//:cc-toolchain-x86_64-linux

build:prod --compilation_mode=opt
build:prod --strip=always
```

```bazel
# prod/versor/BUILD.bazel
load("@aspect_bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
load("@rules_oci//oci:defs.bzl", "oci_image", "oci_push", "oci_tarball")
load("@rules_pkg//pkg:tar.bzl", "pkg_tar")

pkg_tar(
    name = "tar_versor",
    srcs = ["//go/cmd/versor"],
)

pkg_tar(
    name = "tar_env",
    srcs = [".env"],
)

oci_image(
    name = "image",
    base = "@distroless_static",
    entrypoint = ["/versor"],
    env = {"ENV": "/.env"},
    tars = [
        ":tar_versor",
        ":tar_env",
    ],
)

platform_transition_filegroup(
    name = "image_preprod",
    srcs = [":image"],
    target_platform = "@toolchains_llvm//platforms:linux-aarch64",
)

platform_transition_filegroup(
    name = "image_prod",
    srcs = [":image"],
    target_platform = "@toolchains_llvm//platforms:linux-x86_64",
)

oci_tarball(
    name = "tarball",
    image = ":image_preprod",
    repo_tags = ["versor:latest"],
)

oci_push(
    name = "push",
    image = ":image_prod",
    remote_tags = ["latest"],
    repository = "[REDACTED]",
)
```

Issue #645, opened 2024-07-03
rules_oci@v2beta1 - Image not compatible with AWS lambdas


juanzolotoochin:
AWS allows to create lambdas using container image, see https://docs.aws.amazon.com/lambda/latest/dg/images-create.html.

We are seeing the following error when trying to run a lambda built with version 2beta1 of `rules_oci`: 

```
"ImageLayerFailure: UnsupportedImageLayerDetected - Layer Digest sha256:dd2d1b3af2eef813d8cf22b1e17c4500e6c34fd80eddb0f0a78e80632fc872da"
```

The exact same image works fine when using `rules_oci` v1.

thesayyn:
Could be a bug. Do you have a reproducer for this?

juanzolotoochin:
I thought I'd added this to this bug but it was only on the slack channel.

It's hard to give repro instructions because I don't understand the AWS infra enough and this part is buried under a ton of TF code from me.

But after inspecting the two images (rules_oci v1 vs v2) I suspect the problem comes from the missing history entries for the docker layers.  The v2 image does not have `history` entries for any of the tar layers.

```  "architecture": "amd64",
  "author": "github.com/chainguard-dev/apko",
  "created": "2024-06-28T05:42:58Z",
  "history": [
    {
      "author": "apko",
      "created": "2024-06-28T05:42:58Z",
      "created_by": "apko",
      "comment": "This is an apko single-layer image"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    }
  ],
  "os": "linux",
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:8952da8c429f5b89a61500dc9b5e52ce326da208332a8c16f8e51b0563bf53f7",
      "sha256:dd2d1b3af2eef813d8cf22b1e17c4500e6c34fd80eddb0f0a78e80632fc872da",
      "sha256:45aca4f0d57586f6ab7e6c7b9104204b93f02a4afef6e0f5ffaad4ac003440e8"
    ]
  },``` 
  
I suspect AWS assumes that `len(history)` == `len(layers`) somewhere. Ideally this history would contain useful metadata as sugested by https://github.com/bazel-contrib/rules_oci/pull/395 but even if not that, it would be great to have something consistent with `rules_oci v1` at least.

thesayyn:
[...]

Issue #438, opened 2023-12-15
Created time is showing as 0001-01-01T00:00:00Z when it should be 1970-01-01T00:00:00Z


prestonvanloon:
Not sure how to best explain this issue, but we are seeing layers uploaded with an invalid creation time when it is expected to be unix epoch 0 (`1970-01-01T00:00:00Z`)

Example of an image created with rules_oci and the history is all invalid timestamps

```
crane config gcr.io/prysmaticlabs/prysm/beacon-chain:HEAD-344e68 | jq . 
{
  "architecture": "amd64",
  "created": "0001-01-01T00:00:00Z",
  "history": [
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    },
    {
      "created": "0001-01-01T00:00:00Z"
    }
  ],
  "os": "linux",
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:54ad2ec71039b74f7e82f020a92a8c2ca45f16a51930d539b56973a18b8ffe8d",
      "sha256:6fbdf253bbc2490dcfede5bdb58ca0db63ee8aff565f6ea9f918f3bce9e2d5aa",
      "sha256:7bea6b893187b14fc0a759fe5f8972d1292a9c0554c87cbf485f0947c26b8a05",
[...]

PR #654, opened 2024-07-18
fix: add history entries for tars attribute

thesayyn:
image-spec says that this field is optional, however some systems like seen in https://github.com/bazel-contrib/rules_oci/issues/645 require this field. Also, skipping history entries lead to broken images if the base image already had one. 

fixes https://github.com/bazel-contrib/rules_oci/issues/645

### DIFF
--- a/oci/private/image.bzl
+++ b/oci/private/image.bzl
@@ -105,6 +105,7 @@ def _calculate_descriptor(ctx, idx, layer, zstd, jq, coreutils, regctl):
     args = ctx.actions.args()
     args.add(layer)
     args.add(descriptor)
+    args.add(layer.owner)
     ctx.actions.run(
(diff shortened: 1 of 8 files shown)


thesayyn:
This partially fixes https://github.com/bazel-contrib/rules_oci/issues/438 too.

oci/private/image.bzl
```diff
@@ -105,6 +105,7 @@ def _calculate_descriptor(ctx, idx, layer, zstd, jq, coreutils, regctl):
     args = ctx.actions.args()
     args.add(layer)
     args.add(descriptor)
+    args.add(layer.owner)
```
alexeagle, line 108: confused how this is related to the PR, is it just an unrelated drive-by fix?

alexeagle (approved):

PR #657, opened 2024-07-22
fix typo in load.sh.tpl

kopp:
Currently the generated template looks like

```bash
if [ -e "$LOADER}" ]; then
    CONTAINER_CLI="$LOADER"
elif command -v docker &> /dev/null; then
    CONTAINER_CLI="docker"
# ...
```

This will never evaluate the first if clause, because `$LOADER}` will for an empty `LOADER` expand to `}`.


kopp:
@thesayyn  could you please have a brief look at this super-mini PR? :)

thesayyn:
Hmm, this seems to be failing. i guess we need to check if the loader is a file.

Issue #659, opened 2024-07-25
2.0.0beta1: Entrypoints with comma separated values are not correctly built


BillyBlaze:
When you have an entrypoint that requires a comma separated value then the image is not correctly built:

Example:
```
oci_image(
    base = "@wiremock",
    entrypoint = [
        "/docker-entrypoint.sh",
        '--permitted-system-keys="MY_HOST_V1,MY_HOST_V2"',
    ],
)
```

Outcome (with docker inspect):
```
"Entrypoint": [
  "/docker-entrypoint.sh",
  "--permitted-system-keys=\"MY_HOST_V1",
  "MY_HOST_V2\"",
]
```

I have tried to put the entrypoint within single quotes, double quotes, no qoutes and escaping the comma, but all lead to the same outcome.

Issue #631, opened 2024-06-19
Multiline cmd and entrypoint handling


alexnovak:
Hello!

I noticed that oci_image seems to struggle when it's given a multiline cmd or entrypoint input. When an array containing a multiline input is provided, eg

```
[
  "/bin/bash",
  "-c",
  """
  echo hello
  echo goodbye
  """
]
```

`oci_image` will truncate the input on the newline while providing it to crane, either causing the mutation to fail, or the command/entrypoint to render inaccurately. 

`crane` expects that the cmd or entrypoint is delivered as a CSV, following [rfc4180](https://www.rfc-editor.org/rfc/rfc4180#section-2), and the newlines in cmd or entrypoint can be presented within a quoted entry.

It looks like the current script would interpret each newline in the cmd or entrypoint file as a separate --cmd entry for the mutate command. https://github.com/bazel-contrib/rules_oci/blob/main/oci/private/image.sh#L105
My understanding is that crane only expects a single cmd or entrypoint, is there a reason we handle it in this read loop?

I'd imagine we instead want to:
1. Cat the contents of the entrypoint and cmd file as the inputs to the corresponding `--cmd` or `--entrypoint`
2. When gathering the cmd/entrypoint lines for the file, ensure that we quote them, and escape all double quotes according the above rfc (`""`). 

Happy for any thoughts here, would love to understand if I'm offbase, or there's another way to do this.

alexnovak:
Ah, seeing this is directly referenced as a consequence here https://github.com/bazel-contrib/rules_oci/pull/297#issuecomment-1708249851

Apologies if this is answered elsewhere, but is there a reason newline delimited was the decision over comma delimited?

I threw together a quick diff with what changes I think would be necessary here https://github.com/bazel-contrib/rules_oci/compare/main...alexnovak:rules_oci:main - but it fails the location_expansion test because of this convention.

thesayyn:
> but is there a reason newline delimited was the decision over comma delimited?

i don't think so, any character would have the same consequences, as we no longer use crane.  i suggest we fix this in 2.x and leave 1.x the way it is.

alexnovak:
Sounds fine by me! I'll double check how things work on the 2.0 branch and submit a PR

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/663

PR #663, opened 2024-07-26
fix: comma in entrypoint and cmd

thesayyn:
Closes https://github.com/bazel-contrib/rules_oci/issues/659 and https://github.com/bazel-contrib/rules_oci/issues/631

Reason why i changed the delimiter from `,` to `\n`

- Easier to create a file in a genrule,  simply `echo` without the `-n` makes it work.
- We no longer use crane so don't need to conform to their  cmd and entrypoint must be a valid csv


thesayyn:
cc @mrmeku

oci/private/image.bzl
```diff
@@ -69,8 +69,8 @@ _attrs = {
         The authors recommend [dive](https://github.com/wagoodman/dive) to explore the layering of the resulting image.
     """),
     # See: https://github.com/opencontainers/image-spec/blob/main/config.md#properties
-    "entrypoint": attr.label(doc = "A file containing a comma separated list to be used as the `entrypoint` to execute when the container starts. These values act as defaults and may be replaced by an entrypoint specified when creating a container. NOTE: Setting this attribute will reset the `cmd` attribute", allow_single_file = True),
-    "cmd": attr.label(doc = "A file containing a comma separated list to be used as the `command & args` of the container. These values act as defaults and may be replaced by any specified when creating a container.", allow_single_file = True),
+    "entrypoint": attr.label(doc = "A file containing a newline separated list to be used as the `entrypoint` to execute when the container starts. These values act as defaults and may be replaced by an entrypoint specified when creating a container. NOTE: Setting this attribute will reset the `cmd` attribute", allow_single_file = True),
```
alexeagle, line 72: nit: do we need to be specific about line feed and check this on Windows?

alexeagle (approved):

oci/private/image.bzl
thesayyn: currently state of windows is unknown. I'd imagine supporting it would be as easy as using the platform specific \n, but that's not important at the moment.

PR #666, opened 2024-08-04
chore: remove deprecated config_path attribute

thesayyn:
Removes the deprecated config_path attribute. Replacement is to use `config` attribute which accepts a label, or use `DOCKER_CONFIG` standard env for cases such as overriding the config in CI.


thesayyn:
@alexeagle PTAL

alexeagle (approved):

PR #673, opened 2024-08-16
make `oci_image()` accept a file for workdir

lazcamus:
This clones much of the same logic as `entrypoint` and friends. Also refactor docstring a little to illuminate "wrapper vs the rule" and the dual typed args.

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -85,7 +85,7 @@ oci_image(
 | <a id="oci_image_rule-user"></a>user |  The <code>username</code> or <code>UID</code> which is a platform-specific structure that allows specific control over which user the process run as. This acts as a default value to use when the value is not specified when creating a container. For Linux based systems, all of the following are valid: <code>user</code>, <code>uid</code>, <code>user:group</code>, <code>uid:gid</code>, <code>uid:group</code>, <code>user:gid</code>. If <code>group/gid</code> is not specified, the default group and supplementary groups of the given <code>user/uid</code> in <code>/etc/passwd</code> from the container are applied.   | String | optional | <code>""</code> |
 | <a id="oci_image_rule-variant"></a>variant |  The variant of the specified CPU architecture. eg: <code>v6</code>, <code>v7</code>, <code>v8</code>. See: https://github.com/opencontainers/image-spec/blob/main/image-index.md#platform-variants for more.   | String | optional | <code>""</code> |
 | <a id="oci_image_rule-volumes"></a>volumes |  A file containing a comma separated list of volumes. (e.g. /srv/data,/srv/other-data)   | <a href="https://bazel.build/concepts/labels">Label</a> | optional | <code>None</code> |
-| <a id="oci_image_rule-workdir"></a>workdir |  Sets the current working directory of the <code>entrypoint</code> process in the container. This value acts as a default and may be replaced by a working directory specified when creating a container.   | String | optional | <code>""</code> |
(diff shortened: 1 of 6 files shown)


lazcamus:
I used this to clone the language rule's "set a default `workdir` for language targets" logic [from `rules_docker`](https://github.com/bazelbuild/rules_docker/blob/3040e1fd74659a52d1cdaff81359f57ee0e2bb41/lang/image.bzl#L209)

My intent is to extract some of my local changes and add default `entrypoint` and `workdir` setting to https://github.com/aspect-build/bazel-examples/tree/main/oci_python_image

thesayyn:
I don’t think it makes sense to have an attribute that takes an array for a single value. Doesn’t make sense from public api standpoint.

jjmaestro:
@lazcamus I love the idea to be able to use the output of a target by just passing the label but I agree with @thesayyn re. not using an array. Also, per the Dockerfile spec, `WORKDIR` only has a single argument. IMHO it maps much better to just one string.

Following your notation (which I also think is great and clarifies the wrapper-vs-rule) I'd prefer if `workdir` could be `STR_OR_LABEL`. I think that brings in all the goodies from your PR and still maintains a more natural API that's aligned with the Dockerfile spec.

lazcamus:
On Sun, Sep 1, 2024 at 2:22 AM Sahin Yort ***@***.***> wrote:

> I don’t think it makes sense to have an attribute that takes an array for
> a single value. Doesn’t make sense from public api standpoint.
>

Makes sense. I'll up my skylark skills and re-work it :)

lazcamus:
I pushed a commit that makes `workdir` a `STRING_OR_LABEL`, so PTAL

The reason I originally ended up on the list is b/c it's easy to check with `types` in a macro, and everybody else was using lists so it kinda blended in 😆

There is no "is this string a valid label?" check in macro context, so I did some hacky string matching.

It comes with the corner case of: what do you do if you get "foo" ? It's a valid target (I'd use `":foo"` but not everybody is me). "foo" is also a valid relative `workdir` (I also only use absolute `workdir`s). I chose label for this weird in-between case, but could see it either way. A workaround if you want "foo" to be a relative workdir string is to use `"./foo"`, which seems OK 🤷

lazcamus:
CI status seems like a runner failure, hopefully intermittent?
```
ERROR: /home/runner/work/rules_oci/rules_oci/examples/dockerfile/BUILD.bazel:18:11: BuildDocker examples/dockerfile/base [for tool] failed: (Exit 1): buildx failed: error executing BuildDocker command (from target //examples/dockerfile:base) bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/examples/dockerfile/buildx build ./examples/dockerfile --builder builder-docker ... (remaining 1 argument skipped)
error: dial unix /run/buildkit/buildkitd.sock: connect: no such file or directory
ERROR: listing workers for Build: failed to list workers: Unavailable: connection error: desc = "error reading server preface: EOF"
INFO: Elapsed time: 20.426s, Critical Path: 10.96s
INFO: 363 processes: 308 internal, 40 linux-sandbox, 15 local.
ERROR: Build did NOT complete successfully
```

thesayyn:
i am afraid only possible solution here is to allow string only from the macro since we can't distinguish between string vs label despite best efforts. I believe it was one of the big mistakes of bazel not making it apparently obvious between string vs label from use-site perspective.

thesayyn:
I am going to close this PR as there is no easy way to tell between label and a string

PR #676, opened 2024-08-19
refactor: remove mtree as the default output

thesayyn:
BREAKING CHANGE:

It's unfortunate that we make a breaking change in an RC phase, but we haven't seen anybody using this output so far. 

Removal of the mtree as the default output allows `$(location :oci_load_label)` to return the `executable` instead of the `DefaultInfo.files[0]` which what most expect.

### DIFF
--- a/oci/private/load.bzl
+++ b/oci/private/load.bzl
@@ -218,7 +218,6 @@ def _load_impl(ctx):
 
     return [
         DefaultInfo(
-            files = depset([mtree_spec]),
             runfiles = runfiles,
             executable = runnable_loader,
         ),


oci/private/load.bzl
```diff
@@ -218,7 +218,6 @@ def _load_impl(ctx):
 
     return [
         DefaultInfo(
-            files = depset([mtree_spec]),
```
alexeagle, line 221: Is there any reason to consider adding another output group to reference output?

alexeagle (approved):

oci/private/load.bzl
thesayyn: I don't see any benefit for now. That can be added later as a feature if anybody asks for it

PR #681, opened 2024-08-25
chore(docs): mention rules_multirun

alexeagle:
We get this question a lot so the answer belongs in the docs.


alexeagle:
The failed test is present on main as well :(

thesayyn (approved):

PR #682, opened 2024-08-26
fix: include examples

thesayyn:
See: https://github.com/bazelbuild/bazel-central-registry/pull/2563#issuecomment-2310972670


alexeagle (approved):
This is going to cause KILOBYTES of data transfer. LGTM anyway 😉

alexeagle:
Those CI failures are present at HEAD too

thesayyn:
> Those CI failures are present at HEAD too

fixed those.

PR #702, opened 2024-09-21
Add example of building a Dockerfile with local tar file build context

njlr:
This PR adds an example of using a Dockerfile with a local tar as a build context. 

This is not the ideal approach, which this is clearly signposted in the README!, but it shows an easy migration path from `rules_docker`. The existing example does not show how Bazel build artefacts can be passed to BuildX. 

How it works:

 1. Create a tar file using `rules_pkg`
 2. Pipe the tar file into BuildX using `sh_binary` and `run_binary`
 3. Use the BuildX output as a base for an `oci_image`

This is a replacement for `container_run_and_commit`. 

The example could easily be adapted to behave like `container_run_and_extract` by changing the `--output` flag to a `tar`.


thesayyn:
We have https://github.com/bazel-contrib/rules_oci/tree/main/examples/dockerfile, how is this different than that one?

njlr:
> We have https://github.com/bazel-contrib/rules_oci/tree/main/examples/dockerfile, how is this different than that one?

That example does not show how to use the output of other Bazel targets as input to the Dockerfile build

thesayyn:
> That example does not show how to use the output of other Bazel targets as input to the Dockerfile build

Than can we change it so it does?  i don't see the difference between this example and the other one, other than putting the srcs into a tar file beforehand so you don't have to deal with paths that start with `bazel-bin/cfg/bin`

njlr:
> > That example does not show how to use the output of other Bazel targets as input to the Dockerfile build
> 
> Than can we change it so it does? i don't see the difference between this example and the other one, other than putting the srcs into a tar file beforehand so you don't have to deal with paths that start with `bazel-bin/cfg/bin`

Yes, happy to do this.

The reason I made a new example is that it's a little more complicated than the existing example, so might hinder understanding. 

What do you think?

thesayyn:
We can add some comments explaining why there is two different buildx targets in the same package and the difference.

njlr:
> We can add some comments explaining why there is two different buildx targets in the same package and the difference.

I have merged the two examples

thesayyn (approved):
thanks!

examples/dockerfile/BUILD.bazel
```diff
@@ -87,3 +89,85 @@ assert_oci_image_command(
            ||     ||
 """,
 )
+
+# This tar can be arranged in arbitrary ways.
+# For example, one could grab pkg_files from elsewhere in the workspace.
+# Anything in the tar can be used by a Dockerfile ADD or COPY.
+pkg_tar(
```
thesayyn, line 96: actually can we not use pkg_tar here? we have bazel-lib dependency here already. you can use https://github.com/bazel-contrib/bazel-lib/blob/main/docs/tar.md

(i don't want a pkg_tar dependency even if it's dev_dep)

thesayyn:
There are some failures, can you take a look?

examples/dockerfile/BUILD.bazel
njlr: I might need some assistance with this one.

When I try `bazel build //examples/dockerfile:buildx_context` I get:

```
tar: Unrecognized archive format
tar: Error exit delayed from previous errors.
Target //examples/dockerfile:buildx_context failed to build
```

Any ideas?

examples/dockerfile/BUILD.bazel
```diff
@@ -87,3 +88,95 @@ assert_oci_image_command(
...
+    ] + glob([
+        "src/*",
+    ]),
+)
+
+mtree_mutate(
+    name = "buildx_context_tree_prefixed",
+    mtree = ":buildx_context_tree",
+    strip_prefix = package_name(),
+    package_dir = "/app",
```
thesayyn, line 109: you need a `./app` to fix the issue.

examples/dockerfile/BUILD.bazel
thesayyn: See my comment above.

examples/dockerfile/BUILD.bazel
njlr: Thanks - this works

One thing I found a little surprising is that I need to add the `srcs` for both the `mtree_spec` and the `tar`. 

Is this expected? Am I using it correctly?

thesayyn:
Sorry, i dropped the ball on this one.  Are you still interested in this?

njlr:
> Sorry, i dropped the ball on this one. Are you still interested in this?

Yes I still think it's useful for people.

Can update to `main` soon 👍

PR #709, opened 2024-10-04
fix: pass global args to all crane invocations

### LINKED ISSUE #527: oci_push with remote_tags does not work with insecure repositories
When migrating from container_push to oci_push with a repository that fails to be detected as insecure, it appears the alternative for insecure_repository is to specify `args = ["--insecure"]` ([taken from this website](https://docs.aspect.build/guides/rules_oci_migration/#container_push)). This works like a charm if you're simply pushing, but if you combine this with the `remote_tags` attribute, the request to add tags will fail.

Example Bazel target
```
oci_push(
    name = "engine.push",
    args = ["--insecure"],
    image = ":engine",
    remote_tags = ":version_tag",
    repository = "k3d-registry.localhost:9999/engine",
)
```
will fail with the following exception
```
Error: fetching "k3d-registry.localhost:9999/engine@sha256:774d75357a396d3659774d71754130646e517874444a4b6d7355376d68723065": Get "https://k3d-registry.localhost:9999/v2/": http: server gave HTTP response to HTTPS client
```

This is due to [args being passed to the push command](https://github.com/bazel-contrib/rules_oci/blob/8f64741dc00d75248f9af014e655b6ca4e2e59e5/oci/private/push.sh.tpl#L48) but not the [subsequent tags commands](https://github.com/bazel-contrib/rules_oci/blob/8f64741dc00d75248f9af014e655b6ca4e2e59e5/oci/private/push.sh.tpl#L50-L57).

If I patch push.sh.tpl to pass ARGS into the tags command as well, then the bazel target above works as expected. I'm wondering if this is the proper solution, or if there's a reason ARGS shouldn't be passed to the tags command. If that is the case, is there an alternative to tagging with an insecure repository?

ahans:
This fixes #527. Please let me know if I missed anything. Thanks!

### DIFF
--- a/oci/private/push.sh.tpl
+++ b/oci/private/push.sh.tpl
@@ -52,13 +52,13 @@ done
 DIGEST=$("${JQ}" -r '.manifests[0].digest' "${IMAGE_DIR}/index.json")
 
 REFS=$(mktemp)
-"${CRANE}" push ${VERBOSE} "${IMAGE_DIR}" "${REPOSITORY}@${DIGEST}" "${ARGS[@]+"${ARGS[@]}"}" --image-refs "${REFS}"
+"${CRANE}" push ${VERBOSE} "${IMAGE_DIR}" "${REPOSITORY}@${DIGEST}" "${ARGS[@]}" --image-refs "${REFS}"
 
 for tag in "${TAGS[@]+"${TAGS[@]}"}"
 do
-  "${CRANE}" tag ${VERBOSE} $(cat "${REFS}") "${tag}"
+  "${CRANE}" tag ${VERBOSE} "${ARGS[@]}" $(cat "${REFS}") "${tag}"
 done
 
 if [[ -e "${TAGS_FILE:-}" ]]; then
-  cat "${TAGS_FILE}" | xargs -r -n1 "${CRANE}" tag ${VERBOSE} $(cat "${REFS}")
+  cat "${TAGS_FILE}" | xargs -r -n1 "${CRANE}" tag ${VERBOSE} "${ARGS[@]}" $(cat "${REFS}")
 fi


oci/private/push.sh.tpl
```diff
@@ -52,13 +52,13 @@ done
 DIGEST=$("${JQ}" -r '.manifests[0].digest' "${IMAGE_DIR}/index.json")
 
 REFS=$(mktemp)
-"${CRANE}" push ${VERBOSE} "${IMAGE_DIR}" "${REPOSITORY}@${DIGEST}" "${ARGS[@]+"${ARGS[@]}"}" --image-refs "${REFS}"
```
ahans, line 55: This change was not strictly necessary, but since we have `ARGS=()` above, the `+` check seems redundant.

thesayyn:
this change looks like a breaking change, before args are passed only to crane push, but now args are passed to every subcommand. 

principled fix for this is to only pass global flags to all by plucking them out of args first.

oci/private/push.sh.tpl
ahans: What exactly would happen with an empty array with `"${ARGS[@]}"`? Wouldn't it just evaluate to the empty string? But I can take it back for sure.

thesayyn: depends on flavor of bash, old bash implementations complain, i don't remember exact error message.

ahans:
> this change looks like a breaking change, before args are passed only to crane push, but now args are passed to every subcommand.
> 
> principled fix for this is to only pass global flags to all by plucking them out of args first.

I see, so you only want to pass on the following flags?
```bash
Global Flags:
      --allow-nondistributable-artifacts   Allow pushing non-distributable (foreign) layers
      --insecure                           Allow image references to be fetched without TLS
      --platform platform                  Specifies the platform in the form os/arch[/variant][:osversion] (e.g. linux/amd64). (default all)
  -v, --verbose                            Enable debug logs
```

TBH, I found it a bit surprising that one is supposed to use `args = ["--insecure"]` to support repositories whose certificate can't be validated. I would prefer an explicit flag such as `insecure_repository` as we had with `rules_docker`. I'm not sure if it's such a good idea to allow the user to pass flags to `crane`. Isn't usage of `crane` an implementation detail here? Anyway, I'm happy to implement the approach above that filters out global flags, but would like to understand the motivation for this design first. What are other usecases for passing in `args`?

thesayyn:
`args` something bazel adds support for runnable rules by default and there is no way to disable it. Also, crane is not implementation detail for oci_push unfortunately, we have supported passing arbitrary args to `oci_push` since the beginning of rules_oci.

thesayyn:
> I would prefer an explicit flag such as `insecure_repository` as we had with `rules_docker`

Not every flag deserves its own attribute, if it translates 1:1 to the underlying tool, if we did support something like this then there would be two different way of getting the same outcome, eg: `args` and `insecure_repository` and we'd have to maintain both and resolve conflicts if `args = ["--insecure"]` and `insecure = True|False`. I'd rather keep it simple and just pass all the global flags to all crane invocations.

ahans:
Ok, I think disabling `args` support would be the cleanest solution here. But since that's not possible and we want to maintain backwards compatibility, allowing users to pass everything via `args` makes sense. Thanks for the explanation, that helps a lot!

I updated the PR now to put all global flags into a `GLOBAL_FLAGS` list and pass that separately to each `crane` invocation. Since `--verbose` (or `-v`) is just one of those special flags, I removed the special handling for that. I hope it's better now! Looking forward to your review!

thesayyn (approved):

ahans:
Anything I can do about the test failure? Looks like some infrastructure issue ...

ahans:
Anything I can do about the test failure? Looks like some infrastructure issue ...

PR #712, opened 2024-10-09
fix: Support loading tarballs with older podman

### LINKED ISSUE #711: oci_load incompatible with older version of Podman
I'm running Bazel v7.2.0, rules_oci v2.0.0 on an Ubuntu 22.04.5 LTS system. I've installed the Ubuntu package for podman which is currently v3.4.4.

My build uses `oci_load` to build and import containers into podman:

```
oci_load(
    name = "foobar_tarball",
    image = ":foobar_image",
    repo_tags = ["foobar:latest"],
)
```

The resulting script fails to run:

```
$ bazel-out/aarch64-fastbuild/bin/foobar_tarball.sh
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Error: payload does not match any of the supported image formats (oci, oci-archive, dir, docker-archive)
tar: Write error
```

Looking into the issue, this appears to be a compatibility issue with how older versions of podman implement `podman load`. In particular the following two commands lines work differently:

```
# fails with "payload does not match any of the supported image formats"
podman load --input <(tar -c $FILES)
# succeeds and loads the image
tar -c $FILES | podman load  
```

This issue appears fixed in newer versions of podman, however I haven't tracked down the details.

The fix is relatively straightforward; to change the `podman load` command line [oci/private/load.sh.tpl](https://github.com/bazel-contrib/rules_oci/blob/a002dafcddffecf5a670adaea53bcb7a4ee4eeb5/oci/private/load.sh.tpl#L31-L33)

nholstein:
Older versions of Podman appear to handle loading a tarball differently depending upon with the input is /dev/stdin or not. In particular:

    podman load --input <(cat foo.tar)

raises an error, while:

    cat foo.tar | podman load

works without error. (Podman loads from stdin if no --input is specified.)

This enables support for podman v3.4.4 included in Ubuntu 22.04.

fixes #711

### DIFF
--- a/oci/private/load.sh.tpl
+++ b/oci/private/load.sh.tpl
@@ -28,6 +28,5 @@ mtree_contents="$(cat $MTREE)"
 mtree_contents="${mtree_contents//"$image_root"/}"
 mtree_contents="${mtree_contents//"$manifest_root"/}"
 
-"$CONTAINER_CLI" load --input <(
-    "$TAR" --cd "$RUNFILES_DIR/{{workspace_name}}" --create --no-xattr --no-mac-metadata @- <<< "$mtree_contents"
-)
+"$TAR" --cd "$RUNFILES_DIR/{{workspace_name}}" --create --no-xattr --no-mac-metadata @- <<< "$mtree_contents" | \
+    "$CONTAINER_CLI" load


thesayyn:
This is a breaking change unfortunately. It changes invocation of the underlying binary from `docker load -i` to `docker load | input`

nholstein:
@thesayyn, thanks for looking into this and you feedback. I wasn't aware that this was a breaking change under Docker; my understanding was that docker supports the same `tar $FILES | docker load` semantics.

Can you suggest an alternate approach I could investigate that would be compatible?

thesayyn:
It's a breaking change for rules_oci to change how we invoke the loader cli, it could podman or docker or some shell script user wrote. If we change it from `-i` to stdin, it's a breaking change.

thesayyn:
Okay i think we can't land this PR as is right now. We might consider landing this change on a 3.x branch though. So i'll close it for now.

Issue #617, opened 2024-06-08
FR: oci_tarball should have a `.digest` target that gives imageid which would match the daemon. 


thesayyn:
See: https://github.com/bazel-contrib/rules_oci/issues/60#issuecomment-2156130242

PR #715, opened 2024-10-11
feat: Output the image id of an oci_load target.

gonzojive:
Not the most elegant starlark, so this is a draft not meant for review, but it demos how to compute the image id.

Fixes https://github.com/bazel-contrib/rules_oci/issues/617


thesayyn:
Thank you, nice to see you again :) 

We'd want this to be subtarget that simply uses jq rule, also this is a breaking change since it adds more fields to defaultinfo which will break people who are using $(location) etc.

thesayyn:
hey @gonzojive do you want to finish this PR?

PR #722, opened 2024-10-17
Set `created` date based on `org.opencontainers.image.created` annotation

### LINKED ISSUE #661: oci_image should support setting `created` and stamp
https://github.com/bazel-contrib/rules_oci/issues/49

hanneskaeufler:
Fixes #661 

# Why?

Doing a `docker inspect --format='{{.Created}}' my/image:latest` on an
image built with `oci_image` previously always returned the static
beginning of the unix timestamp (1. January 1970 00:00:00). To be more
precise, that's what it returns for images that are built "from
scratch". When an image is built on a base, then the created date is set
to the created date of the base image.

Both are fine strategies for reproducible builds, but in stamped builds
that are eventually shipped/deployed, you probably want to set that time
to the actual build time.

Stamping is already supported, e.g. by setting annotations, which can
use stamp variables through a workspace status command.

Furthermore, the OCI image spec already [defines an annotation](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) to describe the time and date the image was built.
Which of course, when `docker load -i mytarball.tar` is dutifully
ignored by docker.

We can leverage this and, if the OCI annotation is provided, also set
the created date to that same value. When it isn't, we simply fall back
to the old behaviour.

### Further thoughts

I of course debated

* adding a separate attribute `created` on the `oci_image` rule itself
* Documentation? We could add some explanation to the `oci_image` macro, but it already talks about annotations and somehow I feel like a mention of a specific label and creation date is too detail-oriented? Happy to change my mind ...
* not caring about docker. I have to check my specific use case again, but it might be that AWS ECR metadata APIs already interpret read the OCI annotation over the `created` field? Not sure ...
* How do we add a test for this? Seems the `image.sh` script itself is not tested in isolation so far. I have various local incarnations but eventually landed on the diff test using the docker CLI client. Seems like that's not available on CI though 🤷🏽

### DIFF
--- a/e2e/smoke/workspace_status.sh
+++ b/e2e/smoke/workspace_status.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+BUILD_TIMESTAMP=${BUILD_TIMESTAMP:-$(date +%s)}
+BUILD_ISO8601=$(date -u -r "$BUILD_TIMESTAMP" +"%Y-%m-%dT%H:%M:%SZ")
+echo "BUILD_ISO8601 $BUILD_ISO8601"
--- a/oci/private/image.sh
+++ b/oci/private/image.sh
@@ -176,16 +176,19 @@ for ARG in "$@"; do
     CONFIG=$(jq --rawfile labels "${ARG#--labels=}" '.config.Labels += ($labels | split("\n") | map(select(. | length > 0)) | map(. | split("=")) | map({key: .[0], value: .[1:] | join("=")}) | from_entries)' <<<"$CONFIG")
     ;;
   --annotations=*)
+    annotation_key_vals=$(jq -n --rawfile annotations "${ARG#--annotations=}" \
+      '([($annotations | split("\n") | .[] | select(. != ""))] | map(. | split("=")) | map({key: .[0], value: .[1:] | join("=")}) | from_entries)')
+
     get_manifest |
-      jq --rawfile annotations "${ARG#--annotations=}" \
-      '.annotations += ([($annotations | split("\n") | .[] | select(. != ""))] | map(. | split("=")) | map({key: .[0], value: .[1:] | join("=")}) | from_entries)' |
+      jq --argjson annotations "$annotation_key_vals" '.annotations += $annotations' |
       update_manifest
 
     created_date="$(jq -r -n \
+      --argjson annotations "$annotation_key_vals" \
       --arg created_key "${OPENCONTAINERS_CREATED_KEY}" \
       --arg default_created_date "${DEFAULT_CREATED_DATE}" \
-      --rawfile annotations "${ARG#--annotations=}" \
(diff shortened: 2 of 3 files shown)


e2e/smoke/workspace_status.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+BUILD_TIMESTAMP=${BUILD_TIMESTAMP:-$(date +%s)}
+BUILD_ISO8601=$(date -u -r "$BUILD_TIMESTAMP" +"%Y-%m-%dT%H:%M:%SZ")
```
hanneskaeufler, line 4: Ah, that was macOS specific `date -u -r` ... would have to make that portable

oci/private/image.sh
```diff
@@ -27,7 +29,11 @@ function base_from_scratch() {
     layers: []
   }' | update_manifest
   # Create the image config when there is annotations
-  jq -n --argjson platform "$platform" '{created: "1970-01-01T00:00:00Z", config:{}, history:[], rootfs:{type: "layers", diff_ids:[]}} + $platform' | update_config >/dev/null
+  jq -n \
+    --argjson platform "$platform" \
+    --arg default_created_date "$DEFAULT_CREATED_DATE" \
+    '{created: $default_created_date, config:{}, history:[], rootfs:{type: "layers", diff_ids:[]}} + $platform' | \
+    update_config >/dev/null
```
hanneskaeufler, line 36: This is just formatting and using the variable for the default date

oci/private/image.sh
```diff
@@ -174,6 +180,13 @@ for ARG in "$@"; do
       jq --rawfile annotations "${ARG#--annotations=}" \
       '.annotations += ([($annotations | split("\n") | .[] | select(. != ""))] | map(. | split("=")) | map({key: .[0], value: .[1:] | join("=")}) | from_entries)' |
       update_manifest
+
+    created_date="$(jq -r -n \
+      --arg created_key "${OPENCONTAINERS_CREATED_KEY}" \
+      --arg default_created_date "${DEFAULT_CREATED_DATE}" \
+      --rawfile annotations "${ARG#--annotations=}" \
+      '([$annotations | split("\n") | .[] | select(. != "") | split("=") | {key: .[0], value: .[1]}]) | map(select(.key==$created_key)) | first | .value // $default_created_date')"
```
hanneskaeufler, line 188: I'm sure we could remove some duplication in this and the above annotations handling

thesayyn:
This implicit behavior is something i am really scared of, i always imagined we would add a new attribute to oci_image that takes a file (so its also stampable.)

hanneskaeufler:
> This implicit behavior is something i am really scared of

I totally get that! But what I feel is also dubious, is supplying both the `created` annotation and a `created` file.  How do we consolidate that, or do we simply not?

hanneskaeufler:
@thesayyn  I implemented the alternative: #724 we can discuss over there for any open questions I have regarding that approach.

PR #724, opened 2024-10-18
Introduce new attribute on `oci_image` for `created` datetime

### LINKED ISSUE #661: oci_image should support setting `created` and stamp

hanneskaeufler:
Fixes #661 . Alternative to #722.

### Why?

Doing a docker inspect --format='{{.Created}}' my/image:latest on an image built with oci_image previously always returned the static beginning of the unix timestamp (1. January 1970 00:00:00). To be more precise, that's what it returns for images that are built "from scratch". When an image is built on a base, then the created date is set to the created date of the base image.

Both are fine strategies for reproducible builds, but in stamped builds that are eventually shipped/deployed, you probably want to set that time to the actual build time.

### What?

To allow setting that particular `Created` in the manifest json, we introduce a separate file label, allowing to pass in a file containing a stamp variable.


hanneskaeufler:
@alexeagle @thesayyn ping in case you haven't seen it. If you'll allow the CI to run I'll make sure to pass that :) Thanks for your time!

thesayyn (commented):
Looks good to me, i wish this property was not just named `created` but something else like `created_at`

thesayyn:
Oh did not mean that you should rename it, we should use the `created` property as that's what the OCI spec says.

hanneskaeufler:
> Oh did not mean that you should rename it, we should use the `created` property as that's what the OCI spec says.

Sorry I don't follow. I feel `created_at` as an attribute here in rules_oci is indeed the better choice. In the json manifest of course I still keep `created` https://github.com/bazel-contrib/rules_oci/pull/724/files#diff-4d907c3439c162250ee007024daf813b7354ab62d3a510f68955ca9751d33618R173. 

Let me know if that clears it up :) Thanks!

thesayyn:
> Let me know if that clears it up :) Thanks!

the name of the attribute should be identical to config property. so both of them should be `created`. i was merely suggesting that OCI spec should have named it differently but there is nothing we can do now :)

e2e/smoke/workspace_status.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# This serves as an example of how the workspace_status.sh script can be used to
+# stamp the `created` attribute of an image.
+BUILD_TIMESTAMP=${BUILD_TIMESTAMP:-$(date +%s)} # macOS specific
```
thesayyn, line 5: e2e/smoke is run on multiple platforms, including linux and macos. Can we simplify this so it's not platform specific?

e2e/smoke/workspace_status.sh
```diff
@@ -0,0 +1,7 @@
...
+BUILD_ISO8601=$(date -u -r "$BUILD_TIMESTAMP" +"%Y-%m-%dT%H:%M:%SZ")
```
thesayyn, line 6: same as above. simple `echo static_date_string_here` would suffice

thesayyn (commented):
some concerns on e2e, otherwise lgtm

e2e/smoke/workspace_status.sh
hanneskaeufler: We sure can, note though that in fact this file isn't actually being used (since CI doesn't --stamp). That's why I just added the comment above. I have no strong feelings about that file, I might as well remove it.

hanneskaeufler:
> some concerns on e2e, otherwise lgtm

Pushed everything I have now, _should_ pass CI as well :) Thx again!

thesayyn:
@hanneskaeufler  could you rebase?

hanneskaeufler:
@thesayyn sure can, done ✅

thesayyn (approved):

PR #725, opened 2024-10-19
feat: output digest for image push

jrandolf:
In CI, the digest is useful for deployment after building.


thesayyn:
This is already the case for push. Reference of the image is already being printed to stdout.

thesayyn:
1- This is a breaking change 
2- Could you please describe the change 
3- How is this is this fixing it

jrandolf:
> This is already the case for push. Reference of the image is already being printed to stdout.

It's better to just output the image digest for easier consumption, otherwise it needs to be grepped.

In any case, we are using a different model for our deployment now.

PR #740, opened 2024-12-04
feat: override image labels

genert:
Currently labels are merged with existing ones. Proposing to fully override existing labels.


thesayyn:
Well, to begin with this is a breaking change, and second i am not sure if this is the right behaviour.

thesayyn:
Going to close this as inactive.

Issue #735, opened 2024-11-13
Using `oci_image_index` doesn't produce a `.digest` target while `oci_image` does


mislavmandaricaxilis:
After #346 got merged, using `oci_image` macro also produces a `.digest` target, which contains a digest of the image.

But this was not done for `oci_image_index`, so switching to a multi-arch images causes a regression in experience because `.digest` target is missing.

Same as with `oci_image`, it's easy to write it ourselves, but having it in `rules_oci` directly makes it easy and consistent.

I'm willing to contribute a PR myself for this change if maintainers agree with the suggestion.

thesayyn:
This is a reasonable request, PRs are welcome.

mislavmandaricaxilis:
Here's the PR for this issue.

mislavmandaricaxilis:
Fixed in #742.

PR #742, opened 2024-12-11
feat: Transition oci_image_index into a macro with additional [name].digest target

mislavmandaricaxilis:
Related issue #735.

Since `oci_image_index` was a rule directly exposed in `defs.bzl`, I had to introduce a new macro. I opted to make it in the same way as `oci_index`, by using the `oci_index_image` name for the macro, while renaming the existing rule as `oci_index_image_rule`. I'm not sure if this kind of change would be considered a breaking change or not.

Please review if you find this kind of change from a rule to a macro acceptable or not. If not, I can expose this new macro with a different name (if you have any suggestions on what name to use, I'm all ears 😅) keeping the old rule unchanged.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -18,7 +18,11 @@ stardoc_with_diff_test(
 
 stardoc_with_diff_test(
     name = "image_index",
-    bzl_library_target = "//oci/private:image_index",
+    bzl_library_target = "//oci:defs",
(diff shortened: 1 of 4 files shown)


thesayyn (approved):

alexeagle (approved):
nicely done :)
Not a breaking change since macros are isomorphic with rules (callsite cannot distinguish)

PR #745, opened 2024-12-11
chore(cd): automate release tagging

alexeagle:
Simplifies life for maintainers a bit, ensuring we're never more than two weeks without commits getting shipped.

Downstreams https://github.com/bazel-contrib/rules-template/pull/129

### DIFF
--- a/.github/workflows/tag.yaml
+++ b/.github/workflows/tag.yaml
@@ -1,0 +1,43 @@
+# Tag a new release using https://github.com/marketplace/actions/conventional-commits-versioner-action
+#
+# This is easier than having to run manual `git` operations on a local clone.
+# It also runs on a schedule so we don't leave commits unreleased indefinitely
+# (avoiding users having to ping "hey could someone cut a release").
+
+name: Tag a Release
+on:
+  # Allow devs to tag manually through the GitHub UI.
+  # For example after landing a fix that customers are waiting for.
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


.github/workflows/tag.yaml
```diff
@@ -0,0 +1,43 @@
...
+# It also runs on a schedule so we don't leave commits unreleased indefinitely
+# (avoiding users having to ping "hey could someone cut a release").
+
+name: Tag a Release
+on:
+  # Allow devs to tag manually through the GitHub UI.
+  # For example after landing a fix that customers are waiting for.
+  workflow_dispatch:
+
+  # Run twice a month, on the 12th and 27th at 3PM UTC (8AM PST)
```
alexeagle, line 13: note, I'm setting these dates arbitrarily. We don't want all rulesets shipping on the same day of the month I think. Ideally the dependency graph would inform the release schedule, but I don't have time to think that through.

In this case I just want it to trigger tomorrow.

thesayyn: This sounds fine, it means 24 extra releases per year which sounds fine IMO.

thesayyn (approved):

.github/workflows/tag.yaml
alexeagle: Yeah, also note that it won't create a tag if all commits since last release were just `chore` so it may be fewer releases than that.

PR #757, opened 2024-12-19
feat: add config data in image manifest

### LINKED ISSUE #756: oci_image inherits data field from config in base image
The latest debian image now adds the `data` field under `config` in the image manifest:
```
docker manifest inspect debian@sha256:ec54b6327d5099ab29b38d70f7290e42d8769ef676fc262b34a18b688104f61b
{
        "config": {
                "data": "eyJhcmNoaXRlY3R1cmUiOiJhbWQ2NCIsImNvbmZpZyI6eyJDbWQiOlsiYmFzaCJdLCJFbnRyeXBvaW50IjpbXSwiRW52IjpbIlBBVEg9L3Vzci9sb2NhbC9zYmluOi91c3IvbG9jYWwvYmluOi91c3Ivc2JpbjovdXNyL2Jpbjovc2JpbjovYmluIl19LCJjcmVhdGVkIjoiMjAyNC0xMi0wMlQwMDowMDowMFoiLCJoaXN0b3J5IjpbeyJjb21tZW50IjoiZGVidWVycmVvdHlwZSAwLjE1IiwiY3JlYXRlZCI6IjIwMjQtMTItMDJUMDA6MDA6MDBaIiwiY3JlYXRlZF9ieSI6IiMgZGViaWFuLnNoIC0tYXJjaCAnYW1kNjQnIG91dC8gJ2Jvb2t3b3JtJyAnQDE3MzMwOTc2MDAnIn1dLCJvcyI6ImxpbnV4Iiwicm9vdGZzIjp7ImRpZmZfaWRzIjpbInNoYTI1NjozMDFjMWJiNDJjYzBiYzY2MThmY2FmMDM2ZTg3MTFmMmFhZDY2Zjc2Njk3ZjU0MWUyMDE0YTY5ZTFmNDU2YWE0Il0sInR5cGUiOiJsYXllcnMifX0K",
                "digest": "sha256:ff869c3288a47c9625a60473a3d5108ec45bd095a00e23568a82ee8b95d12954",
                "mediaType": "application/vnd.oci.image.config.v1+json",
                "size": 453
        },
        "layers": [
                {
                        "digest": "sha256:fdf894e782a221820acf469d425b802be26aedb5e5d26ea80a650ff6a974d488",
                        "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
                        "size": 48497210
                }
        ],
        "mediaType": "application/vnd.oci.image.manifest.v1+json",
        "schemaVersion": 2
}
```

In the `oci_image` rule, the image built directly inherits the `data` field from the `config` in the base image, despite having a different config. This causes the data field to no longer match the `digest` for the config, causing verification errors. This can break many libraries and tools (such as https://github.com/google/go-containerregistry) because they will try and verify the manifest and fail to do so.

chenhunter:
This PR adds the `data` field to the config of the manifest in the image of `oci_image`. 

Fixes #756

### DIFF
--- a/oci/private/image.sh
+++ b/oci/private/image.sh
@@ -55,9 +55,11 @@ function get_config() {
 function update_config() {
   local digest=
   local config=
+  local data=
   config="$(coreutils cat -)"
   digest="$(echo -n "$config" | regctl blob put "$REF")"
-  get_manifest | jq '.config.digest = $digest | .config.size = $size' --arg digest "$digest" --argjson size "${#config}" | update_manifest >/dev/null
+  data="$(echo -n "$config" | coreutils base64)"
+  get_manifest | jq '.config.digest = $digest | .config.size = $size | .config.data = $data' --arg digest "$digest" --argjson size "${#config}" --arg data "$data" | update_manifest >/dev/null
   echo "$digest"
 }
 


thesayyn:
I think the right behavior is to just drop that field as its not in the spec.

chenhunter:
I originally thought it wasn't in the spec either, but had to look again as well. Turns out it's been in the spec for year, and listed as REQUIRED. 

Dropping the field definitely is a solution, but considering it's in the spec as required, and the work is already done in the pull request, I don't see a good reason to NOT just add it? Unless the implementation is wrong or doesn't follow best practices. If so, please advise.

thesayyn:
Still, though, i think we should only inline the config if we see that the base has it too.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_oci/pull/774

PR #759, opened 2024-12-24
Fix compatibilty with symbolic macros

dzbarsky:
Symbolic macros require labels to start with their own name, so if you have a symbolic macro wrapping one of the OCI legacy macros, it will fail. Adjust the names to satisfy this new requirement.

See https://bazel.build/extending/macros#naming, these naming conventions are violated by this legacy macro (and thus by a symbolic macro wrapping this legacy macro)


alexeagle (approved):

dzbarsky:
@alexeagle looks like tests need approval to trigger

PR #761, opened 2025-01-03
oci_pull: add setting for disabling token retrieval in rules_oci

### LINKED ISSUE #760: Allow users to disable builtin credential handling
I have a custom Bazel credential helper that can derive working `Authorization` headers for oci registries. In one case, this is for `ghcr.io`, for which `rules_oci` has a token realm preconfigured.
This can lead to situations where my credential helper would inject the right headers for pulling an image, but the extra requests (and command executions for docker credential helpers) fail before I get there.

I would like to have a setting (like a `repo_env` var) for disabling the builtin credential handling completely, since my docker config may be missing or misconfigured.

Happy to implement such a setting, but I‘d like to know if that’s something that can be merged first.

malt3:
This is helpful when using an external credential helper (via `--credential_helper`).

Closes #760

### DIFF
--- a/oci/private/authn.bzl
+++ b/oci/private/authn.bzl
@@ -275,6 +275,8 @@ To enable this feature, add `common --repo_env=OCI_ENABLE_OAUTH2_SUPPORT=1` to t
 """
 
 def _get_token(rctx, state, registry, repository):
+    if rctx.os.environ.get("OCI_DISABLE_GET_TOKEN"):
+        return {}
     pattern = _get_auth(rctx, state, registry)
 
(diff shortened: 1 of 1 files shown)


mering:
Could you also mention this in the docs somewhere?

malt3:
> Could you also mention this in the docs somewhere?

Yes. I’ll rewrite this to allow smooth fallback (instead of disabling the builtin behavior) as stated in #760. I’ll document the flag in the process.

thesayyn:
Should we close this PR?

PR #762, opened 2025-01-16
use transition to resolve correct tar in oci_load

contagnas:
This is similar to and based on https://github.com/bazel-contrib/rules_oci/pull/590

We're building `oci_load` targets with RBE on linux machines, and the tar in the load script fails to execute on macos hosts.


thesayyn:
I am not sure if this workaround is fixing problems, or introducing more, i am even considering reverting https://github.com/bazel-contrib/rules_oci/pull/590. 

See: https://github.com/bazel-contrib/rules_oci/pull/590#issuecomment-2440798610

PR #763, opened 2025-01-17
Do not announce platform repos as module dependencies

### LINKED ISSUE #727: Bzlmod doesn't detect repo usage correctly
When following the instructions from the [release](https://github.com/bazel-contrib/rules_oci/releases/tag/v2.0.1):

```
bazel_dep(name = "rules_oci", version = "2.0.1")
# For testing, we also recommend https://registry.bazel.build/modules/container_structure_test

# Declare external images you need to pull, for example: 
oci.pull(
    name = "distroless_base",
    # 'latest' is not reproducible, but it's convenient.
    # During the build we print a WARNING message that includes recommended 'digest' and 'platforms'
    # values which you can use here in place of 'tag' to pin for reproducibility.
    tag = "latest",
    image = "gcr.io/distroless/base",
    platforms = ["linux/amd64"],
)

# For each oci.pull call, repeat the "name" here to expose them as dependencies.
use_repo(oci, "distroless_base")
```

I get the following warning:
```
WARNING: REPO/MODULE.bazel:320:20: The module extension oci defined in @rules_oci//oci:extensions.bzl reported incorrect imports of repositories via use_repo():

    distroless_base

 ** You can use the following buildozer command to fix these issues:

buildozer 'use_repo_remove @rules_oci//oci:extensions.bzl oci distroless_base' //MODULE.bazel:all
```

mering:
Fixes #727

### DIFF
--- a/oci/extensions.bzl
+++ b/oci/extensions.bzl
@@ -46,8 +46,6 @@ def _oci_extension(module_ctx):
             if mod.is_root:
                 deps = root_direct_dev_deps if module_ctx.is_dev_dependency(pull) else root_direct_deps
                 deps.append(pull.name)
-                for platform in pull.platforms:
-                    deps.append("_".join([pull.name] + platform.split("/")))
 
         for toolchains in mod.tags.toolchains:
             if toolchains.name != "oci" and not mod.is_root:


thesayyn:
There is not much we could do about this; we have to report what can imported by the module, and that includes platform specific repos as well. 

If bazel had a way to let us tell it what *can* be imported, then we could use that instead.

mering:
> There is not much we could do about this; we have to report what can imported by the module, and that includes platform specific repos as well.
> 
> If bazel had a way to let us tell it what _can_ be imported, then we could use that instead.

Bazel already knows what _can_ be imported: all repos generated by the extension. This is also how it calculates the _indirect_ dependencies (all generated ones minus the direct ones).
The information about direct dependencies is used by `bazel mod tidy` to automatically update the `use_repo()` call. Is there any use case of using the platform specific repos over the general one (which would use the platform specific ones as indirect dependency)?

thesayyn:
> Bazel already knows what _can_ be imported: all repos generated by the extension

right, sorry.

> Is there any use case of using the platform specific repos

Yes, if you want to build an image for a specific architecture without using transitions, eg you want to build a python image, and don't want to use transitions to get correct base image.  

So instead of `base = "@base"`, you use `@base_linux_amd64`, that's why we report these platform specific repos for tidy.

mering:
> > Is there any use case of using the platform specific repos
> 
> Yes, if you want to build an image for a specific architecture without using transitions, eg you want to build a python image, and don't want to use transitions to get correct base image.
> 
> So instead of `base = "@base"`, you use `@base_linux_amd64`, that's why we report these platform specific repos for tidy.

This only works when using pure Python without any FFI pip dependencies, right?

Maybe it might be better not to report anything? I think the currently state is quite bad.

In one repo the platform specific stuff is added which we don't need and which confuses people. In the other repo we want to use the same (version of) a base image from another repo. There it is reported as indirect dependency and gets removed and breaks our build. So if there is no good way for the extension to determine which imports should be used, maybe leave it to be always manually specified by the user?

thesayyn:
> This only works when using pure Python without any FFI pip dependencies, right?

Yes, but python is just an example, nodejs, java all can work on any platform similar to python, so we need to report these. 

I don't understand *exactly* how this is creating a problem for you.  

Right here we report deps if the extension was called from from the root module, so you shouldn't see any warnings from "other" repos, also we always add these to direct dependencies. 
https://github.com/bazel-contrib/rules_oci/blob/a911e9eea205568939b67c8d90f6014476dc00c4/oci/extensions.bzl#L47

As for the confusion part, as i said there is *no* way for the extension to know what you import and don't, there we can't conditionally report platform specific ones, what we need is bazel supporting auto removal of `use_repo` calls reported by the extension if they are not used anywhere in the repo.   Its confusing for you use case to have these `use_repo` calls, it is also confusing to not have them, that's why we added them in the first place.  

See when this was introduced and discussion around it https://github.com/bazel-contrib/rules_oci/pull/611

thesayyn:
See: https://github.com/bazel-contrib/rules_oci/pull/611/files#r1646018127
https://github.com/bazel-contrib/rules_oci/pull/611/files#r1646582755

thesayyn:
I am going to close this as there is not much to be fixed here in rules_oci. See the links i have pasted above for more context.

PR #767, opened 2025-01-26
refactor: remove workspace

alexeagle:
Prefactoring for upgrade to Bazel 8.

We don't need to test the root module under WORKSPACE anymore.

To make it cleaner, I moved tool downloads to rules_multitool:
- Adds convenience symlinks for running tools directly.
- More compact and easy for users to copy to their tools.lock file.


.gitattributes
```diff
@@ -8,3 +8,7 @@ docs/load.md linguist-generated=true
 #################################
 # Configuration for 'git archive'
 # See https://git-scm.com/docs/git-archive#ATTRIBUTES
+
+# Don't include examples or tools in the distribution artifact, to reduce size
+examples export-ignore
```
thesayyn, line 13: i briefly remember that removing these created some problems for us in BCR

thesayyn (approved):

Issue #736, opened 2024-11-15
Add retry field to oci_push and possibly other oci commands


davidhao3300:
Similar to https://github.com/bazelbuild/rules_docker/issues/1989, oci_push() can sometimes flake due to transient network issues, and retrying would be helpful for mitigating the impact. The current alternative is to retry the whole bazel command

PR #776, opened 2025-02-14
feat: add retry_count to image_push

LavaTime:


LavaTime:
This is supposed to solve #736

oci/private/push.bzl
```diff
@@ -180,6 +180,10 @@ _attrs = {
         """,
         allow_single_file = True,
     ),
+    "retry_count": attr.int(
```
thesayyn, line 183: AFAIK, crane already has a retry built into the wire which handles thins like network jitter, and retryable http statuses.  

What's the reason for this addition?  

Also, you can easily do this in your shell calling into the `bazel run :push`.  Why not do that?

thesayyn:
Housekeeping, please comment if you'd like to finish this PR.

PR #778, opened 2025-02-17
authn: allow builtin credential handling to fail conditionally

### LINKED ISSUE #760: Allow users to disable builtin credential handling

malt3:
Introduces the OCI_GET_TOKEN_ALLOW_FAIL repo env.
If set, the builtin credential handling will be allowed to fail and a credential helper can be used instead.

Fixes #760

### DIFF
--- a/docs/pull.md
+++ b/docs/pull.md
@@ -127,6 +127,13 @@ For more information about the credential helpers checkout the [documentation](h
 
 See the [examples/credential_helper](/examples/credential_helper/auth.sh) directory for an example of how this work.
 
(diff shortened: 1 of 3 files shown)


malt3:
@thesayyn FYI

thesayyn (approved):

thesayyn:
Thank you

PR #780, opened 2025-02-24
Add support for registry_mirrors

sluongng:
Most of the PR was generated by Github Copilots.

### DIFF
--- a/docs/pull.md
+++ b/docs/pull.md
@@ -132,8 +132,8 @@ See the [examples/credential_helper](/examples/credential_helper/auth.sh) direct
 ## oci_pull
 
 <pre>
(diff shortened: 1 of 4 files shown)


sluongng:
I tested it manually with BuildBuddy repo with
```
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -536,6 +536,7 @@ oci_pull(
     name = "busybox",
     digest = "sha256:c230832bd3b0be59a6c47ed64294f9ce71e91b327957920b6929a0caa8353140",
     image = "mirror.gcr.io/library/busybox:1.36.1",
+    registry_mirrors = ["mirror.gcr.io"],
     platforms = ["linux/amd64"],
 )
```

and `common --override_repository=rules_oci=/Users/sluongng/work/misc/rules_oci`.

Then do

sluongng:
hmm i missed that I double set the mirror in oci_pull... need to fix it a bit more

apesternikov:
It could be very useful to have an ability to configure all parts of the mirrored image location, not just a registry. It is a common use case to mirror external images to a private repository with additional prefix using  https://github.com/fasterci/rules_gitops/tree/main/mirror. Is it possible to expose an additional images[] parameter?

thesayyn:
Thank you for the PR, i feel like this needs a little more thought.  It is possibly a dangerous feature since oci_pull allows mutable tags, so this could really break builds if used incorrectly.

alexeagle:
This adds complexity for rules_oci, and the mutable tag issue is tricky.

Also, we think that solving the "unreliable download" problem for oci.pull specifically is the wrong layering. Users should have a more reliable `rctx.download`, by using `--remote_downloader` or a read-through proxy.

PR #784, opened 2025-03-26
feat: support transition in image_index

thesayyn:
EXPERIMENTAL feature for transitioning oci_image_index to multiple platforms rendering use of platform_transition_filegroup unnecessary. 

Fixes https://github.com/bazel-contrib/rules_oci/issues/228

### DIFF
--- a/oci/private/image_index.bzl
+++ b/oci/private/image_index.bzl
@@ -2,7 +2,7 @@
 
 _DOC = """Build a multi-architecture OCI compatible container image.
 
-It takes number of `oci_image` targets to create a fat multi-architecture image conforming to [https://github.com/opencontainers/image-spec/blob/main/image-index.md](OCI Image Index Specification).
+It takes number of `oci_image` targets to create a fat multi-architecture image conforming to [OCI Image Index Specification](https://github.com/opencontainers/image-spec/blob/main/image-index.md).
 
 Image indexes can be created in two ways:
 
--- a/docs/image_index.md
+++ b/docs/image_index.md
@@ -16,7 +16,7 @@ oci_image_index_rule(<a href="#oci_image_index_rule-name">name</a>, <a href="#oc
 
 Build a multi-architecture OCI compatible container image.
 
-It takes number of `oci_image` targets to create a fat multi-architecture image conforming to [https://github.com/opencontainers/image-spec/blob/main/image-index.md](OCI Image Index Specification).
+It takes number of `oci_image` targets to create a fat multi-architecture image conforming to [OCI Image Index Specification](https://github.com/opencontainers/image-spec/blob/main/image-index.md).
 
 Image indexes can be created in two ways:
 


oci/private/image_index.bzl
```diff
@@ -40,6 +65,9 @@ def _expand_image_to_args(image, expander):
     return args
 
 def _oci_image_index_impl(ctx):
+    if len(ctx.attr.platforms) > 0 and len(ctx.attr.images) != len(ctx.attr.platforms):
+        fail("platforms can only be specified when there is exactly one image in the images attribute.")
```
sitaktif, line 69:
```suggestion
        fail("If set, 'platforms' needs to have the same number of elements as 'images'.")
```

thesayyn: No that’s no right, number of images is  len(platforms) * len(images), for only one it has to be exactly len(platforms)

sitaktif: Ok, that makes sense. What is surprising to me is that the conditional doesn't seem to cover the failure message that's being displayed. Did you mean the following conditional instead?

```
if len(ctx.attr.platforms) > 0 and len(ctx.attr.images) != 1:
```

alexeagle:
Would you like to pull in folks from that buildbarn multi-arch image change where we promised this?

oci/private/image_index.bzl
thesayyn:
```
oci_image_index(
  platforms = ["platform1", "platform2"]
  images = ["image1"]
)
```

Length of `ctx.attr.images` will be 2 even though there is one image, ctx.attr.images contains a matrix of `images * platforms`

thesayyn: So its correct.

thesayyn:
> Would you like to pull in folks from that buildbarn multi-arch image change where we promised this?

I don't think we need it at the moment, this is experimental and we can change it however we like if we decide that it needs to change.

thesayyn:
@alexeagle https://github.com/buildbarn/bb-storage/pull/242 i can confirm that this is working identical to platform_transition_filegroup.

alexeagle (approved):

oci/private/image_index.bzl
moroten, line 1: Maybe you would like to bring in some documentation examples for both `push` and `image_index` from #531 but annotate them as experimental.

examples/image_index/say_hello_amd64.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+echo "Hello, amd64!"
```
moroten, line 3: Add an empty line at end of file?

examples/image_index/say_hello_arm64.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+echo "Hello, arm64!"
```
moroten, line 3: Add an empty line at end of file?

oci/private/image_index.bzl
sitaktif: Thanks for taking the time to explain this very slowly to me, I get it now 🙏

oci/private/image_index.bzl
thesayyn: Yes, that's something i should do here.

thesayyn: Done. left push intanct, with a reference to oci_image_index for completeness.

PR #786, opened 2025-03-28
feat: allow custom www-authenticate challenges 

thesayyn:
We have been maintaining a list of WWW-Authenticate challenges for  well known registries https://github.com/bazel-contrib/rules_oci/blob/main/oci/private/authn.bzl#L9. While this worked for commonly known registries, it failed short when private registries got involved. 

This PR implements a `www_authenticate_challenges` attribute to `oci.pull` to allow extending of this list with additional entries provided by the call site. 

Example: 

```
oci.pull(
    name = "my_private_image",
    digest = "sha256:deadbeef",
    image = "corp.myprivate.reg.io/empty_image",
    www_authenticate_challenges = {
        "corp.myprivate.reg.io": 'Bearer realm="corp.myprivate.reg.io/token",service="localhost"',
    },
)
```

fixes https://github.com/bazel-contrib/rules_oci/issues/126


oci/repositories.bzl
```diff
@@ -107,12 +107,13 @@ def oci_register_toolchains(name, register = True):
         register: whether to call through to native.register_toolchains.
             Should be True for WORKSPACE users, but false when used under bzlmod extension
     """
-    bazel_features_deps()
-    register_jq_toolchains(register = register)
-    register_tar_toolchains(register = register)
-    register_coreutils_toolchains(register = register)
-    register_copy_to_directory_toolchains(register = register)
-    register_zstd_toolchains(register = register)
+    if register:
```
thesayyn, line 110: This is needed to allow calling of `oci.toolchains` in bzlmod.

alexeagle: Sure, but it seems surprising to couple that change with this added attribute? Is it just for convenience to put them in the same PR?

oci/pull.bzl
```diff
@@ -131,6 +131,28 @@ When using a credential helper, it may be desirable to allow the built-in creden
 ```
 common --repo_env=OCI_GET_TOKEN_ALLOW_FAIL=1
 ```
+
+### Providing WWW-Authenticate challenges
```
alexeagle, line 135: nit: the header should probably be about authenticating to registries, the tech detail of using www-authenticate header/challenge is probably not known to a casual reader

alexeagle (approved):

oci/repositories.bzl
thesayyn: this surfaced by https://github.com/bazel-contrib/rules_oci/pull/786/files#diff-913c901b3a516744550c8d29a001ca03b41ca34d06bbbf585842553367c96dd5R13 and was wrong from get go. i can probably split it into a separate PR, its just going to take more cycles for us to land this.

oci/pull.bzl
```diff
@@ -131,6 +131,28 @@ When using a credential helper, it may be desirable to allow the built-in creden
...
+challenges to be provided via the `www_authenticate_challenges` attribute. This is useful for private registries that require additional authentication.
+
+```starlark
+oci.pull(
+    name = "my_private_image",
+    image = "my.private.registry/my/image",
+    digest = "sha256:deadbeef",
+    www_authenticate_challenges = {
+        # https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/WWW-Authenticate
+        "index.docker.io": 'Bearer realm="https://auth.docker.io/token",service="registry.docker.io"'}
```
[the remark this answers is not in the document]
thesayyn: oh right, i should change that. that's confusing.

oci/pull.bzl
thesayyn: i agree, in ideal case, we'd be able to read rctx.download  headers but since that's not possible, we'd have to expose this to advanced users.

oci/pull.bzl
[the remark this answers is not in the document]
thesayyn: done

PR #788, opened 2025-04-04
refactor: publish better buildx api

thesayyn:
For a long time we had dockerfile example but its DX was never good, this updates our  examples/dockerfile to be standalone and publishes the api with better DX that we created for somebody else.

### DIFF
--- a/examples/dockerfile/buildx.bzl
+++ b/examples/dockerfile/buildx.bzl
@@ -1,6 +1,142 @@
 "repos for buildx"
 
 load("@aspect_bazel_lib//lib:repo_utils.bzl", "repo_utils")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
+
+BUILDX_WORKAROUND = """
+def _oci_layout_buildx_workaround_impl(ctx):
+    '''workaround for buildx to accept oci-layout with lock file and ingest directory already created
+    See: https://github.com/docker/buildx/issues/2753https://github.com/docker/buildx/issues/2753#issuecomment-2436601290https://github.com/containerd/containerd/issues/10885https://github.com/docker/buildx/issues/2753#issuecomment-2436324404
+    '''
+    output = ctx.actions.declare_directory(ctx.label.name)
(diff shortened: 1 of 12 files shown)


examples/dockerfile/buildx.bzl
```diff
@@ -14,48 +148,93 @@ def _impl_configure_buildx(rctx):
     compatible_with = "[]"
     builder_name = "builder-docker"
     if has_docker:
-        # TODO(alex): a less hacky way for a repo rule to find this transitive repo label?
-        if repo_utils.platform(rctx) == "darwin_amd64":
-            multitool_label = "@@rules_multitool~~multitool~multitool.macos_x86_64//tools/buildx:macos_x86_64_executable"
```
thesayyn, line 19: @alexeagle sorry for removing your multitool setup from this example but i need the buildx.bzl file to be standalone and only source that needs to be copied over.

examples/dockerfile/buildx.bzl
```diff
@@ -14,48 +150,100 @@ def _impl_configure_buildx(rctx):
...
-                buildx,
-                "create",
-                "--name",
-                builder_name,
-                "--driver",
-                "docker-container",
-                "--use",
-                "--bootstrap",
-            ])
+            r = rctx.execute([buildx, "create", "--name", builder_name, "--driver", "docker-container"])
```
vinnybod, line 157: We added `--use --bootstrap` to fix a race condition when this is used in parallel https://github.com/bazel-contrib/rules_oci/pull/775

```suggestion
            r = rctx.execute([buildx, "create", "--name", builder_name, "--driver", "docker-container", "--use", "--bootstrap"])
```

thesayyn: right, sorry, we should not have dropped it.

examples/dockerfile/buildx.bzl
alexeagle: haha, that's okay if it's getting in the way

alexeagle (approved):

Issue #700, opened 2024-09-18
`base` does no longer accept a tarball


mering:
In v1, `base` was accepting a tarball. I assume this is the reason why it sets `allow_single_file = True`. In v2, this fails with the following error:
```
ls: unknown io error: '"bazel-out/k8-fastbuild/bin/path/image_tarball/tarball.tar/blobs/*/*"', 'Os { code: 20, kind: NotADirectory, message: "Not a directory" }'
```

The corresponding code is https://github.com/bazel-contrib/rules_oci/blob/c74c2d35fb65450aed690815ee6e92e1b753ed1e/oci/private/image.sh#L33-L48

Could you detect if `base` is a tarball and extract it in this case?

We use this feature to gradually migrate our large code base from `rules_docker` to `rules_oci`.

thesayyn:
oci_image never accepted a tarball as the base, if it did though, probably because of crane we were running at the time. We won't be supporting tarballs as they are usually huge and not cache friendly.

mering:
@thesayyn I agree that one shouldn't _create_ tarballs as outputs but it might be necessary to support tarballs as _inputs_ to support migrations _to_ rules_oci.

So this is about unpacking a tarball _which already exists_ instead of _creating a new one_. So this doesn't make things worse as they already are today but instead paves the way to a better future with fewer tarballs.

thesayyn:
I agree with the sentiment here, however adding this feature back is not something i'd be warm to, i'd be much comfortable with an example showing use of regctl with a genrule to unpack a tarball into an oci-layout which is then used as the base.

mering:
> I agree with the sentiment here, however adding this feature back is not something i'd be warm to, i'd be much comfortable with an example showing use of regctl with a genrule to unpack a tarball into an oci-layout which is then used as the base.

Or maybe providing an explicit `oci_import()` rule to unpack the tar. Preferably also backporting to 1.x to allow for a smoother upgrade path (using that explicit rule already in 1.x and then upgrade).

mering:
@thesayyn which `regctl` command would you use to unpack a tarball into an oci-layout?

thesayyn:
`regctl import path/to/tarball oci-layout://path/to/output`

thesayyn:
Something like genrule would work but genrules no longer support outputting a directory, so you'd have to create a custom rule.

mering:
> `regctl image import path/to/tarball oci-layout://path/to/output`

[...]

PR #789, opened 2025-04-04
example: use docker tarball as base image

thesayyn:
Adds an example for using a docker tarball as a base image to oci_image. In 1.x version of rules_oci this worked by accident where passing a tarball to the base attribute of oci_image would extract the tarball and use it as the base image but in 2.x this stopped working because we actually started asserting that the base image was in fact an oci-layout. 

Users are required to register the regctl toolchain in their MODULE.bazel if bzlmod is in use. 

```
oci.toolchains()
use_repo(oci, "oci_regctl_toolchains")
```

Closes https://github.com/bazel-contrib/rules_oci/issues/700

### DIFF
--- a/examples/tarball_as_base/BUILD.bazel
+++ b/examples/tarball_as_base/BUILD.bazel
@@ -1,0 +1,33 @@
+load("@aspect_bazel_lib//lib:run_binary.bzl", "run_binary")
+load("@rules_oci//oci:defs.bzl", "oci_image", "oci_load")
+
+filegroup(
+    name = "regctl",
+    srcs = ["@oci_regctl_toolchains//:current_toolchain"],
+)
+
+run_binary(
+    name = "base",
+    srcs = ["tarball.tar"],
+    args = [
+        "image",
+        "import",
+        "ocidir://$@",
+        "$(location :tarball.tar)",
+    ],
+    execution_requirements = {"local": "1"},
+    mnemonic = "ExtractTarball",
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
could use some comments so a novice can read and follow along with how to adapt into their repo

jtszalay:
I gave adding this in my repo a shot and when adding the:
```
oci.toolchains()
use_repo(oci, "oci_regctl_toolchains")

register_toolchains("@oci_regctl_toolchains//:all")
```

I get the following error:
```
Error in repository_rule: A repo named bazel_features_version is already generated by this module extension at .../bazel-user-root/output/external/bazel_tools/tools/build_defs/repo/utils.bzl:268:18
```

Removing the `oci.toolchains()` call makes the build work but incorrectly reports:
```
WARNING: MODULE.bazel:367:20: The module extension oci defined in @rules_oci//oci:extensions.bzl reported incorrect imports of repositories via use_repo():

    oci_regctl_toolchains

Fix the use_repo calls by running 'bazel mod tidy'.
```

I would greatly benefit from being able to use this however!

thesayyn:
> I get the following error: 

Hmm, we put this behind a register flag that does not register it at all in the bzlmod context: https://github.com/bazel-contrib/rules_oci/blob/99ada11b01c99ee2546a6257fbca6c34a4546674/oci/repositories.bzl#L110-L116

PR #790, opened 2025-04-16
image_index.sh.tpl: use "dirname" from COREUTILS

dmivankov:
`dirname` is part of `coreutils` and may be unavailable on remote executors environments, let's use hermetic version

### DIFF
--- a/oci/private/image_index.sh.tpl
+++ b/oci/private/image_index.sh.tpl
@@ -35,7 +35,7 @@ function copy_blob() {
     local output_path="$2"
     local blob_image_relative_path="$3"
     local dest_path="${output_path}/${blob_image_relative_path}"
-    mkdirp "$(dirname "${dest_path}")"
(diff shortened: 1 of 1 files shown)


oci/private/image_index.sh.tpl
```diff
@@ -35,7 +35,7 @@ function copy_blob() {
     local output_path="$2"
     local blob_image_relative_path="$3"
     local dest_path="${output_path}/${blob_image_relative_path}"
-    mkdirp "$(dirname "${dest_path}")"
+    mkdirp "$("${COREUTILS}" dirname "${dest_path}")"
```
thesayyn, line 38: Lets do the same for mkdirp?

alexeagle: I'll send a followup

alexeagle (approved):

PR #792, opened 2025-04-21
chore: use new Publish-to-BCR

alexeagle:
Now it's a reusable workflow and includes provenence attestations for release artifacts


kormide (approved):
:+1:  Reminder to remove the app installation from repos as you switch to the workflow, so we don't get double publishes.

alexeagle:
Yes I switched the App registration to be only on specific 15 repos in bazel-contrib before embarking on this project, thanks!

PR #794, opened 2025-04-25
Remove transition of crane & jq to target platforms

lionelfleury:
- these binaries are run on exec platform


addisonjones3:
+1 would love to see this merged. It's becoming a challenge to do cross-platform Linux builds on macOS with the current transition.

brianmcarey:
+1 - this would be great to see added. If I can help in anyway to get this in let me know.

lionelfleury:
fixed in #869

PR #804, opened 2025-05-15
fix: Use rlocation for runfiles in push script

contagnas:
Currently oci_push generates a script with relative paths for crane, jq, etc:

```
readonly CRANE="../oci_crane_darwin_arm64/crane"
readonly JQ="../jq_darwin_arm64/jq"
readonly IMAGE_DIR="path/to/image"
```

Because of this the push can only be run when the working directory is the push's runfiles directory or via `bazel run`.

This change resolves these with `rlocation` so the script can be run from anywhere. oci_load is already using this: https://github.com/bazel-contrib/rules_oci/blob/677ab75097a4e15c95dfb94ecbd34a741fe6c327/oci/private/load.sh.tpl#L4-L10


thesayyn (approved):
This is great. i can't believe i didn't do this before. however, its giving me PTSD, where i had to go through a lot of pain to make this work well. Lets see what the CI says.

thesayyn:
Do you mind resolving the conflicts here?

PR #805, opened 2025-05-16
tarball.sh.tpl: add nlink=1 to mspec

dmivankov:
Mitigates https://github.com/buildbarn/bb-remote-execution/issues/162 while https://github.com/libarchive/libarchive/pull/2587 isn't released yet.

OCI layout should be deduplicated already as blobs are stored content-addressable, so setting nlink=1 to disallow detection and creation of hardlinks within tarball should be safe

### DIFF
--- a/oci/private/tarball.sh.tpl
+++ b/oci/private/tarball.sh.tpl
@@ -18,7 +18,7 @@ readonly OUTPUT="{{output}}"
 function add_to_tar() {
     content=$1
     tar_path=$2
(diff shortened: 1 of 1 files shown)


thesayyn:
@dmivankov now that tar.bzl has a new release, do you think this is still important?

dmivankov:
If it is already the default in latest rules_oci release it'd be less important to apply nlink=1 yes

PR #810, opened 2025-05-26
Add missing dependency to //cosign:defs

smertnik3sh:
The cosign/defs.bzl also [depends on attest.bzl](https://github.com/bazel-contrib/rules_oci/blob/main/cosign/defs.bzl#L3). I suppose this dependency should be declared in bzl library.


thesayyn:
FWIW, We don’t support or maintain cosign rules.  they were always experimental and were never promoted to the public api.

PR #838, opened 2025-08-26
fix: Handle non-legacy external runfiles layout in oci_load

### LINKED ISSUE #752: Issue with using oci_load and external dependencies
I have 2 bazel modules. One at the root and the other in a sub directory. Example:

```
root
  /libA
    /BUILD.bazel <-- uses oci_load()
    /MODULE.bazel
  MODULE.bazel <-- Imports libA's module
```

In the root MODULE.bazel I have:
```
bazel_dep(name = "libA", version = "1.0")
local_path_override(module_name = "libA", path = "./libA")
```

And libA's BUILD file has:
```
... build app, package to tar, transition binary

oci_load(
    name = "load",
    image = ":transitioned_image",
    repo_tags = "tags.txt",
)
```

When I run `bazel run @libA//:load`, the command fails with the error:
```
tar: Error reading archive -: Can't open external/libA+/image/blobs/sha256/d057aa77ea6c6e060f84c1bbe1a261478509b8d0f11bdbbd3ca6a5a3bbd1e2d0
tar: Error exit delayed from previous errors.
open /var/lib/docker/tmp/docker-import-1307345443/repositories: no such file or directory
```

If I run `bazel run //:load` inside libA dir, it runs fine. I can also run the app from the root directory without any issues. It's only the image creation that fails.
I suspect the '+' symbol in the name is causing some problems for the `tar` command.

So the oci_load() does not work when used with external dependencies. Any workarounds/fixes?

sfc-gh-dbieliaiev:
## Problem
This PR fixes #752 oci_load failures with `--nolegacy_external_runfiles`. When using `oci_load` from external repositories with the `--nolegacy_external_runfiles` Bazel flag, the command would fail with tar errors indicating it cannot access image blob files. This issue only manifested when using the non-legacy external runfiles layout, which is becoming the default in newer Bazel versions.

## Root Cause: Runfiles Layout Assumption

The issue was caused by hardcoded assumptions about Bazel's runfiles directory structure in `oci/private/load.bzl` and `oci/private/load.sh.tpl`.

### Legacy vs Non-Legacy Runfiles Layout

**Legacy External Runfiles**: External repo files are located under `RUNFILES_DIR/main_workspace/external/external_repo/` - all files are under the main workspace directory.

**Non-Legacy External Runfiles**: External repo files are located directly under `RUNFILES_DIR/external_repo/` - external repositories have their own top-level directories.

### The Bug

The old implementation assumed legacy layout by always trying to cd into the main workspace directory with `"$TAR" --cd "$RUNFILES_DIR/{{workspace_name}}"`. With `--nolegacy_external_runfiles`, external repository files aren't under the workspace subdirectory, so tar couldn't find them.

### 1. Workspace-Aware Path Resolution

Added helper functions `_get_workspace_name`, `_get_runfiles_prefix`, and `_get_workspace_root_path` to dynamically determine correct paths based on whether files belong to the main workspace or external repositories.

### 2. Universal Runfiles Access

Changed the tar command to work from `RUNFILES_DIR` root instead of assuming a workspace subdirectory, making it compatible with both runfiles layouts.

### 3. Proper Path Substitution

Updated mtree path replacement to use workspace-specific prefixes, ensuring paths are correctly translated between build-time absolute paths and runtime runfiles-relative paths.

## Testing

Added comprehensive e2e tests covering all combinations:
- bzlmod + legacy runfiles
- bzlmod + non-legacy runfiles  
- WORKSPACE + legacy runfiles
- WORKSPACE + non-legacy runfiles

Tests verify both local and external `oci_load` work correctly in all modes.

## Before/After

**Before**: `bazel run @external_wksp//:load --nolegacy_external_runfiles` would fail with tar errors.

**After**: Same command succeeds and loads the image correctly.

## Impact & Compatibility

- **Forward compatible**: Works with `--nolegacy_external_runfiles` (future default)
- **Backward compatible**: Still works with `--legacy_external_runfiles`  
- **No breaking changes**: Existing usage unaffected
- **Universal support**: Works with both bzlmod and WORKSPACE

This fix ensures `oci_load` works correctly as Bazel evolves toward the non-legacy runfiles layout as the default.


oci/private/load.bzl
```diff
@@ -133,6 +133,24 @@ attrs = {
...
 
+def _get_workspace_name(ctx, file):
+    label = getattr(file, "owner", None)
+    if label and getattr(label, "workspace_name", None):
+        return label.workspace_name
+    return ctx.workspace_name
+
+def _get_runfiles_prefix(ctx, file):
+    ws = _get_workspace_name(ctx, file)
+    if ws and ws != ctx.workspace_name:
```
sfc-gh-ptabor, line 144: I assume the `ws != ctx.workspace_name` conditions is a no-op here 
(we could skip it here).

oci/private/load.bzl
```diff
@@ -219,9 +237,10 @@ def _load_impl(ctx):
             "{{tar}}": to_rlocation_path(ctx, bsdtar.tarinfo.binary),
             "{{mtree_path}}": to_rlocation_path(ctx, mtree_spec),
             "{{loader}}": to_rlocation_path(ctx, ctx.executable.loader) if ctx.executable.loader else "",
-            "{{manifest_root}}": manifest_json.root.path,
-            "{{image_root}}": image.root.path,
-            "{{workspace_name}}": ctx.workspace_name,
+            "{{manifest_root}}": _get_workspace_root_path(ctx, manifest_json),
```
sfc-gh-ptabor, line 240: Would be good to add here a comment describing existence of cases where dependencies (manifest_json, image) are defined in external workspace.

e2e/assertion/wksp/external_lib/BUILD.bazel
sfc-gh-ptabor, line 1: Consider calling it: `external_wksp` (or `external_workspace`), as it's more then just `external_lib`. 

Please put 'README.md' file explaining that this package is used for testing of interactions with cross-workspace dependencies.

sfc-gh-ptabor (approved):

oci/private/load.bzl
sfc-gh-dbieliaiev: Thanks, fixed.

oci/private/load.bzl
sfc-gh-dbieliaiev: Added

PR #842, opened 2025-09-17
Upgrade aspect_bazel_lib 2.7.2 -> 2.21.1

mostynb:
This newer version of aspect_bazel_lib no longer uses local_config_platform, which will not be available in bazel 9.x.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,7 +5,7 @@ module(
     compatibility_level = 1,
 )
 
-bazel_dep(name = "aspect_bazel_lib", version = "2.7.2")
+bazel_dep(name = "aspect_bazel_lib", version = "2.21.1")
 bazel_dep(name = "bazel_features", version = "1.10.0")
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "platforms", version = "0.0.8")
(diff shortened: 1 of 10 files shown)


mostynb:
@thesayyn: upgrading container_structure_test to the latest (1.19.1) seems to fix the e2e/smoke errors locally.

thesayyn:
Some tests are still failing. Do you mind taking a look?

mostynb:
@thesayyn: it looks like rules_jvm_external dropped support for bzlmod with bazel 6.x:  https://github.com/bazel-contrib/rules_jvm_external/issues/1055

Given that bazel 6 is going to be unsupported in december 2025, I wonder if this is worth fixing? How long does rules_oci plan to support bazel 6?
https://bazel.build/release

mostynb:
> We can probably drop support for Bazel 6. Feel free to make that change so that our CI is testing Bazel 7 and Bazel 8.

Here's a PR which tries to do that:
https://github.com/bazel-contrib/rules_oci/pull/854

mostynb:
Oh wait, that was just for the bazel version upgrade. Sorry for the noise.

eldjh3:
Any plans to bump to 3.x (aka bazel-lib) rather than keeping with 2.0 (aspect_bazel_lib)??

Issue #769, opened 2025-01-30
`oci_load` is incompatible with cross architecture remote execution.


sfc-gh-mhazy:
Hi, 
Since `rules_oci` started using tar toolchain, the `oci_load` has a subtle issue with remote execution where `host_platform != exec_platform`

Consider the following example:

```
bazel run //:my_image_tar.amd64 --config=remote --extra_execution_platforms=linux_x86_64 --platforms=linux_amd64
```

The produced load script will select `exec_compatible_with` tar toolchain - `linux_x86_64` and then will be executed on host, for instance `darwin_arm64`. 

I've tried to workaround this by 

```
def host_exec_compatible_rule(rule, name, **kwargs):
    rule(
        name = name,
        exec_compatible_with = [
            host_cpu_constraint(),
            host_os_constraint(),
        ],
        **kwargs
    )
```

However, such approach makes incorrect toolchain resolution for [building tarball itself](https://github.com/bazel-contrib/rules_oci/blob/7de244c3cdc78fb002b1b01b2806a1dc6bd5a1aa/oci/private/load.bzl#L193). The toolchains are chosen for host platform, while the rule is executed on linux_x86_64. 

Similar, but not 100% same issue: https://github.com/bazelbuild/bazel/issues/22848 . The workaround with `local_only_rule` works in this case as well, but it defeats the purpose of remote execution a bit and here there is no reason to use `no-remote-exec` tag.

That said, I think the problem may be very hard to solve:
* executables (load script here) can be both:
   * executed directly with `bazel run` - in this case toolchains should be selected for host
   * in principle be used as dependency to a build rule - in this case, the build rule would be executed remotely, so toolchains should be selected for execution platform.

Looking forward to your thoughts.

fmeum:
`oci_load` generates a script that should run on the target platform and needs a [`tar` binary that it can execute at runtime](https://github.com/bazel-contrib/rules_oci/blob/ba8f19bc82bcf2ac6bc0167081d8aa7b6c459a93/oci/private/load.bzl#L234). That is exactly what a toolchain with `target_compatible_with` constraints and no `exec_compatible_with` constraints provides. This needs a separate toolchain type because it has legitimately different resolution behavior than a toolchain used to run `tar` during rule execution.

There are no correct fancy alternatives. If we started again from scratch, a simpler alternative would have been to only offer toolchains with `target_compatible_with` constraints as those can be reused by build rules via the resolved alias pattern, but this doesn't work the other way round.

PR #844, opened 2025-09-25
fix: Use target tar toolchain for cross-platform execution

sfc-gh-dbieliaiev:
Hi @thesayyn, @alexeagle.

I'd like to help solve the cross-platform remote execution issue described in [bazel-contrib/rules_oci#769](https://github.com/bazel-contrib/rules_oci/issues/769), which significantly affects us. I see this is connected to the fundamental Bazel limitation described in [bazelbuild/bazel#19645](https://github.com/bazelbuild/bazel/issues/19645).

I've noticed that a similar issue was addressed for crane and yq toolchains in [bazel-contrib/rules_oci#590](https://github.com/bazel-contrib/rules_oci/pull/590), but the same transition-based solution for tar toolchain ([bazel-contrib/rules_oci#762](https://github.com/bazel-contrib/rules_oci/pull/762)) got stuck because it doesn't work well in other cases.

## Proposal

I'm suggesting to use target_toolchain which was [recently implemented](https://github.com/bazel-contrib/tar.bzl/pull/65) instead of toolchain to determine it correctly. As a prerequisite we need to migrate to tar.bzl.

## Changes summary
1. tar.bzl was used instead of `aspect_bazel_lib/tar.bzl`
2. target toolchain was used instead of toolchainl.

### DIFF
--- a/docs/cpp.md
+++ b/docs/cpp.md
@@ -76,7 +76,7 @@ use_repo(oci, "docker_lib_ubuntu")
 ```
 
 ```python
-bazel_dep(name = "tar.bzl", version = "0.3.0")
+bazel_dep(name = "tar.bzl", version = "0.7.0")
 ```
 
 To load the image, execute:
--- a/e2e/wasm/MODULE.bazel
+++ b/e2e/wasm/MODULE.bazel
@@ -2,6 +2,7 @@ bazel_dep(name = "rules_oci", version = "0.0.0")
 bazel_dep(name = "aspect_bazel_lib", version = "2.7.2")
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
 bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "tar.bzl", version = "0.7.0")
 bazel_dep(name = "rules_rust", version = "0.45.1")
 bazel_dep(name = "hermetic_cc_toolchain", version = "3.1.0")
 


sfc-gh-dbieliaiev:
@thesayyn Do you have objections against using target toolchain and more in favor of going with toolchain exec groups attribute? I'm not exactly sure how's it going to work(don't have an experience with execution groups) so it might be more elegant solution but if I understand correctly nobody is sure that it's going to work. 

On the other side target toolchain is a proven solution(although is not fancy), that might be a good idea to implement as a short term mitigation of this issue.

Also I encourage @alexeagle and @fmeum to join discussion here as there were comments in related PR with opinions that this solution makes sense.

fmeum:
`oci_load` generates a script that should run on the target platform and needs a [`tar` binary that it can execute at runtime](https://github.com/bazel-contrib/rules_oci/blob/ba8f19bc82bcf2ac6bc0167081d8aa7b6c459a93/oci/private/load.bzl#L234). That is exactly what a toolchain with `target_compatible_with` constraints and no `exec_compatible_with` constraints provides. This needs a separate toolchain type because it has legitimately different resolution behavior than a toolchain used to run `tar` during rule execution.

There are no correct fancy alternatives. If we started again from scratch, a simpler alternative would have been to only offer toolchains with `target_compatible_with` constraints as those can be reused by build rules via the resolved alias pattern, but this doesn't work the other way round.

sfc-gh-dbieliaiev:
@thesayyn @alexeagle @fmeum I noticed that https://github.com/bazel-contrib/rules_oci/pull/854 was merged recently so I rebased my PR and all tests are green now. Could you please consider to accept this PR so we can resolve the issue?

thesayyn (approved):
looks good to me

PR #850, opened 2025-10-16
feat: Allow configuration of tarball name

peakschris:
When an oci_image needs to be manifested into a tarball with specific name it's helpful to be able to specify that upfront

### DIFF
--- a/docs/load.md
+++ b/docs/load.md
@@ -25,7 +25,7 @@ docker run --rm my-repository:latest
 ## oci_load
 
 <pre>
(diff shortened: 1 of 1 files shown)


oci/private/load.bzl
```diff
@@ -118,6 +118,10 @@ attrs = {
         executable = True,
         cfg = "target",
     ),
+    "tarball_name": attr.string(
```
thesayyn: we usually call this output using `attr.output` can you follow that pattern?

Issue #839, opened 2025-08-27
key "mediaType" not found in dictionary on oci_image


WalkerGriggs:
We're having trouble bisecting this issue. Building the envoy OCI image throws the error `key "mediaType" not found in dictionary`. Sure enough, when we log the manifest, it _is_ missing.

``` json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:56a045af0d202cf95f6e8fd5f7e8b12721177406328ddd0b22f92fd6e17b5f86",
      "size": 1996,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:f6aaf78b954cfb6da0e34603157263de76edfc28d4fdbb37c82df2daef23a76b",
      "size": 1996,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    }
  ]
}
```

When I pull the manifest with docker, it _does_ contain the media type.

``` json
$ docker manifest inspect envoyproxy/envoy:v1.35.1
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 1996,
         "digest": "sha256:56a045af0d202cf95f6e8fd5f7e8b12721177406328ddd0b22f92fd6e17b5f86",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 1996,
         "digest": "sha256:f6aaf78b954cfb6da0e34603157263de76edfc28d4fdbb37c82df2daef23a76b",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

Reproducible example

```
+oci.pull(
+    name = "envoy",
+    digest = "sha256:522f4f88bdd741fe87b813bdca678a11c13e42e630cc4304a9e18193813c935d",
+    image = "docker.io/envoyproxy/envoy",
+    platforms = [
+        "linux/amd64",
+    ],
+    tag = "v1.35.1",
+)

+oci_image(
+    name = "envoy-image",
+    base = "@envoy",
+    visibility = ["//visibility:public"],
+)
[...]

PR #868, opened 2025-11-26
Allow manifests without explicit `mediaType`

ouillie:
Closes https://github.com/bazel-contrib/rules_oci/issues/839

The added test fails prior to the fix commit. It passes after.


thesayyn (approved):
perfect!

PR #881, opened 2026-01-12
fix: Propagate non-zero `tar` exit in `load.sh`

Silic0nS0ldier:
In `oci/private/load.sh.tpl` `tar` is invoked with [process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html), which is suppressing it's exit code.

For `tar` errors to be handled, `wait $!` is needed.

As an aside, this class of issues can be linted against with [`shellcheck` `SC2312`](https://www.shellcheck.net/wiki/SC2312).


thesayyn (approved):

Silic0nS0ldier:
@thesayyn Are you waiting on me for anything to get this merged?

PR #883, opened 2026-01-22
feat: add package metadata with PURL to oci_pull

stevebarrau:
Add package_metadata target to oci_pull generated repositories following the PURL spec for Docker images. The PURL includes registry, repository, and identifier information.

This is then picked up by [supply_chain_tools](https://registry.bazel.build/modules/supply_chain_tools) to produce SBOM for oci_images build by Bazel.


thesayyn (approved):

gzm0:
Got here while digging through some of the supply chain stuff. Any reason we're using the docker type and not the oci type (https://github.com/package-url/purl-spec/blob/main/types/oci-definition.json)?

PR #884, opened 2026-01-22
fix: Check builder exists via JSONL

njlr:
Makes builder check more robust using JSONL output


thesayyn (commented):
This probably belongs in https://github.com/aspect-build/rules_buildx

thesayyn:
Yes, i'll have a look.

PR #893, opened 2026-04-27
chore: upgrade cosign toolchain to v3.0.6

suhalvemu:
## Summary

Upgrades cosign toolchain to v3.0.6 and fixes test incompatibilities introduced by cosign v3's new default behaviors.

### Why these changes are needed

**Previous attempt ([#865](https://github.com/bazel-contrib/rules_oci/pull/865)) failed CI** because cosign v3 ships `sigstore-go` for bundle verification, which has two breaking changes for our test scripts:

#### 1. sigstore-go `protojson.Unmarshal` incompatibility

cosign v3 uses `sigstore-go` to verify attestation bundles. Internally, `sigstore-go` calls `protojson.Unmarshal` to parse the in-toto statement inside the bundle — but `protojson` only handles protobuf JSON format, not the legacy in-toto format (identified by `"_type"` key).

The old `--type spdx` with an SPDX tag-value predicate causes cosign to set `LegacyPredicate` on the statement. When serialized, this emits the legacy format. At verify time, sigstore-go fails:
```
invalid attestation: decoding json
```

**Fix:** switch from `--type spdx` + SPDX tag-value file to `--type spdxjson` + JSON SBOM file. The `spdxjson` type parses the predicate as a JSON object and uses `protojson.Marshal`, producing a protobuf JSON statement that sigstore-go parses correctly.

#### 2. `--tlog-upload=false` incompatible with `--use-signing-config=true`

cosign v3 enables `--use-signing-config=true` by default (Sigstore Public Good TUF config), which is incompatible with `--tlog-upload=false`. cosign explicitly rejects the combination:
```
--tlog-upload=false is not supported with --signing-config or --use-signing-config
```

Our test scripts use `--tlog-upload=false` to skip Rekor uploads during local registry testing.

**Fix:** pass `--use-signing-config=false` alongside `--tlog-upload=false` in all sign/attest test scripts.

### Changes

- **`cosign/private/versions.bzl`**: Add v3.0.6 as the first (default) entry; keep v2.6.1 and older versions
- **`cosign/private/attest.bzl`**: Expand `type` attribute allowlist to include all cosign v3 predicate types (`spdxjson`, `slsaprovenance02`, `slsaprovenance1`, `openvex`)
- **`examples/attest/`**, **`examples/assertion/attest_external/`**: Switch predicate from SPDX tag-value (`sbom.spdx`) to JSON SBOM (`sbom.spdx.json`), type from `spdx` to `spdxjson`; add new JSON SBOM files
- **All sign/attest test scripts**: Add `--use-signing-config=false` to signers/attesters; add `--insecure-ignore-tlog` to verifiers; use `jq -S` to normalize JSON before diff comparison (key ordering may differ after cosign round-trip)

### Backward compatibility

All flag changes and the `spdxjson` type exist in cosign v2.6.1 — verified by downloading and testing the v2.6.1 binary locally. Older toolchain versions remain available in `COSIGN_VERSIONS`.

## Test plan

- [x] `bazel test //examples/attest:test` — PASSED
- [x] `bazel test //examples/attest:test_attest_no_repo_passes` — PASSED
- [x] `bazel test //examples/attest:test_attest_no_repo_fails` — PASSED
- [x] `bazel test //examples/assertion/attest_external:test` — PASSED


suhalvemu:
@thesayyn  -  would appreciate a review when you get a chance.

thesayyn (approved):

PR #900, opened 2026-07-04
fix: make OCI helper actions explicit about toolchains

keegancsmith:
The --incompatible_auto_exec_groups flag requires Starlark actions to declare whether their executable/tools come from an action toolchain. This is not a Bazel default today, but it is exposed through strict preset configurations and is intended to catch rules that rely on implicit action toolchain inference.

The descriptor and load helpers combine tools from several toolchains. Mark those helper actions with toolchain = None, matching the main OCI image action, so their action toolchain behavior is explicit without changing their inputs or execution model.


keegancsmith (commented):
Edit: accidental review. deleted

PR #901, opened 2026-07-30
fix: declare repo rules reproducible so they can use the repo contents cache

alexander-scott:
Repo rules whose outputs are fully determined by their attributes must return `repository_ctx.repo_metadata(reproducible = True)`, otherwise Bazel refuses to store them in the repo contents cache - local (`--repo_contents_cache`) or remote (`--experimental_remote_repo_contents_cache`).

None of the repo rules in rules_oci did this, so every one of them was re-fetched from scratch in each fresh output base, and none could be shared remotely.

All the toolchain rules (crane, regctl, cosign, toolchains_repo) are unconditionally reproducible: they either download an artifact pinned by an integrity hash, or template files purely from their attributes.

`oci_pull` and `oci_alias` are reproducible only when pulling by digest. With a tag the registry may move it to different content between fetches, so they stay uncached in that case.


alexander-scott:
@thesayyn Hey if you have a minute, could you give some feedback to this change?

thesayyn (approved):

u9g:
The failing `test_chainguard_static_linux_amd64` job isn't from this PR: Chainguard's `:latest` started carrying `artifactType`, which #904 fixed on 2026-08-24. The re-run started 5s after #904 merged and reused this PR's old merge commit (based on 75ff282, June), so it tested without the fix. A rebase or an empty push should get a fresh merge commit and a green run. Merged locally onto 522cf5b, it applies cleanly and the pulled chainguard index.json includes artifactType.

For context, we'd like this for the repo contents cache: without it a digest-pinned 112 MB JRE base is re-extracted in every CI job.

u9g:
@thesayyn , would you mind merging this in?

